PR #46, opened 2018-07-06
Alternate formats for extracting data from MDX query

DJHig:
Hello!
First, thank you for the really useful TM1py project!
I wanted to pull data in a different format than what your normal execute_mdx gives so I added a couple of additional formats.  I am driving multiple synched Dygraphs from a single query with these new methods. I added a get_data method that works like execute_mdx, but could also be used as a template for a similar get_data_from_view, allowing data to be returned in any format. The get_data could be used as a drop-in replacement for execute_mdx since the additional arguments are optional and default is how execute_mdx works. (If you want to do this, I can change the order of the arguments so calls with positional arguments will still work).

Let me know what you think. I can give you some sample code so you can see the different formats in action. I have one that queries the default "Planning Sample" server from TM1 so it should work for everyone.
Thanks!

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -179,6 +179,111 @@ class CellService:
         finally:
             self.delete_cellset(cellset_id=cellset_id)
 
+    def get_data(self, mdx, format=None, cell_properites=None, dim_properties=None, top=None):
+        """ Execute an MDX query and return the results in the format specified
+
+        :param mdx_or_view: A string that is an MDX Query or view name
+        :param format: String specifying the format to return. Supported: 'celldict' (default), 'csv', 'raw', 'dataframe' (pandas), 'array', dygraph'
+        :param cell_properties: List of properties to be queried from the cell. E.g. ['Value', 'Ordinal', 'RuleDerived', ...]
+        :param dim_properties: List of properties to be queried from the dimension. E.g. ['UniqueName','Attributes', ...]
+        :param top: Integer limiting the number of cells and the number or rows returned
+        :return: Content in format specified above.
+        """
+        cellset_id = self.create_cellset(mdx=mdx)
+        try:
+            return self.extract_cellset(cellset_id=cellset_id, format=format, cell_properties=cell_properites, dim_properties=dim_properties, top=top)
+        finally:
+            self.delete_cellset(cellset_id=cellset_id)
+
+    def extract_cellset(self, cellset_id, format=None, cell_properties=None, dim_properties=None, top=None):
+        """ Extract data from an existing cellset and return the results in the format specified
+
+        :param cellset_id: String; ID of existing cellset
+        :param format: String specifying the format to return. Supported: 'celldict' (default), 'csv', 'raw', 'dataframe' (pandas), 'array', dygraph'
+        :param cell_properties: List of properties to be queried from the cell. E.g. ['Value', 'Ordinal', 'RuleDerived', ...]
+        :param dim_properties: List of properties to be queried from the dimension. E.g. ['UniqueName','Attributes', ...]
+        :param top: Integer limiting the number of cells and the number or rows returned
+        :return: Content in format specified above.
+        """        
+        if format == 'values':
+            data =  self.extract_cellset_values(cellset_id=cellset_id)
+            return (cell["Value"] for cell in data["Cells"])
+        elif format == 'csv':
+            return self.extract_cellset_csv(cellset_id=cellset_id)
+        elif format == 'dataframe':
+            raw_csv = self.extract_cellset_csv(cellset_id=cellset_id)
+            memory_file = StringIO(raw_csv)
+            return pd.read_csv(memory_file, sep=',')
+        else:
+            if not cell_properties:
+                cell_properties = ['Value', 'Ordinal']
+            elif 'Ordinal' not in cell_properties:
+                cell_properties.append('Ordinal')
+
+            data = self.extract_cellset_full(cellset_id=cellset_id, cell_properties=cell_properties, dim_properties=dim_properties, top=top)
+
+            if format == 'raw':
+                return data
+            elif format == 'array':
+                return Utils.build_arrays_from_cellset(raw_cellset_as_dict=data)
+            elif format == 'dygraph':
+                return Utils.build_dygraph_arrays_from_cellset(raw_cellset_as_dict=data)
+            else:
+                return Utils.build_content_from_cellset(raw_cellset_as_dict=data,
+                                                        cell_properties=cell_properties,
+                                                        top=top)
+
+    def extract_cellset_full(self, cellset_id, cell_properties=None, dim_properties=None, top=None):
+        """ Extract full Cellset data and return the raw data from TM1
+        
+        :param cellset_id: String; ID of existing cellset
+        :param cell_properties: List of properties to be queried from the cell. E.g. ['Value', 'Ordinal', 'RuleDerived', ...]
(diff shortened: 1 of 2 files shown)


TM1py/Services/CellService.py
```diff
@@ -179,6 +179,111 @@ def execute_mdx(self, mdx, cell_properties=None, top=None):
         finally:
             self.delete_cellset(cellset_id=cellset_id)
 
+    def get_data(self, mdx, format=None, cell_properites=None, dim_properties=None, top=None):
```
MariusWirtz: typo in cell_properties

MariusWirtz: Maybe dim_properties should be called elem_properties?

MariusWirtz: Maybe rename format parameter. Unnecessary use of builtin names should be avoided

TM1py/Services/CellService.py
```diff
@@ -179,6 +179,111 @@ def execute_mdx(self, mdx, cell_properties=None, top=None):
...
+        """ Execute an MDX query and return the results in the format specified
+
+        :param mdx_or_view: A string that is an MDX Query or view name
```
MariusWirtz: wrong parameter in docstring

TM1py/Utils/Utils.py
```diff
@@ -111,6 +111,122 @@ def build_content_from_cellset(raw_cellset_as_dict, cell_properties, top=None):
...
+def build_dygraph_arrays_from_cellset(raw_cellset_as_dict):
+    """ Transform raw 1,2 or 3-dimension cellset data into dygraph-friendly format
+
+    :param raw_cellset_as_dict: raw data from TM1
+    :return: dict : { titles: [], headers: [axis][], cells: { Page0: [  [column name, column values], [], ... ], ...} }
+    """
+    dimensionality = len(raw_cellset_as_dict['Axes'])
+    cardinality = [raw_cellset_as_dict['Axes'][axis]['Cardinality'] for axis in range(dimensionality)]
+
+    headerMap = build_headers_from_cellset(raw_cellset_as_dict)
```
MariusWirtz: header_map

TM1py/Utils/Utils.py
```diff
@@ -111,6 +111,122 @@ def build_content_from_cellset(raw_cellset_as_dict, cell_properties, top=None):
...
+    cells = {}
+    for z in range(cardinality[2]):
+        zHeader = headers[2][z]['name']
+        page = []
+        for x in range(cardinality[0]):
+            xHeader = headers[0][x]['name']
+            row = [xHeader]
+            for y in range(cardinality[1]):
+                cell_addr = (x + cardinality[0] * y + cardinality[0] * cardinality[1] * z)
+                row.append(float("{0:.1f}".format(raw_cellset_as_dict['Cells'][cell_addr]['Value'] or 0)))
```
MariusWirtz: Would be nice if precision was configurable I think

TM1py/Utils/Utils.py
```diff
@@ -111,6 +111,122 @@ def build_content_from_cellset(raw_cellset_as_dict, cell_properties, top=None):
     return content_as_dict
 
 
+def build_arrays_from_cellset(raw_cellset_as_dict):
+    """ Transform raw 1,2 or 3-dimension cellset data into concise dictionary
```
MariusWirtz: Maybe it makes sense to add a comment what we can use this response type for

TM1py/Services/CellService.py
```diff
@@ -179,6 +179,111 @@ def execute_mdx(self, mdx, cell_properties=None, top=None):
...
+            data =  self.extract_cellset_values(cellset_id=cellset_id)
+            return (cell["Value"] for cell in data["Cells"])
+        elif format == 'csv':
+            return self.extract_cellset_csv(cellset_id=cellset_id)
+        elif format == 'dataframe':
+            raw_csv = self.extract_cellset_csv(cellset_id=cellset_id)
+            memory_file = StringIO(raw_csv)
+            return pd.read_csv(memory_file, sep=',')
+        else:
+            if not cell_properties:
```
MariusWirtz: I think lines 218 to 221 are obsolete

TM1py/Utils/Utils.py
```diff
@@ -111,6 +111,122 @@ def build_content_from_cellset(raw_cellset_as_dict, cell_properties, top=None):
...
+
+    return {'titles':titles, 'headers':headers, 'cells':cells}
+
+def build_dygraph_arrays_from_cellset(raw_cellset_as_dict):
+    """ Transform raw 1,2 or 3-dimension cellset data into dygraph-friendly format
+
+    :param raw_cellset_as_dict: raw data from TM1
+    :return: dict : { titles: [], headers: [axis][], cells: { Page0: [  [column name, column values], [], ... ], ...} }
+    """
+    dimensionality = len(raw_cellset_as_dict['Axes'])
```
MariusWirtz: Maybe outsource these two lines into seperate function to avoid code duplication:
dimensionality = len(raw_cellset_as_dict['Axes'])
cardinality = [raw_cellset_as_dict['Axes'][axis]['Cardinality'] for axis in range(dimensionality)]

Or maybe build_headers_from_cellset could also return dimensionality and cardinality?

DJHig:
Hi Marius,
Thank you for the review and comments. I made the requested changes.
I added an example 'cells' output from both of my added output formats. The arrays one was for an earlier charting library I was evaluating that needed the rows to be arrays, rather than the columns (like dygraphs or pandas). The arrays format is also useful for grids that usually want the data by rows.

I did not add value_precision to the build_content_from_cellset, but I can add that if you would like. 

I don't the get_data to be a single function and it is fine with me to split it into the smaller ones by specific format. I was mainly trying to reduce the code duplication with having to repeat them for Views, in addition to MDX queries. I am currently only using MDX queries so I do not need the Views now anyway.

Regarding dygraphs, I can try to come up with an example. Yes, I was using Flask, but ran into a weird bug so I switched to Sanic.

Thanks!

TM1py/Services/CellService.py
DJHig: Oh, I only added these here since build_content_from_cellset needs the cell_properties to know which properties to add. It could be derived from the data and then it does not need them. Also, not sure if top is needed there since the query would be limiting the data.

DJHig:
I also updated the sample in TM1py-samples to match the new argument names.

TM1py/Services/CellService.py
DJHig: These lines are only needed to make sure the Utils.build_content_from_cellset method has the 'Ordinal' passed in the cell_properties (actually, to make sure it matches what the query in extract_cellset_full used.) I think Utils.build_content_from_cellset could derive that from the data and it does not need to be passed. I'm not sure it needs the top argument either, since the query already limited these values. What do you think about changing that function to take only raw_cellset_as_dict and value_precision, like the two new ones?

DJHig: Sorry, didn't mean to send duplicate comment.

MariusWirtz: You are absolutely right. There is no need to query the Ordinal property from the cell. This needs to be fixed. 

Yes. It doesn’t necessarily need the top argument. We added it, so that theoretically the utility function is independent from the REST call. 
I think it makes sense, to add the value-precision as an argument. 

I wouldn’t want to remove the cell_properties argument. Sometimes it's useful to query other properties from the cell, like RuleDerived or CellAnnotation.

PR #55, opened 2018-07-31
Fix build headers for aggregate queries

DJHig:
Hello,
This fixes the build_content_from_cellset when aggregate queries are used. I see now the difference between member and elements. I mainly wanted Attributes and I see that they are available on Members so I refactored the code a bit assuming that member_properties will be mostly used. Elem_properties can still be used and will be included in the result set, but it does not try to flatten the properties between elements and members anymore. Also, I allow elem_properties to be None meaning we do not even want to query them. I only force the member Name property with dygraphs and arrays.

I did a quick refactor on the execute_mdx and build_content_from_cellset to remove the need to pass cell_properties to the build_content_from_cellset function; all properties included in the query will be returned. The UniqueName is forced on the query for both element and member, but only one is really needed. The build_content_from_cellset uses Element, but I'm not sure if that is any better or different then Member for your needs.

Thank you!

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -200,12 +200,19 @@ class CellService:
         cellset_id = self.create_cellset_from_view(cube_name=cube_name, view_name=view_name, private=private)
         return self.extract_cellset(cellset_id=cellset_id, cell_properties=cell_properties, top=top)
 
-    def execute_mdx_raw(self, mdx, cell_properties=None, elem_properties=None, member_properties=None, top=None):
+    def execute_mdx_raw(
+            self,
+            mdx,
+            cell_properties=None,
+            elem_properties=None,
+            member_properties=None,
+            top=None):
         """ Execute MDX and return the raw data from TM1
 
         :param mdx: String, a valid MDX Query
         :param cell_properties: List of properties to be queried from the cell. E.g. ['Value', 'Ordinal', 'RuleDerived', ...]
         :param elem_properties: List of properties to be queried from the elements. E.g. ['UniqueName','Attributes', ...]
+        :param member_properties: List of properties to be queried from the members. E.g. ['UniqueName','Attributes', ...]
         :param top: Integer limiting the number of cells and the number or rows returned
         :return: Raw format from TM1.
         """
@@ -232,6 +239,7 @@ class CellService:
         :param private: True (private) or False (public)
         :param cell_properties: List of properties to be queried from the cell. E.g. ['Value', 'Ordinal', 'RuleDerived', ...]
         :param elem_properties: List of properties to be queried from the elements. E.g. ['UniqueName','Attributes', ...]
+        :param member_properties: List of properties to be queried from the members. E.g. ['UniqueName','Attributes', ...]
         :param top: Integer limiting the number of cells and the number or rows returned
         :return: Raw format from TM1.
         """
@@ -321,7 +329,13 @@ class CellService:
         cellset_id = self.create_cellset_from_view(cube_name=cube_name, view_name=view_name, private=private)
         return self.extract_cellset_cellcount(cellset_id)
 
-    def execute_mdx_ui_dygraph(self, mdx, value_precision=2):
+    def execute_mdx_ui_dygraph(
+            self,
+            mdx,
+            elem_properties=None,
+            member_properties=["Name"],
+            value_precision=2,
+            top=None):
         """ Execute MDX get dygraph dictionary
         Useful for grids or charting libraries that want an array of cell values per column
         Returns 3-dimensional cell structure for tabbed grids or multiple charts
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Thanks for the PR! I will add a few comments to the code. 

You are right about element-unique-names and member-unique-names 99% of the times.
But in some cases (when the element is a child of multiple parents) the member-unique-name and the element-unique-name can be different:
member-unique-name: [Date].[Next Day^2018-08-01] 
element-unique-name: [Date].[2018-08-01]

In the build_content_from_cellset function we handle it this way:
We use the element-unique-name if it is present. 
If the element property is not present on the member, the member must be a calculated member. Then we can safely consume the member-unique-name (as calculated members don't have ^ in their unique names).
`[member['Element']['UniqueName'] if member['Element'] else member['UniqueName']`
This way we always end up with [Date].[2018-08-01], irrespective of whether [2018-08-01] is a normal member or a calculated member.
Unfortunately, in order for this to work we need to query the element-unique-name and the member-unique-name.

Thanks for the fix on the build_content_from_cellset function. Nice one!

TM1py/Services/CellService.py
```diff
@@ -321,7 +329,13 @@ def execute_view_cellcount(self, cube_name, view_name, private=True):
         cellset_id = self.create_cellset_from_view(cube_name=cube_name, view_name=view_name, private=private)
         return self.extract_cellset_cellcount(cellset_id)
 
-    def execute_mdx_ui_dygraph(self, mdx, value_precision=2):
+    def execute_mdx_ui_dygraph(
+            self,
+            mdx,
+            elem_properties=None,
+            member_properties=["Name"],
```
MariusWirtz: Common trap! Functions shouldn't have mutable default arguments:
https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/

TM1py/Services/CellService.py
```diff
@@ -339,14 +353,28 @@ def execute_mdx_ui_dygraph(self, mdx, value_precision=2):
...
         return Utils.build_ui_dygraph_arrays_from_cellset(raw_cellset_as_dict=data, value_precision=value_precision)
 
-    def execute_view_ui_dygraph(self, cube_name, view_name, private=True, value_precision=2):
+    def execute_view_ui_dygraph(
+            self,
+            cube_name,
+            view_name,
+            private=True,
+            elem_properties=None,
+            member_properties=["Name"],
```
MariusWirtz: better replace with None and check against None inside function.

TM1py/Services/CellService.py
```diff
@@ -376,14 +404,26 @@ def execute_view_ui_dygraph(self, cube_name, view_name, private=True, value_prec
...
+                                        member_properties=list(set(member_properties or []) | {"Name"}),
+                                        top=top)
         return Utils.build_ui_dygraph_arrays_from_cellset(raw_cellset_as_dict=data, value_precision=value_precision)
 
-    def execute_mdx_ui_array(self, mdx, value_precision=2):
+    def execute_mdx_ui_array(
+            self,
+            mdx,
+            elem_properties=None,
+            member_properties=["Name"],
```
MariusWirtz: better replace with None and check against None inside function.

TM1py/Services/CellService.py
```diff
@@ -411,14 +451,28 @@ def execute_mdx_ui_array(self, mdx, value_precision=2):
...
         return Utils.build_ui_arrays_from_cellset(raw_cellset_as_dict=data, value_precision=value_precision)
 
-    def execute_view_ui_array(self, cube_name, view_name, private=True, value_precision=2):
+    def execute_view_ui_array(
+            self,
+            cube_name,
+            view_name,
+            private=True,
+            elem_properties=None,
+            member_properties=["Name"],
```
MariusWirtz: better replace with None and check against None inside function.

TM1py/Services/CellService.py
```diff
@@ -477,25 +537,15 @@ def extract_cellset_raw(
         elif 'Ordinal' not in cell_properties:
             cell_properties.append('Ordinal')
 
-        if not elem_properties:
-            elem_properties = ['UniqueName']
-        elif 'UniqueName' not in elem_properties:
-            elem_properties.append('UniqueName')
-
-        if not member_properties:
```
MariusWirtz: Unfortunately we can't remove this code. Still needed to cater for potential calculated Members

TM1py/Services/CellService.py
```diff
@@ -477,25 +537,15 @@ def extract_cellset_raw(
...
-            member_properties = ['UniqueName']
-        elif 'UniqueName' not in member_properties:
-            elem_properties.append('UniqueName')
```
MariusWirtz: seems to be an old bug, but in line 488 it should say:
member_properties.append('UniqueName')

TM1py/Services/CellService.py
```diff
@@ -477,25 +537,15 @@ def extract_cellset_raw(
         elif 'Ordinal' not in cell_properties:
             cell_properties.append('Ordinal')
 
-        if not elem_properties:
```
MariusWirtz: Unfortunately we can't remove this code. Still needed to cater for non calculated members.

TM1py/Services/CellService.py
```diff
@@ -540,10 +590,14 @@ def extract_cellset(self, cellset_id, cell_properties=None, top=None):
...
             cell_properties.append('Ordinal')
-        raw_cellset = self.extract_cellset_raw(cellset_id, cell_properties=cell_properties, top=top)
-        return Utils.build_content_from_cellset(raw_cellset_as_dict=raw_cellset,
-                                                cell_properties=cell_properties,
-                                                top=top)
+
+        raw_cellset = self.extract_cellset_raw(
+                cellset_id,
+                cell_properties=cell_properties,
+                elem_properties=['UniqueName'],
```
MariusWirtz: I think in 597 and 598 we need to pass elem_properties and member_properties variables

TM1py/Services/CellService.py
DJHig: Hi Marius,
Yes, UniqueName is needed for the execute_mdx / extract_cellset / build_content_from_cells functions, but it might not needed for other uses so I did not want to force returning "UniqueName", except in those functions. The call from extract_cellset to extract_cellset_raw forces those properties (and only those properties) on element and members.

TM1py/Services/CellService.py
DJHig: Hi Marius,
Both elem_properties and member_properties are being passed. Do you mean not as a an explicit list of only 'UniqueName' ?

TM1py/Services/CellService.py
MariusWirtz: that's a fair point. It does blow up the response size, so we should be careful with what we are querying from TM1, but I can't think of a use case where we would not want to cater for potential calculated members. Do you see one?

If there is no actual use case to call this function without asking for the member-unique-names I would prefer to play it safe and always query them.

DJHig:
Hi @MariusWirtz ,

Thank you for the code review. I updated the mutable default arguments. 

I think the other comments are not an issue. I updated the TM1py-samples with the arguments in this version so you can verify that all of the functions return what they should and nothing more than is requested.

TM1py/Services/CellService.py
MariusWirtz: Please ignore my previous comment on this one. It's wrong.

TM1py/Services/CellService.py
DJHig: Hi Marius,
The member UniqueName and element UniqueName will both be queried for execute_mdx / extract_cellset / build_content_from_cells. 

My purpose does not need UniqueName so I was hoping to allow them to not be included when extracting raw.

MariusWirtz: OK

PR #60, opened 2018-08-19
Build content from cellset

ajmyers:
This request includes a rewrite of build_content_from_cellset to be more efficient and work with data returned on any axis.

I also removed sort_addresstuple in the second commit since it seemed like it might be safe to remove. Let me know if you want this to be handled a different way and I can update.

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -1,10 +1,9 @@
 import collections
 import json
-import re
 import sys
 import pandas as pd
 
-from math import floor
+from re import match
 
 from TM1py.Objects.Server import Server
 from TM1py.Objects.Process import Process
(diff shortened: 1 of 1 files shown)


MariusWirtz:
Thanks for the PR! Much appreciated! I will add a few comments to the code.

P.S. 
it isn't actually documented anywhere yet, but if you want to run the Tests you can go to the Tests folder, adjust the config.ini and execute:
`python -m pytest Tests/__init__.py`
Or  you can configure PyCharm to use Nosetests and run them in PyCharm. 

(Depending on the size of the transactionlog of the target instance some Tests may take a very long though)

Ideally it looks like this:
image

MariusWirtz:
Hi @ajmyers ,

@vviau accepted the MR, even though there were still some requested changes open.
I now reverted the MR. 

Please open a new MR and we will accept the changes once we have the issues fixed.

PR #61, opened 2018-08-19
New session_context argument for TM1Service, RESTService

MariusWirtz:
New optional parameter for TM1Service `session_context`, that controls what is displayed in the `session` column in Arc / Pulse / TM1top.  Default value is "TM1py". 
Default Can be overwritten as follows:

```

import configparser

config = configparser.ConfigParser()
config.read('..\config.ini')

APP_NAME = "My TM1py Application"

with TM1Service(**config['tm1srv01'], session_context=APP_NAME) as tm1:
    dimensions = tm1.dimensions.get_all_names()
    for dim in dimensions:
        print(dim)

```

### DIFF
--- a/TM1py/Services/RESTService.py
+++ b/TM1py/Services/RESTService.py
@@ -1,10 +1,12 @@
 # -*- coding: utf-8 -*-
-from TM1py.Exceptions import TM1pyException
-
 import functools
-import requests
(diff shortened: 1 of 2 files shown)


rclapp:
Excellent, this is exactly what we need.

Issue #62, opened 2018-09-04
TM1py should be able to decode Base64 password


vviau:
Hi Marius,

As spoken, it would be better if the TM1 connection parameters between RushTI and TM1py were consistent. In RushTI requires the password to be encode in Base64 whereas TM1py uses plain text password.

To make sure we do not break existing scripts, TM1py should support password in plain text or in Base64 format.
TM1py should check first if it can connect to TM1 by reading the password as it was plain text and if it can't connect it tries to decode it.

I hope this is clear,

Cheers,

Vincent

PR #66, opened 2018-09-25
Retry Authentication with b64 decoded password

MariusWirtz:
Now retry Authentication in RestService with b64decoded password, if authentication with plain password fails.

https://github.com/cubewise-code/TM1py/issues/62


vviau:
Nice one Marius!

PR #114, opened 2019-03-14
CAMPassport authorization

adscheevel:
added functionality to _build_authorization_token function in RestService.py to take gateway as an argument where gateway is the Cognos dispatcher used for CAM security. will use negotiated login attempt to get passport cookie for authorization header instead of taking username & password. If the negotiated login fails, it will set the token to use the username and password, provided they were supplied. Attached is example of config.ini file for on premise installation with gateway parameter defined.

I did this modification because i didnt want to save the base64 password in a config file and am lucky to have cognos CAM with windows integrated authentication set up in our TM1 environment. My python skills are not great, my edits might not be the best way to accomplish what I set out to do.

image

### DIFF
--- a/TM1py/Services/RESTService.py
+++ b/TM1py/Services/RESTService.py
@@ -4,7 +4,7 @@ import sys
 from base64 import b64encode, b64decode
 
 import requests
-
+from requests_negotiate_sspi import HttpNegotiateAuth
 from TM1py.Exceptions import TM1pyException
 
 # import Http-Client depending on python version
@@ -16,7 +16,6 @@ else:
 
 def httpmethod(func):
     """ Higher Order Function to wrap the GET, POST, PATCH, PUT, DELETE methods
-
         Takes care of:
         - encoding of url and payload
         - verifying response. Throws TM1pyException if StatusCode of Response is not OK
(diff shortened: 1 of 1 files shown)


MariusWirtz:
Hi, 

thanks for the contribution to TM1py. Before I will merge the changes I want to challenge it. Unfortunately I currently can't get it to run with my setup.

I got a TM1 instance that is using CAM and supporting SSO (IntegratedSecurityMode=5) ready. 
Now I understand that as the gateway I need to assign the same value that is specified in the tm1s.cfg as `ClientCAMURI`. Is that correct?

adscheevel:
Correct, it's should also match the address returned in the WWW-Authenticate header if you send a request to the instance without any authorization header.

MariusWirtz:
@adscheevel 
sorry for my late response on this one. I finally managed to get my hands on a properly set up TM1 and CAM environment. We tested your changes and it works like a charm! 

I will close this branch in favor of #120 , as I wanted to do a few (rather cosmetical) changes to the branch before it's merged into the master. #120  now contains your commit e58dd1672408a9295e70f517303f260d3709e9f7 and my changes on top 
595ce9f60cff6d729abcfcc2af9c9f053607dec9

Thanks for your contribution!

PR #115, opened 2019-03-19
Added a **timeout** attribute to connection.

csjean:


MariusWirtz:
Hi @csjean,
thanks for the contribution. Much appreciated.

Would be a nice feature if the default timeout could be overruled on individual calls. I will open a seperate issue for that.

Issue #123, opened 2019-04-08
Support HTTPS URL on Cognos Dispatcher for CAM SSO


MariusWirtz:
TM1py needs to support SSO through CAM when the Cognos Dispatcher is adressed through HTTPS.

The TM1Service constructor requires a new optional argument like `gateway_verify`. The value can be `false` if no verification is necessary or the path to the appropriate .cer file.

rclapp:
Works like a charm. SSO will make prod deployment's so much easier.

MariusWirtz:
Nice one! Thanks for the fix.

PR #131, opened 2019-06-05
Updated RESTService to leverage the classes verify setting

rclapp:
Update the RESTService to leverage the service's verify settings when issuing the CAM negotiation request. This allows the verify setting to be standardized for each individual connection and makes working through the SSL challenges easier


MariusWirtz:
Fixes #123

PR #133, opened 2019-06-07
Fixed ChoreService chore update function

ducklingasa:
Hi Marius,

I've realised that the chore update function is not working properly in case the source and destination chore has different number of tasks inside (if there were less, new tasks were not added just caused a TM1 rest error, in case there were more the unnecessary ones were not deleted). Following your logic, I've enhanced the code to make sure that different chore contents are fully synchronised during the update.

Regards,
Adam


MariusWirtz:
I was able to replicate the issue. Thanks for the fix. Much appreciated!

PR #155, opened 2019-07-15
Expand exception detail

clappr-zz:
Expanded TM1Py  Exception to include the returned headers. Returning the headers allow for easier trouble shooting and auto-detection of the CAM Gateway. This will allow for easier roll out to users using SSO. 

` try:
        response = TM1Service(**kwargs)
  except TM1pyException as e:
        response_headers = e.headers
        authenticate_header = response_headers.get("WWW-Authenticate", "No Gateway Found")
        if authenticate_header is not "No Gateway Found":
            gateway = authenticate_header.split("CAMPassport")[1].strip().split(",")[0]
            return gateway
        else:
            raise authenticate_header`

### DIFF
--- a/TM1py/Exceptions/Exceptions.py
+++ b/TM1py/Exceptions/Exceptions.py
@@ -30,7 +30,7 @@ class TM1pyException(Exception):
         return self._headers
 
     def __str__(self):
-        return "Text: {} Status Code: {} Reason: {} Headers {}".format(self._response,
+        return "Text: {} Status Code: {} Reason: {} Headers: {}".format(self._response,
                                                                 self._status_code,
                                                                 self._reason,
                                                                 self._headers)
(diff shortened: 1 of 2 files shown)


TM1py/Exceptions/Exceptions.py
```diff
@@ -24,5 +25,9 @@ def reason(self):
     def response(self):
         return self._response
 
+    @property
+    def headers(self):
+        return self._headers
+
     def __str__(self):
         return "Text: {} Status Code: {} Reason: {}".format(self._response, self._status_code, self._reason)
```
MariusWirtz: I think it would make sense to adjust the `__str__` method of this class. So it would look kinda like this:
``` python
  def __str__(self):
        return "Text: {} Status Code: {} Reason: {} Headers: {}".format(
            self._response,
            self._status_code,
            self._reason,
            self._headers)
```

clappr-zz: Sounds good I will make an update and commit again

Issue #190, opened 2019-11-15
TM1Service doesn't provide possibility to retain live connection


pbuncik:
**Describe the bug**
Imagine a situation when connection is already established from outside of Python with known SessionID and TM1Service object is constructed with this SessionID.
Currently TM1Service implements a cleanup of connection in its __exit__ method, but for the approach described above the cleanup will close the connection. It would be nice to keep the connection live for instances like this.

**To Reproduce**
Allocate SessionID (token) outside Python, use it to initiate TM1Service. After TM1Service is discarded, the session is closed and SessionID can't be used anymore.

**Expected behavior**
A new switch in TM1Service configuration to retain connection.

**Version**
TM1py 1.3.1
TM1 Server Version: 11.6.00000.14

**Additional context**
NA

pbuncik:
PS: this is not bug per se, but rather a feature request. As a temporary solution I can define a new descendant of TM1Service class that will redefine __exit__ method by own implementation.

MariusWirtz:
Hi @pbuncik ,

We have a similar feature already implemented. Perhaps what you need intend to do can be accomplished with the `save_to_file` and `restore_from_file` methods
Here is a sample how to use them:
https://github.com/cubewise-code/tm1py-samples/blob/master/Administration/session_persist.py

https://github.com/cubewise-code/tm1py-samples/blob/master/Administration/session_reuse.py

does it work ?

Cheers,

Marius

pbuncik:
Hi @MariusWirtz ,
I'm not convinced the approach proposed in session_persist.py would work. Here is why:
The idea is:
1) Open TM1 session outside TM1py in a web application (running on a web server). This is interactive application that uses the connection to query TM1 database frequently and also maintains the connection open for given user until the user logs out.
2) At some point of time we will launch a tool written in Python and utilizing TM1py.
3) The tool will reuse session open in step 1) by using session ID token that is transferred from the web application to the tool.
4) After the tool is finished it must retain the session open otherwise the user would experience logout from TM1 and would have to login again from within the web application again - this is unwanted behavior.

This approach is briefly what your second example is demonstrating - the session token is reused to hook-up to already connected session.
First example is demonstrating how to serialize the session into a file, but it doesn't solve problem with logging of the session during `__exit__` method call when TM1Session or RESTService objects are destroyed by garbage collector. At this point I need to avoid the call to logout the session to leave it open for the web application. The user should not be logged off and the call to the tool should be transparent to the user.
[...]

PR #191, opened 2019-11-15
proposed fix of issue #190 including test

pbuncik:
Please find proposed fix of issue #190 to retain live connection.

### DIFF
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -84,6 +84,15 @@ class CubeService(ObjectService):
         request = "/api/v1/Cubes('{}')".format(cube.name)
         return self._rest.PATCH(request, cube.body)
 
+    def check_rules(self, cube):
+        """ Check rules syntax for existing cube on TM1 Server
+
+        :param cube: instance of TM1py.Cube
+        :return: response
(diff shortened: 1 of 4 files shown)


pbuncik:
I have added one more proposal for a method in CubeService to check rule syntax.

TM1py/Services/CubeService.py
```diff
@@ -84,6 +84,15 @@ def update(self, cube):
...
         return self._rest.PATCH(request, cube.body)
 
+    def check_rules(self, cube):
+        """ Check rules syntax for existing cube on TM1 Server
+
+        :param cube: instance of TM1py.Cube
+        :return: response
+        """
+        request = "/api/v1/Cubes('{}')/tm1.CheckRules".format(cube.name)
+        return self._rest.POST(request)
```
MariusWirtz: For this method you are passing an instance of the `Cube` class, but actually we are just making use of the name of the cube.
I suggest we change the argument to `cube_name` (string) or we check whether the argument is of type `Cube` or `str` inside the method

MariusWirtz:
@pbuncik ,

please create a new MR that contains only the `check_rules` method

PR #193, opened 2019-12-06
Fixed ApplicationService request URL in case of empty path (root folder)

ducklingasa:
Hi Marius and Tomasz,

I found out that the new functions (get, delete, create, exists) don't handle properly the **root** folder under the Application tree (since this case the path is empty).
I've modified the functions and tested the functionality.

Cheers,
Adam


MariusWirtz:
Thanks for the MR. Much appreciated!

PR #194, opened 2019-12-07
Proposal of method to check rules syntax

MariusWirtz:


MariusWirtz:
@pbuncik 
I outsourced your commit into a new MR. Please accept if it is OK.

PR #197, opened 2019-12-12
Restrict URL escaping to object names only

ducklingasa:
Hi Marius,

I've changed the way how the escaping works.
It first recognizes the object names inside the brackets and quotes then doubles the quotes afterwards.
@tombrzy will do some more testing, and we will let you know the results.

Cheers,
Adam


MariusWirtz:
@ducklingasa and @tombrzy 

Thanks guys! Will merge in the changes and add a few test cases to it.

PR #205, opened 2020-01-24
Function to execute dimensional mdx queries

rclapp:
I put this in the element service, but I could see it working in the dim or hierarchy service. Let me know if I should move it.

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -279,11 +279,11 @@ class ElementService(ObjectService):
 
     def execute_set_mdx(self, mdx,
                         top_records=None,
-                        member_properties = ['Name', 'Weight'],
-                        parent_properties = ['Name', 'UniqueName'],
-                        element_properties = ['Type', 'Level']):
+                        member_properties = ('Name', 'Weight'),
+                        parent_properties = ('Name', 'UniqueName'),
+                        element_properties = ('Type', 'Level')):
         '''
-        :method to execute an MDX statment against a dimension
+        :method to execute an MDX statement against a dimension
         :param mdx: valid dimension mdx statement
         :param top_records: number of records to return, default: all elements no limit
         :param member_properties: list of member properties to return, will always return name
@@ -291,7 +291,6 @@ class ElementService(ObjectService):
         :param element_properties: list of element properties to return, can be empty
         :return: dictionary of members, unique names, weights, types, and parents
         '''
-
 
         top = f"$top={top_records};" if top_records else ""
 
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -277,20 +277,49 @@ class ElementService(ObjectService):
         get_members(consolidation_tree)
         return members
 
-    def execute_set_mdx(self, mdx, top=10000):
+    def execute_set_mdx(self, mdx,
+                        top_records=None,
+                        member_properties = ['Name', 'Weight'],
+                        parent_properties = ['Name', 'UniqueName'],
+                        element_properties = ['Type', 'Level']):
         '''
         :param mdx: valid dimension mdx statement
-        :param top: number of records to return
-        :return: dictionary of members, unique names, weights, types, ordinals, and parents
+        :param top: number of records to return, default: all elements no limit
+        :return: dictionary of members, unique names, weights, types, and parents
         '''
+        top = f"$top={top_records};" if top_records else ""
 
-        request = f'/api/v1/ExecuteMDXSetExpression?$expand=Tuples($top={top};' \
-                  '$expand=Members($select=Name,Ordinal,Weight;' \
-                  '$expand=Parent($select=Name,UniqueName),' \
-                  'Element($select=Type,Level)))'
+        if not member_properties:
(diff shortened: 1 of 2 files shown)


TM1py/Services/ElementService.py
```diff
@@ -275,3 +276,21 @@ def get_members(element):
 
         get_members(consolidation_tree)
         return members
+
+    def execute_set_mdx(self, mdx, top=10000):
```
MariusWirtz: In order to be compliant with the rest of TM1py I think it would be nice to set the default value for the argument to `None`.

TM1py/Services/ElementService.py
```diff
@@ -275,3 +276,21 @@ def get_members(element):
...
+        '''
+        :param mdx: valid dimension mdx statement
+        :param top: number of records to return
+        :return: dictionary of members, unique names, weights, types, ordinals, and parents
+        '''
+
+        request = f'/api/v1/ExecuteMDXSetExpression?$expand=Tuples($top={top};' \
+                  '$expand=Members($select=Name,Ordinal,Weight;' \
+                  '$expand=Parent($select=Name,UniqueName),' \
+                  'Element($select=Type,Level)))'
```
MariusWirtz: python `f strings` are nice. We should be using them way more often in TM1py I think. 👍

Why ask for the Ordinal of the Member ? I think the ordinal is given implicitly since the Collections in the REST response are ordered.

In a future version of this we could perhaps build the url dynamically based on function arguments, so that the user could specify `member_properties`, `parent_properties` and `element_properties`. If no `parent_properties` are passed we wouldn't need to expand the parent then.

TM1py/Services/ElementService.py
```diff
@@ -275,3 +276,21 @@ def get_members(element):
...
+
+        payload = {"MDX": mdx}
+        response = self._rest.POST(request, json.dumps(payload, ensure_ascii=False))
+        raw_dict = response.json()
+        return [tuples['Members'][0] for tuples in raw_dict['Tuples']]
```
MariusWirtz: I think instead of 

``` python
return [tuples['Members'][0] for tuples in raw_dict['Tuples']]
```

it would be better to write
``` python
return [tuples['Members'] for tuples in raw_dict['Tuples']]
```

so that it also works for "Sets of Tuples":

``` python
with TM1Service(address=ADDRESS, port=PORT, user=USER, password=PASSWORD, ssl=SSL) as tm1:

    mdx = "{ ([region].[europe],[month].[jan]) }"

    response = tm1.dimensions.hierarchies.elements.execute_set_mdx(mdx)
    print(response)
```

rclapp:
Sounds good, I'll make the updates and update the request.

Ryan Clapp
AWS Business Systems
Sr. Manager

Sent from my mobile device

On Feb 2, 2020 12:52 PM, Marius Wirtz <notifications@github.com> wrote:

In TM1py/Services/ElementService.py<https://github.com/cubewise-code/tm1py/pull/205#discussion_r373874123>:

> @@ -275,3 +276,21 @@ def get_members(element):

         get_members(consolidation_tree)
         return members
+
+    def execute_set_mdx(self, mdx, top=10000):


In order to be compliant with the rest of TM1py I think it would be nice to set the default value for the argument to None.

-

MariusWirtz:
I'm fine to keep it in the `ElementService`. 

I see it as an advanced version of the `execute_mdx` method in the `DimensionService`.
The old function has many limitations. It required the `}ElementAttributes_` cube to exist, it always returns a list of element names and it only works with sets of members (not tuples).

Perhaps we could just add a comment (or even a warning?) to the old function that it is no longer "recommended" ?

TM1py/Services/ElementService.py
rclapp: Fixed in the latest commit

TM1py/Services/ElementService.py
rclapp: Made the properties dynamic.

TM1py/Services/ElementService.py
rclapp: Fixed in the latest commit

rclapp:
Added new commits that make the properties dynamic. Also made added a depreciation warning to the old method.

TM1py/Services/ElementService.py
```diff
@@ -275,3 +276,53 @@ def get_members(element):
 
         get_members(consolidation_tree)
         return members
+
+    def execute_set_mdx(self, mdx,
+                        top_records=None,
+                        member_properties = ['Name', 'Weight'],
+                        parent_properties = ['Name', 'UniqueName'],
+                        element_properties = ['Type', 'Level']):
```
MariusWirtz: This is a common gotcha! https://docs.python-guide.org/writing/gotchas/

It's safer to use a `tuple` here, as it is immutable, or to use `None` as a default argument and assign the list to the argument inside the fucntion.

MariusWirtz:
Thanks for adding the warning and making it dynamic! 
Just left one more comment on the code, then we can merge it.

rclapp:
Updated to use tuples. Will still support explicitly turning off those properties by passing none.

Issue #202, opened 2019-12-30
Syntax error with transaction log query while using USER argument


user1493:
I tried to perform a transaction log query using `tm1.server.get_transaction_log_entries`

Passed `since`argument alone and it worked:
**/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=TimeStamp ge 2019-12-30T20:45:05Z**

Passed `since` with `user`, but failed:
**/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=User eq 'TEST/user' and TimeStamp ge 2019-12-30T20:14:57Z**

Received the below error:

> TM1py.Exceptions.Exceptions.TM1pyException: Text: {"error":{"code":"278","message":"Syntax error at position 10, near \"TEST\" in $filter: Expecting end of input."}} Status Code: 400 Reason: Bad Request Headers: {'Content-Length': '123', 'Connection': 'keep-alive', 'Content-Encoding': 'gzip', 'Cache-Control': 'no-cache', 'Content-Type': 'application/json; charset=utf-8', 'OData-Version': '4.0'}

user1493:
Tried with `cube` and same result:
**/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'Sales Forecast' and TimeStamp ge 2019-12-30T21:02:03Z**

> TM1py.Exceptions.Exceptions.TM1pyException: Text: {"error":{"code":"278","message":"Syntax error at position 10, near \"Sales\" in $filter: Expecting end of input."}} Status Code: 400 Reason: Bad Request Headers: {'Content-Length': '124', 'Connection': 'keep-alive', 'Content-Encoding': 'gzip', 'Cache-Control': 'no-cache', 'Content-Type': 'application/json; charset=utf-8', 'OData-Version': '4.0'}

MariusWirtz:
Hi @user1493,

thanks for the catch!

I just looked up the latest documentation of the API. It says there that the `TransactionLogEntries` are not filterable by User.
I think in the TM1 API they made some changes to this functionality. I'm confident that it used to work when I wrote it initially in 2018 (still on 10.2.2).
I suppose if you need to filter by user, you have to do that on the client side.

For me (TM1 version `11.5.00000.23`) filtering by cube + since still works though.
Can you check if this works for you:

``` python
with TM1Service(address=ADDRESS, port=PORT, user=USER, password=PASSWORD, ssl=SSL) as tm1:
[...]

PR #207, opened 2020-01-28
Added 'clear' data spread method

rkvinoth:
1. A new method in `CellService`: `relative_proportional_spread`

**Sample:**

```
import configparser

from TM1py import TM1Service

config = configparser.ConfigParser()
config.read(r'config.ini')

with TM1Service(**config['tm1srv01']) as tm1:
    tm1.cubes.cells.relative_proportional_spread(
        cube="c1",
        unique_element_names=("[d1].[c1]", "[d2].[e3]"))
```

2. Updated the `Utils` module to fix https://github.com/cubewise-code/tm1py/issues/202#issue-543913984 . Replaced the `re` pattern in line **14**

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -122,6 +122,33 @@ class CellService:
                     *Utils.dimension_hierarchy_element_tuple_from_unique_name(unique_element_name))))
 
         self._post_against_cellset(cellset_id=cellset_id, payload=payload, delete_cellset=True)
+        
+    def clear_spread(
+            self,
+            cube,
+            unique_element_names):
+        """ Execute clear spread
+        :param cube: name of the cube
+        :param unique_element_names: target cell coordinates as unique element names (e.g. ["[d1].[c1]","[d2].[e3]"])
+        :return:
+        """
+        mdx = """
+        SELECT
+        {{ {rows} }} ON 0
+        FROM [{cube}]
+        """.format(rows="}*{".join(unique_element_names), cube=cube)
+        cellset_id = self.create_cellset(mdx=mdx)
+
+        payload = {
+            "BeginOrdinal": 0,
+            "Value": "C",
+            "ReferenceCell@odata.bind": list()}
+        for unique_element_name in unique_element_names:
+            payload["ReferenceCell@odata.bind"].append(
+                odata_escape_single_quotes_in_object_names("Dimensions('{}')/Hierarchies('{}')/Elements('{}')".format(
+                    *Utils.dimension_hierarchy_element_tuple_from_unique_name(unique_element_name))))
+
+        self._post_against_cellset(cellset_id=cellset_id, payload=payload, delete_cellset=True)
 
     @tidy_cellset
     def _post_against_cellset(self, cellset_id, payload, **kwargs):
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -11,7 +11,7 @@ if sys.version[0] == '2':
 else:
     import http.client as http_client
 
-REGEX_OBJECT_NAMES = re.compile(r"(?<=[a-zA-Z0-9_])'(?=[a-zA-Z0-9_])")
+REGEX_OBJECT_NAMES = re.compile(r"(?<!\()(?<!eq )'(?!\)|\Z)")
 
 
 def get_all_servers_from_adminhost(adminhost='localhost'):
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -11,7 +11,7 @@ if sys.version[0] == '2':
 else:
     import http.client as http_client
 
-REGEX_OBJECT_NAMES = re.compile(r"(?<=\(').*?(?='\))")
+REGEX_OBJECT_NAMES = re.compile(r"(?<=[A-Za-z])'(?=[A-Za-z])")
 
 
 def get_all_servers_from_adminhost(adminhost='localhost'):
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -11,7 +11,7 @@ if sys.version[0] == '2':
 else:
     import http.client as http_client
 
-REGEX_OBJECT_NAMES = re.compile(r"(?<=[A-Za-z])'(?=[A-Za-z])")
+REGEX_OBJECT_NAMES = re.compile(r"(?<=[a-zA-Z0-9_])'(?=[a-zA-Z0-9_])")
 
 
 def get_all_servers_from_adminhost(adminhost='localhost'):
--- a/Tests/Utils.py
+++ b/Tests/Utils.py
@@ -660,10 +660,35 @@ class TestMDXUtils(unittest.TestCase):
 
     def test_odata_escape_single_quotes_in_object_names(self):
         url = "https://localhost:8099/api/v1/Dimensions('dime'nsion')/Hierarchies('hier'archy')/Elements('elem'ent')"
+        url1 = "https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'Test C'ase cube'"
+        url2 = "https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'TestC_'ase cube'"
+        ulr3 = "https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'Test C9'ase cube'"
+        ulr4 = "https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'Test C9'_ase cube'"
+        ulr5 = "https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'Test Case cube'"
         escaped_url = Utils.odata_escape_single_quotes_in_object_names(url)
+        escaped_url1 = Utils.odata_escape_single_quotes_in_object_names(url1)
+        escaped_url2 = Utils.odata_escape_single_quotes_in_object_names(url2)
+        escaped_url3 = Utils.odata_escape_single_quotes_in_object_names(ulr3)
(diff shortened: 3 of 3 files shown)


MariusWirtz:
Hello @rkvinoth,

thanks for the MR. I like it!

I didn't get the connection to #202  though. Can you please elaborate a bit on that?

I don't fully understand the change on the regular expression. 
Now it would fail to escape a single quote in an element name that has an underscore or a digit before the single quote. 
Ideally you add a new test case here that shows what the change aims to accomplish:
https://github.com/cubewise-code/tm1py/blob/61f83761434c64c0dd6385390561c54301d627a2/Tests/Utils.py#L696

rkvinoth:
Hi Marius,

The previous regular expression was parsing the input URL as follows:

**Input1:**
`https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'Test C'ase cube'`
**Output1:**
`https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq ''Test C''ase cube''`

**Input2:**
`https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'Test Case cube'`
**Output2:**
`https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq ''Test Case cube''`

I have updated the REGEX such that, it will only consider the single quote if it is beginning or ending with characters, digits or underscore(_).

Let me know if I'm missing anything else. I have updated the test case with new examples.

MariusWirtz:
Hi @rkvinoth ,

thanks for the update and the test cases. Very much appreciated!

When I run for the first input url against the current master branch of TM1py I get
`https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'Test C'ase cube'`
as a result. When I run the second input url I get 
`https://localhost:915/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=Cube eq 'Test Case cube'`
as a result.

So the second result is fine IMO. However, the first ouput is problematic. I agree with you. 
When I said earlier that digits and underscores needed to be escaped as well, I wasn't precise enough. Apologies! 
The problem is that object names in TM1 can have almost every character in them. What characters are allowed in object names is discussed in more depth here:
https://cubewise.com/blog/naming-convention-rules-object-element-attribute/
I think the regex needs to be adjust to cater for this.

rkvinoth:
Hey Marius,

Apologies I got held up with something. I have corrected the regular expression as below.

`re.compile(r"(?<!\()(?<!eq )'(?!\)|\Z)").sub(string=url, repl="''")`

`'` in the string **url** will be turned to `''` if they are preceded with `(` or `eq `
`'` in the string **url** will be turned to `''` if they are suffixed with `)` or `end of string`

Though the `eq ` is just a work around.. I don't find a scenario where it will go useless.

Let me know your thoughts on this one.

MariusWirtz:
Hi @rkvinoth,

thanks a lot for the update on this one. I think it's important that we address this problem in TM1py. The regex is definitely more advanced than what we currently have in there. I will merge accept the MR.

However, even this advanced regex doesn't cover all cases. Instead of `eq` we could have `ne`, `lt` or `gt`.
Also when we use `contains`, `startwith` or `endswith` filters in the URL, the regex wouldn't do what we need it to do:

`/api/v1/TransactionLogEntries?$orderby=TimeStamp desc &$filter=endswith(Cube,'something')`


Perhaps we need to take a step back. Perhaps a regex on the URL is not the right approach.
I tried to address the problem today with a decorator that we could add to the methods that take tm1-object-names as arguments.

The preferred solution that I wanted to build today is this:

``` python
@escape_arguments("user", "cube")
def get_transaction_log_entries(self, reverse=True, user=None, cube=None, since=None, top=None):
```

Then the decorator would escape the potential single quote in the specified arguments before passing it to the `get_transaction_log_entries` method.
Unfortunately, I didn't manage to make this one work. Inside the decorator function, I wasn't able to comprehend which arguments or keyword arguments are user and cube (because arguments could be passed as arguments or keyword arguments).

So in the second iteration, I wrote this function (you find the code below):

``` python
@escape_string_arguments
def get_transaction_log_entries(self, reverse=True, user=None, cube=None, since=None, top=None):
```

This decorator function is much simpler. It iterates through all keyword arguments and arguments, that are passed to the `get_transaction_log_entries` method. 
If any argument is a string, the decorator will replace `'` with `''` before passing the arguments on to `get_transaction_log_entries`.

The downside of a decorator approach would be, that we need to add this decorator to most functions in TM1py.
Another downside would be that potentially in some function we may have string arguments that should be escaped and also string arguments that should not be escaped, so for those functions, we would need a separate solution.

What do you think? 
Do you think we should make the regex smarter? Or should we abandon the regex approach and add escaping decorators to all relevant functions instead?

``` python
def escape_string_arguments(func):
    """ Higher Order function to escape ' with '' in string arguments

    :return:
    """

    @functools.wraps(func)
    def wrapper(self, *args, **kwargs):
        # escape
        args = tuple(arg.replace("'", "''") if isinstance(arg, str) else arg
                     for arg
                     in args)
        kwargs = {key: value.replace("'", "''") if isinstance(value, str) else value
                  for key, value
                  in kwargs.items()}

        # call actual method
        response = func(self, *args, **kwargs)
        return response

    return wrapper
```

MariusWirtz:
Hi  @rkvinoth ,

just noticed that this MR broke one of the Tests: `test_odata_escape_single_quotes_in_object_names_custom_request_threads`

TM1py should not adjust the single quotes in this URL: 
`"https://localhost:8099/api/v1/Threads?$top=0&$filter=ObjectType eq 'Process' and ObjectName ne 'Process - Get Params REST'&$count=true`

I believe the regex approach is not the most promising. I will create a MR to move the escaping of the single quotes into the function layer, probably through a decorator on the function.

rkvinoth:
Hey Marius,

Thanks for looking into the fix in detail and come up with the possible issues. Yes, I agree that the fix is not still solving all the possible scenarios.
Seeing the scenarios you brought up, I believe REGEX is not the right way to proceed further.

As mentioned we can start with a decorator and I will try to check if we have any other possible solution.

rclapp:
Since this isn't a unique problem, we may want to look at how other Odata libraries are handling advanced filters.

https://www.odata.org/libraries/
Ryan Clapp
AWS Business Systems
Sr. Manager

Sent from my mobile device



On Mar 23, 2020 9:07 PM, Vinoth Kumar Ravi <notifications@github.com> wrote:

Hey Marius,

Thanks for looking into the fix in detail and come up with the possible issues. Yes, I agree that the fix is not still solving all the possible scenarios.
Seeing the scenarios you brought up, I believe REGEX is not the right way to proceed further.

As mentioned we can start with a decorator and I will try to check if we have any other possible solution.

-

MariusWirtz:
Hi @rkvinoth, Hi @rclapp,

actually I think there is a fairly easy way out, that doesn't involve regex or decorators.

when building the URL, instead of using plain string formatting like:
``` python
url = "/api/v1/Processes('{}')/tm1.ExecuteWithReturn?$expand=*".format(process_name)
```

we could call a utility method to build the URL and escape single quotes in the arguments:
``` python
url = format_url("/api/v1/Processes('{}')/tm1.ExecuteWithReturn?$expand=*", process_name)
```

This should do the trick:
``` python
def format_url(url, *args, **kwargs):
    """ build url and escape single quotes in args and kwargs

    :param url: url with {} placeholders
    :param args: arguments to placeholders
    :return:
    """
    args = [arg.replace("'", "''") if isinstance(arg, str) else arg
            for arg
            in args]

    kwargs = {key: value.replace("'", "''") if isinstance(value, str) else value
              for key, value
              in kwargs.items()}

    return url.format(*args, **kwargs)
```

The only downside is, I would have to adjust all methods in TM1py. 
However, that would be a good opportunity to add type hints and challenge the docstrings...
What do you guys think?

rclapp:
I like it! Seems like a good way to add some features and get these changes in place.

Thanks,

Ryan Clapp

rkvinoth:
> Hi @rkvinoth, Hi @rclapp,
> 
> actually I think there is a fairly easy way out, that doesn't involve regex or decorators.
> 
> when building the URL, instead of using plain string formatting like:
> 
> ```python
> url = "/api/v1/Processes('{}')/tm1.ExecuteWithReturn?$expand=*".format(process_name)
> ```
> 
> we could call a utility method to build the URL and escape single quotes in the arguments:
> 
> ```python
> url = format_url("/api/v1/Processes('{}')/tm1.ExecuteWithReturn?$expand=*", process_name)
> ```
> 
> This should do the trick:
> 
> ```python
> def format_url(url, *args, **kwargs):
>     """ build url and escape single quotes in args and kwargs
> 
>     :param url: url with {} placeholders
>     :param args: arguments to placeholders
>     :return:
>     """
>     args = [arg.replace("'", "''") if isinstance(arg, str) else arg
>             for arg
>             in args]
> 
>     kwargs = {key: value.replace("'", "''") if isinstance(value, str) else value
>               for key, value
>               in kwargs.items()}
> 
>     return url.format(*args, **kwargs)
> ```
> 
> The only downside is, I would have to adjust all methods in TM1py.
> However, that would be a good opportunity to add type hints and challenge the docstrings...
> What do you guys think?

This is a good opportunity to incorporate any smaller changes you may have in your bucket list.

I had something like this in mind but was skeptical about the scale of changes and tests this would involve.

scrambldchannel:
For what it's worth, I like this idea too, happy to help with re-factoring and testing, I suspect I might have some free time over the next few weeks....

MariusWirtz:
@scrambldchannel 
Thank you! That would be great. I will start working on it in the coming days. I will make sure to keep my remote branch #212  up to date. I am planning to refactor the services one by one alphabetically, starting from the top. If you are keen, you could start from the bottom.

If you would like to contribute to TM1py differently, you would also be more than welcome. The tests could really use some refactoring. It would be better if our test classes were mapping one to one onto the original classes. E.g. have a test class for `cube` and another one for `CubeService`.

Also, the more recent version of the REST API supports batch requests. It would be cool (in particular for performance) if we could group requests together, and then send them as a batch.

PR #211, opened 2020-03-19
A couple of small fixes for tests

scrambldchannel:
Hi 

Thanks for tm1py. I've been doing a bit of cross-platform testing and found a couple of little fixes to how paths are managed. Let me know what you think.

I'm still getting a few failures when running on 3.7.x on windows and Linux but think they're mostly down to issues with the server I'm testing against. I might try to refactor the tests a bit as I go.

Cheers
Alex

### DIFF
--- a/Tests/Annotation.py
+++ b/Tests/Annotation.py
@@ -1,6 +1,6 @@
 import configparser
 import json
-import os
+from pathlib import Path
 import random
 import string
 import unittest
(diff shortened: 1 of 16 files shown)


rclapp:
We may want to explore moving to PathLib as it makes path management much easier compared to the old os libraries.

https://docs.python.org/3/library/pathlib.html

Thanks,

Ryan Clapp

scrambldchannel:
Cool, will take a look at it, happy to contribute.

scrambldchannel:
I added another commit to convert to using pathlib

MariusWirtz:
Hi @scrambldchannel,

thanks for the contribution! I like it.

PR #215, opened 2020-03-30
Allow custom string encoding for cell data requests

MariusWirtz:
- `http_method` decorator now supports custom encoding (e.g. `'latin-1'`). Default value is `'utf-8'`
- common read and write methods in `CellService` accept `encoding` argument through `**kwargs` and pass it down to the http call


MariusWirtz:
Hi @pbuncik,

please check if this solves your issue.

PR #217, opened 2020-04-01
Cast the timeout dict value to int

ykud:
Configparser returns values as string by default, so using timeout in server configuration in the config file will throw an type error in urllib3, with timeout being neither int or float, like this:
  File "site-packages\urllib3\util\timeout.py", line 140, in _validate_timeout
TypeError: '<=' not supported between instances of 'str' and 'int'


MariusWirtz:
@ykud 
Thank you!

PR #227, opened 2020-04-24
add enabled and usertype property to User

rclapp:
Add enabled property to user object. Updated supporting init methods

### DIFF
--- a/TM1py/Objects/User.py
+++ b/TM1py/Objects/User.py
@@ -3,10 +3,17 @@
 import collections
 import json
 from typing import Iterable, Optional, List, Dict
+from enum import Enum
 
 from TM1py.Objects.TM1Object import TM1Object
 from TM1py.Utils.Utils import CaseAndSpaceInsensitiveSet, format_url
 
+class UserType(Enum):
+    User = 0
+    SecurityAdmin = 1
+    DataAdmin = 2
+    Admin = 3
+    OperationsAdmin = 4
 
 class User(TM1Object):
     """ Abstraction of a TM1 User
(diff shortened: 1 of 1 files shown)


scrumthing:
Awesome but as afar as I can see you made the admin property Boolean. Actually you get the different types of admins (admin, security admin, data admin, operations admin) via the api. So IMHO making it Boolean would mean to loose some information. What do you think?

rclapp:
Admin is unchanged, from the way it was originally implemented.

I added enabled property.

However admin as a bool makes some sense since you can get the admin detail from the group membership.

Ryan Clapp
AWS Business Systems
Sr. Manager

Sent from my mobile device

On Apr 23, 2020 8:06 PM, Christoph Hein <notifications@github.com> wrote:

Awesome but as afar as I can see you made the admin property Boolean. Actually you get the different types of admins (admin, security admin, data admin, operations admin) via the api. So IMHO making it Boolean would mean to loose some information. What do you think?

-

rclapp:
@scrumthing I added a user_type property that is an ENUM that matches how it is expressed on the TM1 server.

MariusWirtz:
closed in favor of https://github.com/cubewise-code/tm1py/pull/228

@rclapp 
I added one commit on top of yours and created a new MR from there.
I noticed that if we only change the `user_type` property and send the user back to TM1 it does not apply in TM1. So we need to change the user's groups behind the scenes if someone changes the `user_type` property.
I also made the `UserType` Emun case and space insensitive and added some test cases.
Could you double-check my changes please?

@scrumthing 
Could you check if it works as expected? You can install the branch with pip like this:
`pip install https://github.com/cubewise-code/tm1py/archive/feature/new-user-properties.zip --upgrade
`

scrumthing:
Thanks guys. It took me a while to get back to that topic. Stupid daily business...
Finally tried it today and it worked like a charm. Thank you! :-)

PR #228, opened 2020-04-24
add enabled and usertype property to User

MariusWirtz:
New properties in `User` class: `user_type`, `enabled`


rclapp:
Thanks @MariusWirtz. Thanks for building the testing for it too!

PR #232, opened 2020-05-03
Add clear_with_mdx method to CellService

MariusWirtz:
Add `clear_with_mdx` method to CellService

``` python
with TM1Service(address=ADDRESS, port=PORT, user=USER, password=PWD, ssl=SSL) as tm1:
    mdx = """
    SELECT
    {Tm1SubsetAll([Dimension])} ON 0,
    {[CubeMeasure].[Measure]} ON 1
    FROM [Cube]
    WHERE ([Period].[202003])
    """

    tm1.cells.clear_with_mdx(cube="Cube", mdx=mdx)
```

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -6,17 +6,18 @@ import uuid
 import warnings
 from collections import OrderedDict
 from io import StringIO
-from typing import List, Union, Dict, Iterable, Tuple
+from typing import List, Union, Dict, Iterable
 
 import pandas as pd
 from requests import Response
 
+from TM1py.Exceptions.Exceptions import TM1pyBaseException
 from TM1py.Objects.Process import Process
 from TM1py.Services.ObjectService import ObjectService
 from TM1py.Services.RestService import RestService
 from TM1py.Utils import Utils, CaseAndSpaceInsensitiveSet, format_url
 from TM1py.Utils.Utils import build_pandas_dataframe_from_cellset, dimension_name_from_element_unique_name, \
-    CaseAndSpaceInsensitiveTuplesDict
+    CaseAndSpaceInsensitiveTuplesDict, abbreviate_mdx
 
 
 def tidy_cellset(func):
@@ -158,7 +159,7 @@ class CellService(ObjectService):
 
         return self._post_against_cellset(cellset_id=cellset_id, payload=payload, delete_cellset=True, **kwargs)
 
-    def clear_with_mdx(self, cube: str, mdx: str, **kwargs) -> Tuple[bool, str, str]:
+    def clear_with_mdx(self, cube: str, mdx: str, **kwargs):
         view_name = "".join(['}TM1py', str(uuid.uuid4())])
         code = "".join([
             f"ViewCreateByMdx('{cube}','{view_name}','{mdx}',1);",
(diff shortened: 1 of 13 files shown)


TM1py/Services/CellService.py
```diff
@@ -156,6 +157,16 @@ def clear_spread(
...
 
+    def clear_with_mdx(self, cube: str, mdx: str, **kwargs) -> Response:
+        view_name = "".join(['}TM1py', str(uuid.uuid4())])
+        ti = [
+            f"ViewCreateByMdx('{cube}','{view_name}','{mdx}',1);",
+            f"ViewZeroOut('{cube}','{view_name}');"]
+
+        from TM1py import ProcessService
+        process_service = ProcessService(self._rest)
+        return process_service.execute_ti_code(lines_prolog=ti, **kwargs)
```
rclapp, line 168: I suggest using execute_process_with_return instead. It eliminates dependencies on uuid and removes the burden of creating and deleting a TI on the server, which sometimes can be prone to lock contention.

MariusWirtz: Good point! Thanks for the catch.

TM1py/Services/CellService.py
```diff
@@ -156,6 +158,18 @@ def clear_spread(
...
+        view_name = "".join(['}TM1py', str(uuid.uuid4())])
+        code = "".join([
+            f"ViewCreateByMdx('{cube}','{view_name}','{mdx}',1);",
+            f"ViewZeroOut('{cube}','{view_name}');"])
+        process = Process(name="")
+        process.prolog_procedure = code
+
+        from TM1py import ProcessService
+        process_service = ProcessService(self._rest)
+        return process_service.execute_process_with_return(process, **kwargs)
```
MariusWirtz, line 171: I think in order to be compliant with the rest of TM1py it would be better to raise a `TM1pyException`, when the clear fails, instead of returning the result tuple.
@rclapp what do you think?

rclapp:
@MariusWirtz I would agree. Using TI this way is really a work around for not having a true rest based zero out function. A user shouldn't have to know that an unbound TI is being used to complete the 0.

I would also assume that executing that mdx and patching all zeros over it would be slower?

MariusWirtz:
Yes. Patching all zeros is slower than calling a ViewZeroOut process. 
This new method should bring speed and flexibility together 🤗

Agree. This is actually a workaround. Once TM1 exposes a ViewZeroOut call through the REST API we could refactor the method while the arguments and the potentially raised Exception would stay the same.

vicmora:
I tried using this but receive a "view does not exist" error message from the server. We can't use ViewZeroOut with an MDX view unless we're on 2.0.9 right?

https://www.ibm.com/support/knowledgecenter/SSD29G_2.0.0/com.ibm.swg.ba.cognos.tm1_nfg.2.0.0.doc/pa_nf_viewzeroout_support.html

MariusWirtz:
Yes.

@vicmora 
do you think TM1py should throw a dedicated, descriptive error if a function is not supported on your TM1 instance, due to your TM1 version? 
I do think we should build this into TM1py. At least for the functions that use recently added functionality.

vicmora:
Thanks @MariusWirtz! I think that would be really helpful.

PR #235, opened 2020-05-05
Get the Percent change in memory

rkvinoth:
Cube dimension re-order can now return the percent change in memory footprint.

### DIFF
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -160,7 +160,7 @@ class CubeService(ObjectService):
         response = self._rest.GET(url, **kwargs)
         return [dimension["Name"] for dimension in response.json()["value"]]
 
-    def update_storage_dimension_order(self, cube_name: str, dimension_names: Iterable[str]) -> Float[str]:
+    def update_storage_dimension_order(self, cube_name: str, dimension_names: Iterable[str]) -> Float:
         """ Update the storage dimension order of a cube
 
         :param cube_name:
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -160,12 +160,12 @@ class CubeService(ObjectService):
         response = self._rest.GET(url, **kwargs)
         return [dimension["Name"] for dimension in response.json()["value"]]
 
-    def update_storage_dimension_order(self, cube_name: str, dimension_names: Iterable[str]) -> Response:
+    def update_storage_dimension_order(self, cube_name: str, dimension_names: Iterable[str]) -> Float[str]:
         """ Update the storage dimension order of a cube
 
         :param cube_name:
         :param dimension_names:
-        :return: percent change in memory usage
+        :return:  Float: -23.076489699337078 (percent change in memory usage)
         """
         url = format_url("/api/v1/Cubes('{}')/tm1.ReorderDimensions", cube_name)
         payload = dict()
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -160,7 +160,7 @@ class CubeService(ObjectService):
         response = self._rest.GET(url, **kwargs)
         return [dimension["Name"] for dimension in response.json()["value"]]
 
-    def update_storage_dimension_order(self, cube_name: str, dimension_names: Iterable[str]) -> Float:
+    def update_storage_dimension_order(self, cube_name: str, dimension_names: Iterable[str]) -> float:
         """ Update the storage dimension order of a cube
 
         :param cube_name:
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -165,14 +165,15 @@ class CubeService(ObjectService):
 
         :param cube_name:
         :param dimension_names:
-        :return:
+        :return: percent change in memory usage
         """
         url = format_url("/api/v1/Cubes('{}')/tm1.ReorderDimensions", cube_name)
         payload = dict()
         payload['Dimensions@odata.bind'] = [format_url("Dimensions('{}')", dimension)
                                             for dimension
                                             in dimension_names]
-        return self._rest.POST(url=url, data=json.dumps(payload))
+        response = self._rest.POST(request=url, data=json.dumps(payload))
+        return response.json()["value"]
 
     def load(self, cube_name: str, **kwargs) -> Response:
         """ Load the cube into memory on the server


MariusWirtz:
Nice one! I didn't know TM1 would return this.
Please update the return annotation of the function. I guess it should say `-> float` now.

rkvinoth:
Was working on it and surprised to see that there was no way to check the change in memory.
Quickly gone through the REST API doc and found that TM1 is actually giving a _ReturnType_ for **ReorderDimensions**

Hope this will help someone!

MariusWirtz:
Thanks for the update. 

I am not sure I understand: `str[Float]`. 
The `response.json()['value']` would return either an `int` or a `float` I think. So I think it should say `-> float` for the return type since float implies int.
https://stackoverflow.com/questions/50928592/mypy-type-hint-unionfloat-int-is-there-a-number-type

MariusWirtz:
I am actually working on a hobby project to find "the right" dimension order for a cube. This change will be very helpful for me 🤗

rkvinoth:
> Thanks for the update.
> 
> I am not sure I understand: `str[Float]`.
> The `response.json()['value']` would return either an `int` or a `float` I think. So I think it should say `-> float` for the return type since float implies int.
> https://stackoverflow.com/questions/50928592/mypy-type-hint-unionfloat-int-is-there-a-number-type

You are right, updated it as `Float`

MariusWirtz:
I think it needs to be `float` instead of `Float`

rkvinoth:
> I am actually working on a hobby project to find "the right" dimension order for a cube. This change will be very helpful for me 🤗

I have also been working on the exact thing, I believe it will improve the performance at an unimaginable level.
But I crossed a weird issue, the memory footprint we gain actually disappears post server restart.

**specifics**: 
cube size: 17GB 
PA version: 11.4.00009.4

**Steps performed**:
1. re-ordered dimension and brought the cube down to 2GB
2. Executed Savedata
3. Restart Server

Now the cube again occupying 17GB. Surprise the dimension order is in the way I changed it.

I have the below _APAR_:


**PH25056 DIMENSION REORDERING NOT WORKING AS EXPECTED**

rkvinoth:
> I think it needs to be `float` instead of `Float`

Other methods had `List` instead of `list`
So, I used `Float` instead of `float`

Do you still want it `float`?

MariusWirtz:
> > I think it needs to be `float` instead of `Float`
> 
> Other methods had `List` instead of `list`
> So, I used `Float` instead of `float`
> 
> Do you still want it `float`?

For primitive types you have to write the hint like `-> str`, `-> int`, `-> bool`, `-> float`. For the collections like lists and sets you have to import the type hint from the `typing` module: `from typing import Dict, Tuple, List`. 
For custom classes like Subset you can write `-> Subset`

I am not entirely sure why they implemented it that way.

MariusWirtz:
> > I am actually working on a hobby project to find "the right" dimension order for a cube. This change will be very helpful for me 🤗
> 
> I have also been working on the exact thing, I believe it will improve the performance at an unimaginable level.
> But I crossed a weird issue, the memory footprint we gain actually disappears post server restart.
> 
> **specifics**:
> cube size: 17GB
> PA version: 11.4.00009.4
> 
> **Steps performed**:
> 
> 1. re-ordered dimension and brought the cube down to 2GB
> 2. Executed Savedata
> 3. Restart Server
> 
> Now the cube again occupying 17GB. Surprise the dimension order is in the way I changed it.
> 
> I have the below _APAR_:
> 
> **PH25056 DIMENSION REORDERING NOT WORKING AS EXPECTED**

Yeah, that's weird.
Perhaps we can combine our efforts at some point. I'm still trying to find the right logic to come up with the "ideal" order. You find the WIP state here: https://github.com/MariusWirtz/tm1-optipyzer

rkvinoth:
> > > I think it needs to be `float` instead of `Float`
> > 
> > 
> > Other methods had `List` instead of `list`
> > So, I used `Float` instead of `float`
> > Do you still want it `float`?
> 
> For primitive types you have to write the hint like `-> str`, `-> int`, `-> bool`, `-> float`. For the collections like lists and sets you have to import the type hint from the `typing` module: `from typing import Dict, Tuple, List`.
> For custom classes like Subset you can write `-> Subset`
> 
> I am not entirely sure why they implemented it that way.

thanks for making things clear as always!
Updated the Annotation, feel free to let me know if any other update is required.

rclapp:
There is a known bug that the memory used post reorder is wrong after you reorder once. IBM has said they will fix it.  Once fixed you can just have python iterate some possible dim orders. We have a script that reorders, runs a TI, refreshes a view, and logs all the results. Makes finding the best order for the job much easier.




On May 5, 2020 5:15 AM, Vinoth Kumar Ravi <notifications@github.com> wrote:

I think it needs to be float instead of Float

Other methods had List instead of list
So, I used Float instead of float
Do you still want it float?

For primitive types you have to write the hint like -> str, -> int, -> bool, -> float. For the collections like lists and sets you have to import the type hint from the typing module: from typing import Dict, Tuple, List.
For custom classes like Subset you can write -> Subset

I am not entirely sure why they implemented it that way.

thanks for making things clear as always!
Updated the Annotation, feel free to let me know if any other update is required.

-

MariusWirtz:
Hi @rclapp,
Yes, that makes it easier. 
Do you guys also evaluate the TI-write-speed that a cube dimension order provides?
I think part of the difficulty is to generate possible dimension-orders that are to be evaluated. How do you address this?

rclapp:
3 tests
1) TI write Speed
2) MDX view no rules speed
3) MDX view with rules speed

Then I either specify a few permutations to test or have it do some at random. Most impact full is usually swapping the first and last dimension.



Sent from my mobile device

On May 5, 2020 2:04 PM, Marius Wirtz <notifications@github.com> wrote:

Hi @rclapp<https://github.com/rclapp>,
Yes, that makes it easier.
Do you guys also evaluate the TI-write-speed that a cube dimension order provides?
I think part of the difficulty is to generate possible dimension-orders that are to be evaluated. How do you address this?

-

MariusWirtz:
Hi @rkvinoth, Hi @rclapp,

I am planning to continue my efforts to find the ideal dimension order with TM1py. 
Has the PH25056 issue already been addressed by IBM? If yes, could you please let me know which release delivers the fix?

Thank you

Marius

rkvinoth:
Hey Marius,

I had a call with IBMers two weeks back on this one. Explained them about the issue again.
Even the Manager who connected with me is not sure about how this functionality works. He said we will connect with the development team and get back. I will try adding a follow-up today.

One thing he mentioned was that we cannot trust the percentage change in memory that TM1 Architect shows. Then I told him that we are looking at the REST API's return value (he didn't know about this feature).

From the entire call I was thrown an impression that IBM is not going to support this feature anymore. Anyway as mentioned will add another follow-up.

MariusWirtz:
Thanks for update @rkvinoth 

For my test case and my TM1 version (11.7.00002.1), the rest API return value actually maps to the memory that is shown in the `}StatsByCube` cube. For me it looks kinda like this:

|Dimension Order|RAM (}StatsbyCube) |RAM Change (REST) |
|---------------|---|----------|
|1              |100|          |
|2              |200|50%       |
|3              |50 |-75%      |

I am going to trust the REST value for now but take them "with a grain of salt".

rclapp:
Based on my last conversation with IBM 2.0.9.1 has fixed the issue with memory showing negative in the stats cube.

Thanks,

Ryan Clapp

user1493:
Marius,


I just got the below response from IBM team:


Hi Vinoth - This issue is still being investigated at our end,interestingly the behavior observed in PA 2.0.9.1 onward is different than what is being observed in PA 2.0.9 and prior versions.

 

In PA 2.0.9.1 Architect does not showcase any memory difference post reordering dimensions,Development team is reviewing the issue and an update will be shared as soon as some feedback is received.

The REST API is also returning the memory delta value as zero in PA 2.0.9.1 similar to Architect which uses legacy C API

As an interim solution please use TM1 }StatsByCube control cube to observe the delta in memory consumption post dimension reordering.


Thanks
Vinoth Kumar

rkvinoth:
Marius,


I just got the below response from IBM team:


Hi Vinoth - This issue is still being investigated at our end,interestingly the behavior observed in PA 2.0.9.1 onward is different than what is being observed in PA 2.0.9 and prior versions.

 

In PA 2.0.9.1 Architect does not showcase any memory difference post reordering dimensions,Development team is reviewing the issue and an update will be shared as soon as some feedback is received.

The REST API is also returning the memory delta value as zero in PA 2.0.9.1 similar to Architect which uses legacy C API

As an interim solution please use TM1 }StatsByCube control cube to observe the delta in memory consumption post dimension reordering.

PR #252, opened 2020-06-15
wip: Introduce skip argument in all execute methods

MariusWirtz:


MariusWirtz:
Hi @scrumthing,

it turned out to be easier than anticipated. 
You can upgrade your local TM1py to this branch with pip:
`pip install https://github.com/cubewise-code/tm1py/archive/feature/skip-argument-in-execute-functions.zip --upgrade`


Please let me know if this works as expected. I will add a few test cases for the functionality over the next days, but it seems to be working well so far.

scrumthing:
Thanks @MariusWirtz for the quick implementation. Will check it out tomorrow. 
With skip and top proper phonation (within the odata standard) should be possible. Awesome!

scrumthing:
It works as expected. If you set top to 100 and skip to 0 you get the first 100 entries. If you set it top to 100 and skip to 100 you get the entries 101 till 200.
But just to clarify. At the moment for both request a new cellset is created. Wouldn't it be more efficient if we could reuse the cellset for the second batch? Because in my test enviroment it takes 76 seconds to create the cellset every time.

MariusWirtz:
Hi @scrumthing,

thanks for testing!
You are right. Every `execute_view` call creates a cellset with the full view, but retrieves only n cells.
That is indeed inefficient.

With the CellService you can also operate on the cellset level. 
I think is what you need. Here is the sample:

``` python
from TM1py import TM1Service

ADDRESS = "localhost"
PORT = 12354
SSL = True
USER = "admin"
PASSWORD = "apple"

    view = tm1.cubes.views.get("Sales", "Margin By State", False)

    cellset_id = tm1.cells.create_cellset_from_view(cube_name="Sales", view_name="Margin By State", private=False)

    cells = tm1.cells.extract_cellset(cellset_id=cellset_id, top=10, skip=0, skip_contexts=True, delete_cellset=False)

    cells = tm1.cells.extract_cellset(cellset_id=cellset_id, top=10, skip=10, skip_contexts=True, delete_cellset=False)

    cells = tm1.cells.extract_cellset(cellset_id=cellset_id, top=10, skip=20, skip_contexts=True, delete_cellset=False)
```

PS
don't worry about deleting the cellset. If you log out (leave the `with` context in python) the REST API will destroy your cellsets.

scrumthing:
Thanks. Thats works great! :-)

PR #259, opened 2020-06-26
Improving build_cellset_from_pandas_dataframe method

rkvinoth:
Instead of getting the list via dict(orient), getting it via numpy improves the efficiency by 50%.

df.shape = (106920, 12)
new method took - 0.9022 seconds
old method took - 1.7175 seconds


MariusWirtz:
Nice one!

PR #261, opened 2020-07-01
Method to convert the Duration to seconds

rkvinoth:
REST API returned duration like this {'ElapsedTime': 'P0DT00H02M16S', 'WaitTime': 'P0DT00H00M00S'} can be converted to seconds (136).

```
import pandas as pd
import re

def get_seconds_from_duration(time_str: str) -> int:
    """
    This function will convert the TM1 time to seconds
    :param time_str: P0DT00H01M43S
    :return: int
    """
    pattern = re.compile('\w(\d+)\w\w(\d+)\w(\d+)\w(\d+)\w')
    matches = pattern.search(time_str)
    d, h, m, s = matches.groups()
    seconds = (int(d) * 86400) + (int(h) * 3600) + (int(m) * 60) + int(s)
    return seconds

    threads = pd.DataFrame(tm1.monitoring.get_threads())
    threads[['ElapsedTime', 'WaitTime']] = threads[['ElapsedTime', 'WaitTime']].applymap(get_seconds_from_duration)
    print(threads[['ElapsedTime', 'WaitTime']])
```

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -401,6 +401,20 @@ def lower_and_drop_spaces(item: str) -> str:
     return item.replace(" ", "").lower()
 
 
+def get_seconds_from_duration(time_str: str) -> int:
+    """
+    This function will convert the TM1 time to seconds
+    :param time_str: P0DT00H01M43S
+    :return: int
+    """
+    import re
+    pattern = re.compile('\w(\d+)\w\w(\d+)\w(\d+)\w(\d+)\w')
+    matches = pattern.search(time_str)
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Hi @rkvinoth,

I like it. That will be useful. Thank you. 
Please add a small test case for the function in the Tests/utils.py module. Then I will merge.

rkvinoth:
> Hi @rkvinoth,
> 
> I like it. That will be useful. Thank you.
> Please add a small test case for the function in the Tests/utils.py module. Then I will merge.

I'm using this myself for cancelling long running user threads, just added a test case.

PR #272, opened 2020-07-07
Only apply UTC time if not tz info provided

MariusWirtz:
allows filtering log when log time is not UTC


rkvinoth:
shall we add `until` argument to `get_transaction_log_entries` method as well?

MariusWirtz:
Yes. That could be useful.

PR #279, opened 2020-07-12
feature/remove-content-hack

MariusWirtz:
Add `skip_zeros`, `skip_consolidated`, `skip_rulederived` to all relevant execute_mdx / execute_view functions

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -309,8 +309,8 @@ class CellService(ObjectService):
 
         :param mdx: MDX Query, as string
         :param cell_properties: properties to be queried from the cell. E.g. Value, Ordinal, RuleDerived, ... 
-        :param top: integer
-        :param skip: integer
+        :param top: Int, number of cells to return (counting from top)
+        :param skip: Int, number of cells to skip (counting from top)
         :param skip_contexts: skip elements from titles / contexts in response
         :param skip_zeros: skip zeros in cellset (irrespective of zero suppression in MDX / view)
         :param skip_consolidated_cells: skip consolidated cells in cellset
@@ -330,15 +330,16 @@ class CellService(ObjectService):
             delete_cellset=True,
             **kwargs)
 
-    def execute_view(self, cube_name: str, view_name: str, cell_properties: Iterable[str] = None, private: bool = False,
+    def execute_view(self, cube_name: str, view_name: str, private: bool = False, cell_properties: Iterable[str] = None,
                      top: int = None, skip_contexts: bool = False, skip: int = None, skip_zeros: bool = False,
                      skip_consolidated_cells: bool = False, skip_rule_derived_cells: bool = False,
                      **kwargs) -> CaseAndSpaceInsensitiveTuplesDict:
         """ get view content as dictionary with sweet and concise structure.
             Works on NativeView and MDXView !
 
-        :param cube_name: String
-        :param view_name: String
+        :param cube_name: String, name of the cube
+        :param view_name: String, name of the view
+        :param private: True (private) or False (public)
         :param cell_properties: List, cell properties: [Values, Status, HasPicklist, etc.]
         :param private: Boolean
         :param top: Int, number of cells to return (counting from top)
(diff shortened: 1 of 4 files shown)


TM1py/Services/CellService.py
```diff
@@ -950,25 +1019,53 @@ def extract_cellset_cellcount(self, cellset_id: str, **kwargs) -> int:
...
+
         :param kwargs:
         :return:
         """
-        raw_csv = self.extract_cellset_csv(cellset_id=cellset_id, delete_cellset=True, **kwargs)
+        raw_csv = self.extract_cellset_csv(cellset_id=cellset_id, skip_zeros=skip_zeros,
+                                           skip_rule_derived_cells=skip_rule_derived_cells,
+                                           skip_consolidated_cells=skip_consolidated_cells,
+                                           **kwargs)
         if not raw_csv:
```
rkvinoth, line 1069: If the `cellset` has no data, then the `raw_csv` string has header part in it.. Ideally it will not be empty and will not get into the empty `Dataframe` part. So, lets add a `new line` check to identify existence of data lines. 

`if "\n" not in raw_csv:`

MariusWirtz:
@cubewise-gng,

I think this one is done. 
All relevant `execute_` functions now accept `skip_consolidated_cells` and `skip_rule_derived_cells` as arguments. You can also control explicitly if zero cells (irrespective of the zero suppression in the MDX) should be skipped or retrieved. In the previous `execute_mdx_csv` and `execute_mdx_dataframe` they were skipped by default.

Please install this branch with pip and test:
`pip install https://github.com/cubewise-code/tm1py/archive/feature/remove-content-hack.zip --upgrade`

```

with TM1Service(address="localhost", port=12354, ssl=True, user="admin", password="apple") as tm1:
    mdx = """
    SELECT
    HEAD({Tm1SubsetAll([dimension_name1])},1000) ON ROWS,
    HEAD({Tm1SubsetAll([dimension_name2])},1000) ON COLUMNS
    FROM [Large Cube]
    """

    csv = tm1.cells.execute_mdx_csv(mdx, skip_zeros=False, skip_consolidated_cells=True, skip_rule_derived_cells=True)

```

TM1py/Services/CellService.py
MariusWirtz: Thanks for the catch! 

Actually, in order to be fully backwards compatible, we should return an empty string if the cellset is completely empty, just like the `/Content` did.
I will adjust the `extract_cellset_csv method` and add a test.

rkvinoth: Yeah, sounds good!

rkvinoth: I think it would be nicer to add the  `skip` Default arguments to `execute_view_dataframe `and `execute_mdx_dataframe` methods as well , for now `Kwargs` works though.

MariusWirtz: Agree. Added all arguments explicitly. Including `top` and `skip`.

TM1py/Services/CellService.py
```diff
@@ -596,12 +607,38 @@ def execute_mdx_cellcount(self, mdx: str, **kwargs) -> int:
...
+        :param skip_zeros: skip zeros in cellset (irrespective of zero suppression in MDX / view)
+        :param skip_consolidated_cells: skip consolidated cells in cellset
+        :param skip_rule_derived_cells: skip rule derived cells in cellset
+        :return: Pandas Dataframe
+        """
+        cellset_id = self.create_cellset_from_view(cube_name=cube_name, view_name=view_name, private=private, top=top,
+                                                   skip=skip, skip_zeros=skip_zeros,
+                                                   skip_consolidated_cells=skip_consolidated_cells,
+                                                   skip_rule_derived_cells=skip_rule_derived_cells, **kwargs)
+        return self.extract_cellset_dataframe(cellset_id, **kwargs)
```
rkvinoth, line 634: Please pass the `skip` arguments to the `extract_cellset_dataframe` instead of `create_cellset_from_view`.

MariusWirtz:
@rkvinoth thanks for the feedback on this MR. Much appreciated.

rkvinoth:
@MariusWirtz  Wow, everything looks very much scalable and reliable. Great work!

Issue #280, opened 2020-07-13
Add until argument to get_transaction_log_entries function


MariusWirtz:
It would be useful if the `get_transaction_log_entries` function in the `ServerService` would allow filtering by `until`

rkvinoth:
@MariusWirtz 

Why don't we close this one?

PR #282, opened 2020-07-14
'until' method for get_transaction_log_entries

rkvinoth:
for https://github.com/cubewise-code/tm1py/issues/280#issue-656164989

### DIFF
--- a/Tests/Server.py
+++ b/Tests/Server.py
@@ -144,7 +144,6 @@ class TestServerMethods(unittest.TestCase):
         regex = re.compile('TM1ProcessError_.*.log')
         self.assertFalse(regex.search(log_entry))
 
-    @unittest.skip("Doesn't work in TM1 11")
     def test_get_last_transaction_log_entries(self):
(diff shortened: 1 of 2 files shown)


Tests/Server.py
```diff
@@ -217,6 +217,22 @@ def test_get_transaction_log_entries_from_today(self):
             entry_date = entry_timestamp.date()
             today_date = datetime.date.today()
             self.assertTrue(entry_date == today_date)
+            
+    @unittest.skip("Doesn't work in TM1 11")
```
MariusWirtz, line 221: I think the `skip` decorator here and the one above can be removed. 
I remember I added for one of the early releases of TM1 11, when the transaction log filtering was broken in TM1.

But now since it works I think we shouldn't skip the test.

rkvinoth: Enabled below three tests:
`test_get_last_transaction_log_entries`
`test_get_transaction_log_entries_from_today`
`test_get_transaction_log_entries_until_yesterday`

PR #287, opened 2020-07-16
Allow custom value separator on csv functions

MariusWirtz:


TM1py/Services/CellService.py
```diff
@@ -1109,7 +1113,7 @@ def extract_cellset_dataframe(
         """
         raw_csv = self.extract_cellset_csv(cellset_id=cellset_id, top=top, skip=skip, skip_zeros=skip_zeros,
                                            skip_rule_derived_cells=skip_rule_derived_cells,
-                                           skip_consolidated_cells=skip_consolidated_cells,
+                                           skip_consolidated_cells=skip_consolidated_cells, value_separator="|",
```
rkvinoth, line 1116: Do you see any performance gain here? If not why don't we stick with comma?

I tried with few views and don't see any difference in performance.

MariusWirtz: I used a different separator because @cubewise-gng found an issue when you use the function on string cells that contain commas

rkvinoth: Ahhh.. makes sense.

PR #290, opened 2020-07-19
Issue/fix create subset with ampersand

MariusWirtz:

PR #291, opened 2020-07-20
Revert 290 issue/fix create subset with ampersand

rkvinoth:
Reverts cubewise-code/tm1py#290

As we have `replace('%', '%25')` in two places `format_url` and `_url_and_body`... `%` becomes `%2525`

```Python
subset = Subset(subset_name='t% eeest', dimension_name='test %', hierarchy_name='test %',
                expression="{TM1FILTERBYLEVEL({TM1SUBSETALL ([test %])}, 0)}")
tm1.subsets.create(subset)
```

URL: `https://adminhost:port/api/v1/Dimensions('test%20%2525')/Hierarchies('test%20%2525')/Subsets`

We should remove `replace` in any one place.

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -33,11 +33,11 @@ def format_url(url, *args: str, **kwargs: str) -> str:
     :param args: arguments to placeholders
     :return:
     """
-    args = [arg.replace("'", "''").replace('%', '%25') if isinstance(arg, str) else arg
+    args = [arg.replace("'", "''") if isinstance(arg, str) else arg
             for arg
             in args]
 
-    kwargs = {key: value.replace("'", "''").replace('%', '%25') if isinstance(value, str) else value
+    kwargs = {key: value.replace("'", "''") if isinstance(value, str) else value
               for key, value
               in kwargs.items()}
 
--- a/Tests/Cube.py
+++ b/Tests/Cube.py
@@ -23,7 +23,7 @@ class TestCubeMethods(unittest.TestCase):
         PREFIX + "dimension3"]
 
     @classmethod
-    def setUp(cls):
+    def setUpClass(cls):
         cls.tm1 = TM1Service(**config['tm1srv01'])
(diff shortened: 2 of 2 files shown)


MariusWirtz:
Hi @rkvinoth,

thanks for the catch! I will rollback the change.

I indeed broke the functionality to create subsets in a dimension with `%` in the name.
However, now it allows creating subsets with elements that contain `%` in the name.
First thing I do now is to create a test case for subset creation with dimensions, subsets, and elements that contain `%`,`#`, `'`

rkvinoth:
> Hi @rkvinoth,
> 
> thanks for the catch! I will rollback the change.
> 
> I indeed broke the functionality to create subsets in a dimension with `%` in the name.
> However, now it allows creating subsets with elements that contain `%` in the name.
> First thing I do now is to create a test case for subset creation with dimensions, subsets, and elements that contain `%`,`#`, `'`

The **main branch** already allows creation of subset/dimension names with `%` in them, including element names with `%`.

I don't see the need for `replace` in `format_url` ( would appreciate if you can give me an example ).

Also, I performed a quick test on this one with multiple special characters.

https://github.com/cubewise-code/tm1py/pull/290#discussion_r457109766

MariusWirtz:
Hi @rkvinoth,

thanks for your comment.

Yes, the master branch does allow that. However, the master branch does not allow to create a subset with elements that have for instance `%AD` in the name.
If you run the test cases from https://github.com/cubewise-code/tm1py/pull/292 against the master branch, you will see the issue. https://github.com/cubewise-code/tm1py/pull/292 should address this. It also moves the escaping to one function in order to follow the DRY principle.
Then in the `RestService`, there is no longer a need to replace any characters in the URL.

rkvinoth:
> Yes, the master branch does allow that. However, the master branch does not allow to create a subset with elements that have for instance `%AD` in the name.

Yeah, makes sense. I was not able to find `&AD` to replicate the issue.

PR #295, opened 2020-07-23
New methods

rkvinoth:
New methods for getting `dict` from `MDX` or `view` created `cellset`.

```Python

tm1 = TM1Service(****)


# MDX Dict
mdx = "SELECT " \
      "NON EMPTY {TM1SUBSETALL( [}Clients] )} on ROWS, " \
      "NON EMPTY {TM1SUBSETALL( [}Groups] )} ON COLUMNS " \
      "FROM [}ClientGroups]"
values = tm1.cells.execute_mdx_dict(mdx)
for key, value in values.items():
    print(f'{key}:{value}')

# View Dict
values = tm1.cells.execute_view_dict(cube_name='}ClientGroups', view_name='Default')
for key, value in values.items():
    print(f'{key}:{value}')
```

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -549,7 +549,7 @@ class CellService(ObjectService):
     def execute_mdx_dict(self, mdx: str, top: int = None, skip: int = None, skip_zeros: bool = True,
                               skip_consolidated_cells: bool = False, skip_rule_derived_cells: bool = False,
                               **kwargs) -> Dict:
-        """ Get Dict from MDX Query.
+        """ Optimized for performance. Get Dict from MDX Query.
 
         :param mdx: Valid MDX Query
         :param top: Int, number of cells to return (counting from top)
@@ -632,7 +632,7 @@ class CellService(ObjectService):
                           skip: int = None, skip_zeros: bool = True, skip_consolidated_cells: bool = False,
                           skip_rule_derived_cells: bool = False, value_separator: str = "|", **kwargs) -> dict:
         """ Optimized for performance. Get a Dict(tuple, value) from an existing Cube View
-        Context dimensions are omitted in the resulting Dataframe !
+        Context dimensions are omitted in the resulting Dictionary !
         Cells with Zero/null are omitted by default, but still configurable!
 
         :param cube_name: String, name of the cube
@@ -1111,7 +1111,7 @@ class CellService(ObjectService):
             skip_rule_derived_cells: bool = False,
             value_separator: str = "|",
             **kwargs) -> dict:
-        """ Execute cellset and return only the 'Content', in csv format
+        """ Execute cellset and return only the 'Content' as a dictionary
 
         :param cellset_id: String; ID of existing cellset
         :param top: Int, number of cells to return (counting from top)
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -18,7 +18,7 @@ from TM1py.Services.RestService import RestService
 from TM1py.Services.ViewService import ViewService
 from TM1py.Utils import Utils, CaseAndSpaceInsensitiveSet, format_url
 from TM1py.Utils.Utils import build_pandas_dataframe_from_cellset, dimension_name_from_element_unique_name, \
-    CaseAndSpaceInsensitiveTuplesDict, abbreviate_mdx, build_csv_from_cellset_dict
+    CaseAndSpaceInsensitiveTuplesDict, abbreviate_mdx, build_csv_from_cellset_dict, build_dict_from_cellset_dict
 
 
 def tidy_cellset(func):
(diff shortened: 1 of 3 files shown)


MariusWirtz:
Hi @rkvinoth,

thanks for the contribution. I see where you are coming from and why the data is useful for you in that shape.
I have two concerns about this MR though. 

1. From the result of the `execute_mdx_csv` function, you could produce the dictionary with one Dict Comprehension. See  below.
In order to avoid additional code, we could perhaps just use the `extract_cellset_csv` inside the new function.

``` python

    # MDX Dict
    mdx = "SELECT " \
          "NON EMPTY {TM1SUBSETALL( [}Clients] )} on COLUMNS, " \
          "NON EMPTY {TM1SUBSETALL( [}Groups] )} ON ROWS " \
          "FROM [}ClientGroups]"
    values1 = tm1.cells.execute_mdx_dict(mdx)

    mdx = "SELECT " \
          "NON EMPTY {TM1SUBSETALL( [}Clients] )} on COLUMNS, " \
          "NON EMPTY {TM1SUBSETALL( [}Groups] )} ON ROWS " \
          "FROM [}ClientGroups]"
    lines = tm1.cells.execute_mdx_csv(mdx, value_separator="|", line_separator="\r\n")
    values2 = {"|".join(entries.split("|")[0:-1]): entries.split("|")[-1]
               for entries
               in lines.split("\r\n")[1:]}

    print(values1 == values2)

>>> True
```

2. I find the name irritating. The classic `execute_mdx` function also returns a dictionary. and the `execute_mdx_rows_and_values` also returns a dictionary. If we do include it, I think it needs a more indicative name.

What do you think about the 2 points? Do you think my suggestion in 1. would be too inefficient?

rkvinoth:
Hi @MariusWirtz,

1. I tried testing them with a cube view with **834932** number of cells, the results were almost same (1 or 2 seconds difference).
    I'm okay this approach, it would reduce many lines of code. Converting a `Dataframe` to dict is _very costly_ but turning string is 
   definitely _cheap_.
  Something like this should do:
```Python
    def execute_view_tuple_value_dict(self, cube_name: str, view_name: str, private: bool = False, 
                                      top: int = None, skip: int = None, skip_zeros: bool = True, 
                                      skip_consolidated_cells: bool = False, skip_rule_derived_cells: bool = False, 
                                      value_separator: str = "|", **kwargs) -> dict:
        """ Optimized for performance. Get a Dict(tuple, value) from an existing Cube View
        Context dimensions are omitted in the resulting Dataframe !
        Cells with Zero/null are omitted by default, but still configurable!

        :param cube_name: String, name of the cube
        :param view_name: String, name of the view
        :param private: True (private) or False (public)
        :param top: Int, number of cells to return (counting from top)
        :param skip: Int, number of cells to skip (counting from top)
        :param skip_zeros: skip zeros in cellset (irrespective of zero suppression in MDX / view)
        :param skip_consolidated_cells: skip consolidated cells in cellset
        :param skip_rule_derived_cells: skip rule derived cells in cellset
        :param value_separator: value separator the tuple
        :return: Dict  {'2020|Jan|Sales': 2000, '2020|Feb|Sales': 3000}
        """
        lines = self.execute_view_csv(cube_name=cube_name, view_name=view_name, private=private, **kwargs)
        return {value_separator.join(entries.split(value_separator)[:-1]): entries.split(value_separator)[-1]
                for entries in lines.split("\r\n")[1:]}
```
2. How does `execute_view_tuple_value_dict` sound 😄 ?

MariusWirtz:
@rkvinoth 
thanks for the performance testing! I agree that 1 or 2 sec is not worth the extra code. 

Regarding the naming, how about `execute_view_elements_value_dict`. To me, tuple kinda sounds as if the key would be of the actual type `tuple`.

I think on the type hint you could be more specific:  `-> Dict[str, Union[str, float]]:`

rkvinoth:
> Regarding the naming, how about `execute_view_elements_value_dict`. To me, tuple kinda sounds as if the key would be of the actual type `tuple`.

I guess we have no choice but to go with `execute_view_elements_value_dict`.

TM1py/Services/CellService.py
```diff
@@ -546,6 +546,27 @@ def execute_view_csv(self, cube_name: str, view_name: str, private: bool = False
...
+                                        value_separator: str = "|", **kwargs) -> Dict[str, Union[str, float]]:
+        """ Optimized for performance. Get Dict from MDX Query.
+
+        :param mdx: Valid MDX Query
+        :param top: Int, number of cells to return (counting from top)
+        :param skip: Int, number of cells to skip (counting from top)
+        :param skip_zeros: skip zeros in cellset (irrespective of zero suppression in MDX / view)
+        :param skip_consolidated_cells: skip consolidated cells in cellset
+        :param skip_rule_derived_cells: skip rule derived cells in cellset
+        :param value_separator: value separator the tuple
```
MariusWirtz, line 560: perhaps it makes sense to call it `element_seperator` here

MariusWirtz:
> > Regarding the naming, how about `execute_view_elements_value_dict`. To me, tuple kinda sounds as if the key would be of the actual type `tuple`.
> 
> I guess we have no choice but to go with `execute_view_elements_value_dict`.

Thanks for the update. Looking good! 

We can still refactor when we come up with a better name.

TM1py/Services/CellService.py
rkvinoth: done.

MariusWirtz:
@rkvinoth 
I just thought about this function again and it made me wonder why we didn't use the `CaseAndSpaceInsensitiveDict` instead of the classic dict.
Can you think of a reason why not to use it? Does the `CaseAndSpaceInsensitiveDict` class lack features that the normal dict has? 
I suppose it would be more intuitive (and useful) for a TM1er if the dictionary is not case sensitive on element names.

rkvinoth:
Hi @MariusWirtz 

TBH I blindly made that as a `Dict`. Yes, we should convert them into `CaseAndSpaceInsensitiveDict`.
I will make the changes and push in a PR.

PR #300, opened 2020-07-28
Add optional `all_leaves_as_default` argument

### LINKED ISSUE #255: Modify clear_with_mdx or add second function
The clear_with_mdx function is pretty helpful since there is no proper way of ZeroOut via Rest. 
But you have to add all dimensions for a proper zero out. Even if for most of the dimensions all n elements (tm1subsetall...tm1filterbylevel) are used. Could it be possible to add a function where you only define the mdx for the relevant dimensions. For all other dimensions a TM1FilterByLevel(Tm1SubsetAll(.. is added to the ROWS-section

MariusWirtz:
Add optional `all_leaves_as_default` argument to `clear_with_mdx` function.
If the argument is passed as True, TM1py will accept incomplete MDX queries and complement it with the missing dimensions.
For the missing dimensions all leaves (`TM1FILTERBYLEVEL({TM1SUBSETALL([Product])}, 0)`) will be added to ROWS or COLUMNS.

This approach is fragile to a certain extend. The `not_referenced_dimensions`
are determined by comparing the titles of the dimension composition in the cellset and the WHERE selection in the MDX.

In order to deal with the fragility of the approach, the function is somewhat fussy about the MDX it accepts. The function will abort if
- MDX does not contain selections on ROWS and COLUMNS
- MDX does not contain either `ON 0` or `ON COLUMNS` (casing is ignored)
- MDX does not contain either `ON 1` or `ON ROWS` (casing is ignored)

Fixes #255

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -18,7 +18,7 @@ from TM1py.Services.RestService import RestService
 from TM1py.Services.ViewService import ViewService
 from TM1py.Utils import Utils, CaseAndSpaceInsensitiveSet, format_url
 from TM1py.Utils.Utils import build_pandas_dataframe_from_cellset, dimension_name_from_element_unique_name, \
-    CaseAndSpaceInsensitiveTuplesDict, abbreviate_mdx, build_csv_from_cellset_dict
+    CaseAndSpaceInsensitiveTuplesDict, abbreviate_mdx, build_csv_from_cellset_dict, get_dimensions_from_where_clause
 
 
 def tidy_cellset(func):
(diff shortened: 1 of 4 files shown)


rkvinoth:
Hey Marius,

What do think think about this approach? We can avoid the `cellset` parsing.
But, I'm not sure if it covers all the cases. 

```python

tm1 = TM1Service(address='', port=5000, ssl=True, user='', password='', namespace='', gateway='abc')
cube = 'Sales'

mdx = f"""
        SELECT NON EMPTY {{[Scenario].[Actuals]}} ON ROWS,
        NON EMPTY {{[Year].[2020]}} ON COLUMNS
        FROM [{cube}]
        """

mdx = mdx.upper()

if not ("ON 0" in mdx or "ON COLUMNS" in mdx):
    raise ValueError(f"MDX: '{mdx}' must contain 'ON 0' or 'ON COLUMNS'")

if not ("ON 1" in mdx or "ON ROWS" in mdx):
    raise ValueError(f"MDX: '{mdx}' must contain 'ON 1' or 'ON ROWS'")

if not ("FROM" in mdx):
    raise ValueError(f"MDX: '{mdx}' must contain 'FROM'")

import re
dimensions = tm1.cubes.get(cube).dimensions
mdx_without_cubename = re.sub(f'(.*)FROM\s+\[?{cube.upper()}\]?(.*)', r'\1' + r'\2', mdx)
missing_dimensions = [dimension for dimension in dimensions if dimension.upper() not in mdx_without_cubename]
additional_selections = []

for dimension in missing_dimensions:
    additional_selections.append(f"{{TM1FILTERBYLEVEL({{TM1SUBSETALL([{dimension}].[{dimension}])}},0)}}")
additional_mdx = "*".join(additional_selections)

for keyword in "ON 0", "ON 1", "ON ROWS", "ON COLUMNS":
    if keyword in mdx:
        mdx = mdx.replace(keyword, "*" + additional_mdx + " " + keyword)
        break
print(mdx)
```

rkvinoth:
Ahhh I just found one case:

1. What if one dimension's name is same as another dimension's element's name or hierarchy's name.

Any way ignore my suggestion.

MariusWirtz:
@rkvinoth 
I would like to avoid parsing the MDX. There are too many edge cases that you can hardly detect from parsing the MDX.
- Filtering one set (e.g. [Product] dimension) over a tuple that references other dimension names (e.g. [Time], [Version], etc.)
- The GENERATE function can be used to produce a set of dimension B from a base set of dimension A

Are you confident we can cover cases like these?

With the original approach, we limit the parsing to the WHERE condition. There we can confidently apply string parsing.
The trade-off is obviously performance, as we need to execute the MDX before we can call the clear. 

Any other opinions on that one?

rkvinoth:
> Are you confident we can cover cases like these?

Not at all 😄 ...


> Any other opinions on that one?

What if we pass element wise `MDX` and build the complete `MDX` from that? Something like this:

```python

tm1 = TM1Service(address='', port=5000, ssl=True, user='', password='', namespace='', gateway='abc')
cube = 'Sales'
element1 = '[Period].[01]'
element2 = '[Year].[2000]'
tm1.cells.clear_with_mdx(cube=cube, element1=element1, element2=element2)
```

MariusWirtz:
> What if we pass element wise `MDX` and build the complete `MDX` from that? Something like this:
> 
> ```python
> from TM1py.Services import TM1Service
> 
> tm1 = TM1Service(address='', port=5000, ssl=True, user='', password='', namespace='', gateway='abc')
> cube = 'Sales'
> element1 = '[Period].[01]'
> element2 = '[Year].[2000]'
> tm1.cells.clear_with_mdx(cube=cube, element1=element1, element2=element2)
> ```

I like this approach!
Perhaps we could go even one step further and allow not just element names, but keyword pairs of dimensions and set-expressions
like this kinda:


``` python
tm1.cells.clear(cube="Sales", product="[product].[ABC]", time="{[Time].[2020].Children}", version="Tm1FilterbyLevel({Tm1SubsetAll[Version]},0)"
```
If one pair is missing we just use all leaves when we build the MDX.

rkvinoth:
> ```python
> tm1.cells.clear(cube="Sales", product="[product].[ABC]", time="{[Time].[2020].Children}", version="Tm1FilterbyLevel({Tm1SubsetAll[Version]},0)"
> ```

lovely, very much readable! Docstring has to be updated accordingly

MariusWirtz:
OK. I will implement as a seperate function and remove the `complement_mdx_with_all_leaf_selections` function and the new `all_leaves_as_default` argument.

TM1py/Services/CellService.py
```diff
@@ -160,55 +163,53 @@ def clear_spread(
...
-            mdx = self.complement_mdx_with_all_leaf_selections(mdx)
+        return self.clear_with_mdx(cube=cube, mdx=mdx_builder.to_mdx(), **kwargs)
 
-        view_name = "".join(['}TM1py', str(uuid.uuid4())])
-        code = "".join([
-            f"ViewCreateByMdx('{cube}','{view_name}','{mdx}',1);",
-            f"ViewZeroOut('{cube}','{view_name}');"])
-        process = Process(name="")
-        process.prolog_procedure = code
+    def clear_with_mdx(self, cube: str, mdx: str, **kwargs):
```
rkvinoth, line 196: Could  you add some `docstring` here?

MariusWirtz:
@rkvinoth thanks for the review and the exchange of ideas

Issue #301, opened 2020-08-02
Tests need re-factoring


scrambldchannel:
**Describe the bug**
The test suite could do with a bit of a tidy up. Currently, a lot of tests error, fail, warn or are skipped. Some objects are also left on the server which makes repeat running will throw more exceptions.  

**To Reproduce**
Running the test suite gives a number of errors, fails, warnings and skipped tests. Eg using pytest as a runner, do ```python -m pytest  Tests/*.py```

**Expected behavior**
Ideally, the test suite should run without any errors and clean up all the objects it creates. Perhaps we could add a bit more documentation on running the tests too. 

**Version**

TM1py master branch
TM1 Server Version: 11.3

**Additional context**

I discussed this briefly and made a start months ago but had to drop it and the codebase has moved on quite a bit anyway. I'm happy to pick this up (feel free to assign me @MariusWirtz-cubewise ).

I'm by no means a python testing expert but think there a few things we can try to improve:

* Fixing fails, errors etc - I realise many of these are down to my server version but would be good to skip them or expect the exception.
* Cleanly manage all the objects created on the test server using fixtures as much as possible
* Align the structure and conventions used in each different test file
* Adding a little entry in the README about testing or even a separate tutorial

Thinking even further ahead:

* Introduce an option to run a suite of mock tests to enable at least some testing when working offline

I'm going to try to work on small PRs rather than one monolith, all help gratefully received.

Note, I ran coverage and even skipping a few problematic tests it gave a score of 90%, which is a pretty good foundation I'd say.

Happy to receive any thoughts and ideas!

MariusWirtz:
Thanks for bringing this topic up and volunteering to address it. Very much appreciated.
I agree with the 4 points you brought up.

For the server version, if something in TM1py doesn't work on a lower TM1 version we should add the new `require` decorator to the function.
Ideally, we also adjust the tests for those functions to not fail if version requirements are not met.

For me, all tests run successfully with version 11.7 with a few exceptions:
- in 11.7 and lower `test_create_chore_with_dst` fails due to an issue in TM1
- If the log filtering doesn't work it may be due to old/corrupt TM1 server log files in the log directory.
- I have seen occasional fails around chore time updates and process executions times (both are unproblematic TM1 issues IMO)


One thing I wanted to change in the test cases for a while is this : 
- map test modules 1 to 1 to the python classes. E.g. I think we should have one test class for the `SubsetService` class and another one for the `Subset` class.
[...]

PR #308, opened 2020-08-07
WIP - First look at Testing/annotations.py

scrambldchannel:
As part of https://github.com/cubewise-code/tm1py/issues/301, I've had a look at the tests for the annotations service as a starting point because it was a relatively simple class. It's still a bit rough but I thought I'd put it out there before getting too carried away. 

What I tried to do:

* Manage the fixtures a bit more cleanly and tried to ensure everything is cleaned up after each run
* Mapped the tests to the individual methods in the AnnotationsService (ie 5 tests instead of 3)
* Tested that each test could be run independently 
* Checked coverage
* Added some simple docstrings to explain what each bit is trying to do

A couple of things to note:

* The fixture creation and naming conventions could still be improved and possibly moved into helper functions shared across all test classes
* I've used more than one assert in each test but tried to relate them all to the method call being tested. We can split these into separate tests if you prefer
* Coverage shows that the ```move``` method of the Annotation object never gets called. I'm guessing this doesn't get used that often anyway. We could write a separate set of tests for all the objects but I guess that's not a huge priority...

Let me know what you think :)

### DIFF
--- a/Tests/Annotation.py
+++ b/Tests/Annotation.py
@@ -13,6 +13,9 @@ class TestAnnotationMethods(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
+        """
+        Establishes a connection to TM1 and creates a dimensions and a cube to use across all tests
+        """
 
         # Connection to TM1
         config = configparser.ConfigParser()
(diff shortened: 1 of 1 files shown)


MariusWirtz:
The tests are looking much better now. Thank you!

I agree that we could outsource the fixtures into a module that is shared through all tests.
We could probably get rid of a lot of duplicated code this way.

It's fine to do one test per tm1py method call IMO.

Agree. The `move` method is not the highest priority.

MariusWirtz:
Can I merge even though the MR has the WIP prefix? 🤔
The code is looking all good to me.

scrambldchannel:
Sure, happy to merge! I will try to start applying the same approach to the rest of the tests.

Cheers
Alex

PR #311, opened 2020-08-08
Skipped tests requiring pandas

scrambldchannel:
I really like the move to making pandas optional and wanted to make sure the tests could still run too when pandas isn't installed.

This PR uses the import logic from Utils/Utils.py and adds a decorator to skip the relevant tests. Tests can now be run without pandas installed without throwing an error and the tests are skipped with a message.

### DIFF
--- a/Tests/Cell.py
+++ b/Tests/Cell.py
@@ -14,7 +14,7 @@ from TM1py.Utils import Utils, element_names_from_element_unique_names
 
 try:
     import pandas as pd
-    
+
     _has_pandas = True
 except ImportError:
     _has_pandas = False
@@ -935,7 +935,7 @@ class TestDataMethods(unittest.TestCase):
         values = [float(value) for _, value in values.items()]
         self.assertEqual(self.total_value, sum(values))
 
-    @unittest.skip
+    @skip_if_no_pandas
     def test_execute_mdx_dataframe(self):
         mdx = MdxBuilder.from_cube(CUBE_NAME) \
             .rows_non_empty() \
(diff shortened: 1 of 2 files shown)


Tests/Cell.py
```diff
@@ -12,11 +12,29 @@
...
 from TM1py.Utils import Utils, element_names_from_element_unique_names
 
+try:
+    import pandas as pd
+
+    _has_pandas = True
+except ImportError:
+    _has_pandas = False
+
+def skipIfNoPandas(func):
```
MariusWirtz, line 22: Good idea to use a decorator here.
Please change the name to `skip_if_no_pandas` for the sake of consistency.

Tests/Cell.py
```diff
@@ -917,6 +935,7 @@ def test_execute_mdx_elements_value_dict(self):
         values = [float(value) for _, value in values.items()]
         self.assertEqual(self.total_value, sum(values))
 
+    @unittest.skip
```
MariusWirtz, line 938: I think we need the `skipIfNoPandas` decorator here instead.

Tests/Cell.py
```diff
@@ -1466,6 +1488,7 @@ def test_execute_view_dataframe(self):
         values = df[["Value"]].values
         self.assertEqual(self.total_value, sum(values))
 
+    @unittest.skip
```
MariusWirtz, line 1491: here as well

scrambldchannel:
Thanks for the feedback. I've added a couple more commits to the PR to address these points :)

Defining the decorator like that in each file is pretty ugly, I'll try to work out a cleaner way of doing it. I think a similar decorator to skip tests where the version isn't supported would be a good idea too. 

Cheers!

PR #312, opened 2020-08-08
Support IntegratedSecurityMode2 and 3

MariusWirtz:
New `integrated_login` argument to `TM1Service` / `RestService` allows login with IntegratedSecurityMode3:

``` python
with TM1Service(address="localhost", port=12354, ssl=True, integrated_login=True) as tm1:
    print(tm1.server.get_product_version())
```

### DIFF
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -125,6 +125,7 @@ class RestService:
         :param async_requests_mode: changes internal REST execution mode to avoid 60s timeout on IBM cloud
         :param connection_pool_size - In a multithreaded environment, you should set this value to a
         higher number, such as the number of threads
+        :param integrated_login: True for IntegratedSecurityMode3
         """
         self._ssl = self.translate_to_boolean(kwargs['ssl'])
         self._address = kwargs.get('address', None)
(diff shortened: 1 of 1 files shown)


scrumthing:
Hey @MariusWirtz , I tried the new integrated login method but was unsuccesful. Got the following error message. Do I need to do some further configuration?

```
TM1py.Exceptions.Exceptions.TM1pyRestException: Text: Error: Forbidden
Server is configured to only accept Kerberos package but the client is using NTLM when attempting to authenticate.
 Status Code: 403 Reason: Forbidden Headers: {'Content-Type': 'text/plain', 'Content-Length': '131', 'Connection': 'keep-alive', 'Content-Encoding': 'gzip'}
```

MariusWirtz:
Hi @scrumthing,

I found a recent IBM TM1 tech note about this problem
https://www.ibm.com/support/pages/node/6219292

Please let us know if their suggestions help to solve the problem.

MariusWirtz:
@scrumthing 
Please let us know once you solved the problem

scrumthing:
Will do. Current idea: Maybe it is connected to the default ssl certificate provided with the server. Waiting for a proper one. TM1 Web is not working properly with Kerberos as well.

scrumthing:
I only got it working by switching back to NTLM. The whole Kerberos thing is really annyoing because there is not much documentation to work on and IBM expects a trust for delegation to all services on the machine and that is not happening in our setup. So at the moment I am not sure in which direction I should go.

PR #313, opened 2020-08-08
Move skip_if_no_pandas to utils

scrambldchannel:
Actually, this might be a more elegant way of handling the pandas checks. I moved the decorator to the same place as ```require_pandas``` and aligned the code. I also removed the globals set in the test files as they aren't necessary.

I think this might be a way to approach a similar decorator for skipping tests were the required version is too low. 

part of https://github.com/cubewise-code/tm1py/issues/301

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -31,6 +31,7 @@ def require(version):
 
 
 def require_pandas(func):
+
     @functools.wraps(func)
     def wrapper(self, *args, **kwargs):
         try:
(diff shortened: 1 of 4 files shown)


MariusWirtz:
I agree it makes sense to outsource that function to somewhere else. Yes, we can take the same approach for skipping tests if the version is too low.

I am not sure if the TM1py Utils module is the ideal place though. Perhaps it makes more sense if we create a Utils module inside the Tests folder. Then we can keep a clean cut between the tests (which doesn't get distributed through pip) and the rest of TM1py. What do you think?

scrambldchannel:
Yeah, I see your point. I will look into a clean way of doing it in the Test folder itself. 

Cheers
Alex

PR #314, opened 2020-08-08
Add decorator to skip tests where version insufficient

scrambldchannel:
I've added another decorator that skips test where the version isn't high enough and applied it to some tests.


MariusWirtz:
Nice one! Thanks

Issue #309, opened 2020-08-07
Minor - Adjust regex in Objects/Process.py to avoid deprecation


scrambldchannel:
**Describe the bug**

```
tm1py/TM1py/Objects/Process.py:24: DeprecationWarning: Flags not at the start of the expression '#\\*\\*\\*\\*Begin: Gene' (truncated)
```

Triggered by https://github.com/cubewise-code/tm1py/blob/master/TM1py/Objects/Process.py#L23

**To Reproduce**
```pytest Tests/Application.py```

**Expected behavior**

We should address the deprecation warning

**Version**

master branch

**Additional context**

Came up while looking at the tests for the application service. 

Googling ```Flags not at the start of the expression ``` suggests that it might be as simple as changing:

```python
pattern = r"#\*\*\*\*Begin: Generated Statements(?s)(.*)#\*\*\*\*End: Generated Statements\*\*\*\*" 
```

To:

```python
pattern = r"(?s)#\*\*\*\*Begin: Generated Statements(.*)#\*\*\*\*End: Generated Statements\*\*\*\*"
 ```
I can see that regex is trying match to the auto generated block but not sure what the ```(?s)``` is trying to do. Maybe someone with a better understanding of regular expressions can help?

**Comment by rkvinoth** [comment]
`(?s)` is the `DOTALL` flag in **REGEX**

It is useful when the pattern you are searching for may span across multiple lines.
Basically it can match all characters including **line breaks**.

**Comment by scrambldchannel** [comment]
Cool, thanks @rkvinoth. Can the ```(?s)``` be safely moved to beginning of the line then? If I've understood correctly, it shouldn't make a difference when trying to match the generated statement block. What do you think ?

**Comment by rkvinoth** [comment]
1. With the new REGEX we are saying that the below statement could span across multiple lines in the new REGEX.
```
#****Begin: Generated Statements***
#****End: Generated Statements****
```

2. With the old REGEX we were saying that there could be one or more **line breaks** after `#****Begin: Generated Statements***` and before `#****End: Generated Statements****`

Lets go with the new approach:
`pattern = r"(?s)#\*\*\*\*Begin: Generated Statements(.*)#\*\*\*\*End: Generated Statements\*\*\*\*"`

scrambldchannel:
Thanks @rkvinoth 

Note I just realised I didn't copy and paste the code properly, have adjusted my comment now. That is, the original statement has the ```?s``` but halfway through the string. 

[...]

PR #317, opened 2020-08-09
?s to start of str to remove deprecation warning

scrambldchannel:
As per the discussion in https://github.com/cubewise-code/tm1py/issues/309

I don't think this change breaks the functionality and suppresses a deprecation warning. 

What do you think? It's not urgent, on Python 3.8 it is still just throwing a warning.

Cheers
Alex


MariusWirtz:
Thanks! I double-checked the new regex and it works like a charm.

PR #318, opened 2020-08-09
Skip tests that crash the server on 11.3

scrambldchannel:
These tests cause the server being tested to crash (11.3). I think it's safe to mark them as requiring a greater version.

What do you think?

Cheers
Alex


MariusWirtz:
Absolutely agree

PR #321, opened 2020-08-10
get_elements_by_wildcard

rkvinoth:
```PYTHON

tm1 = TM1Service()
# Filter 'qtr' from all level elements
print(tm1.elements.get_elements_by_wildcard('Month', 'Month', 'qtr'))
```

**Output**: `['Qtr-01', 'Qtr-02', 'Qtr-01', 'Qtr-04']`

```PYTHON

tm1 = TM1Service()
# Filter 'ju' from level 0 elements
print(tm1.elements.get_elements_by_wildcard('Month', 'Month', 'ju', 0))
```

**Output**: `['June', 'July']`

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -129,8 +129,7 @@ class ElementService(ObjectService):
         mdx_elements = f"{{ Tm1FilterByLevel ( {{ Tm1SubsetAll ([{dimension_name}].[{hierarchy_name}]) }} , 0 ) }}"
         return self.get_element_identifiers(dimension_name, hierarchy_name, mdx_elements, **kwargs)
 
-    def get_elements_by_level(self, dimension_name: str, hierarchy_name: str, level: int,
-                              **kwargs) -> List[str]:
+    def get_elements_by_level(self, dimension_name: str, hierarchy_name: str, level: int, **kwargs) -> List[str]:
         """ Get all element names by level in a hierarchy
 
         :param dimension_name: Name of the dimension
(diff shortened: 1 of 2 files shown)


TM1py/Services/ElementService.py
```diff
@@ -142,7 +141,27 @@ def get_elements_by_level(self, dimension_name: str, hierarchy_name: str, level:
...
+                                 wildcard: str, level: int = None, **kwargs) -> List[str]:
+        """ Get all element names filtered by wildcard (CaseInsensitive) and level in a hierarchy
+
+        :param dimension_name: Name of the dimension
+        :param hierarchy_name: Name of the hierarchy
+        :param wildcard: wildcard to filter
+        :param level: Level to filter
+        :return: List of element names
+        """
+        filter_elements = format_url("contains(tolower(Name),tolower('{}'))", wildcard)
```
MariusWirtz, line 158: 2 things about this filter:

- we have to be careful, not to apply `format_url` on a string twice. I think if you are using it below, you can skip it here.
- If it's not too expensive, I think it would be nice to cater to space insensitivity as well: `tolower(replace(Name, ' ', ''))`

TM1py/Services/ElementService.py
```diff
@@ -142,7 +141,27 @@ def get_elements_by_level(self, dimension_name: str, hierarchy_name: str, level:
             "/api/v1/Dimensions('{}')/Hierarchies('{}')/Elements?$select=Name&$filter=Level eq {}",
             dimension_name,
             hierarchy_name,
-            level)
+            str(level))
+        response = self._rest.GET(url, **kwargs)
+        return [e["Name"] for e in response.json()['value']]
+
+    def get_elements_by_wildcard(self, dimension_name: str, hierarchy_name: str,
```
MariusWirtz, line 148: I would suggest calling this function `get_elements_filtered_by_wildcard`. Then it is "in line" with the  `get_elements_filtered_by_attribute` function in the same class.
What do you think?

rkvinoth: I actually wanted that name and then thought it wouldn't be "in line" with the `get_elements_by_level` method. I will update the name.

TM1py/Services/ElementService.py
```diff
@@ -142,7 +141,27 @@ def get_elements_by_level(self, dimension_name: str, hierarchy_name: str, level:
...
+        """ Get all element names filtered by wildcard (CaseInsensitive) and level in a hierarchy
+
+        :param dimension_name: Name of the dimension
+        :param hierarchy_name: Name of the hierarchy
+        :param wildcard: wildcard to filter
+        :param level: Level to filter
+        :return: List of element names
+        """
+        filter_elements = format_url("contains(tolower(Name),tolower('{}'))", wildcard)
+        if isinstance(level, int):
```
MariusWirtz, line 159: what if someones passes `level` as str or float? Shall we raise a `ValueError` indicating that we expect an integer?

TM1py/Services/ElementService.py
MariusWirtz: Well that's a good point. Then I guess both names are fine :)

TM1py/Services/ElementService.py
rkvinoth: Actually `int` or `str` or `float` should work.. I will update the condition to:
```python
if level is not None:
```

TM1py/Services/ElementService.py
rkvinoth: Performs _CaseAndSpaceInsensitive_ filter now.

rkvinoth:
Sorry about the **commit message**.. stupid Pycharm.

MariusWirtz:
Nice one. Thanks!

PR #323, opened 2020-08-10
Move tests from other and utils tests to separate files for each utils module

scrambldchannel:
Currently, the tests in Tests/Other.py and Tests/Utils.py test against a few different modules:

* Services/Security.py
* Utils/Utils.py
* Utils/MDXUtils.py
* Utils/TIObfuscator.py

I've tried to split them into three separate files for the three utils modules and put the security tests in the existing security test file. I think this makes it a bit clearer what is actually being tested. Ultimately you might want the test structure to mirror the library structure (eg tests for Objects, Servers and Utils in different directories) but I thought this was a useful start.

What do you think?

Cheers
Alex


MariusWirtz:
I agree the test structure should mirror the library structure.

Thanks for breaking it down. It makes sense.
Ultimately, I would like to get rid of the MDXUtils and manage the TIObfuscator as a separate project.

PR #324, opened 2020-08-10
Change pytest.raises to unittest's assertRaises

scrambldchannel:
I noticed a couple of the tests use pytest. I like pytest but since we're using unittest for 99% of the tests, I think it makes sense to replace these pytest methods with equivalent unittest logic.

What do you think?

Cheers
Alex

### DIFF
--- a/Tests/Cell.py
+++ b/Tests/Cell.py
@@ -3,14 +3,15 @@ import types
 import unittest
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


MariusWirtz:
Thank you. The test suite is not the prettiest part of TM1py. I really appreciate your contributions here!

I agree it makes sense to align them.

PR #325, opened 2020-08-10
return list in execute_*_values functions instead of generator

MariusWirtz:


rkvinoth:
@MariusWirtz  Any specific reason?

MariusWirtz:
Thanks for your comment.

I think using a generator here was a mistake, to begin with. 
It's counterintuitive since in the vast majority of functions we do not return generators but actual data structures. 

Also when you are actually trying to do something with the result you would always have to convert it into a list anyways.
Previously:
``` python
from TM1py import TM1Service

DIMENSION_NAME = "plan_department"

with TM1Service(address="", port=12354, user="admin", password="apple", ssl=True) as tm1_source:
    with TM1Service(address="", port=12297, user="admin", password="apple", ssl=True) as tm1_target:
        mdx = f"""
        SELECT 
        {{Tm1SubsetAll([{DIMENSION_NAME}])}} ON ROWS,
        {{Tm1SubsetAll([}}ElementAttributes_{DIMENSION_NAME}])}} ON COLUMNS
        FROM [}}ElementAttributes_{DIMENSION_NAME}]
        """

        cell_values = list(tm1_source.cells.execute_mdx_values(mdx))

        tm1_target.cells.write_values_through_cellset(mdx, cell_values)
```

Now:
```python
from TM1py import TM1Service

DIMENSION_NAME = "plan_department"

with TM1Service(address="", port=12354, user="admin", password="apple", ssl=True) as tm1_source:
    with TM1Service(address="", port=12297, user="admin", password="apple", ssl=True) as tm1_target:
        mdx = f"""
        SELECT 
        {{Tm1SubsetAll([{DIMENSION_NAME}])}} ON ROWS,
        {{Tm1SubsetAll([}}ElementAttributes_{DIMENSION_NAME}])}} ON COLUMNS
        FROM [}}ElementAttributes_{DIMENSION_NAME}]
        """

        cell_values = tm1_source.cells.execute_mdx_values(mdx)

        tm1_target.cells.write_values_through_cellset(mdx, cell_values)
```


I am not too concerned about backwards compatibility, as a list is still iterable. 
What do you think?

rkvinoth:
I thought that the reason we leveraged `generators` is to save memory. But if we are not concerned about the object size, then we should be good.

MariusWirtz:
Yeah. That's what generators are good for normally. 
Here is this case all we are saving is one iteration through the response dictionary.

Memory-wise I think we are doing better without the generator. Now by the end of the function, the python garbage collector should destroy the response object (which should be quite big compared to the list) and leave the new condensed list "alive".

I think it was a mistake to use a generator here. Even when the response contains 1M elements, python can loop through it in no time.

Issue #322, opened 2020-08-10
Enhancement Request - Function to unwind consolidation recursively


cubewise-gng:
Could I please request a function to unwind a dimension hierarchy consolidation specified, with the option to unwind recursively to its children.

**Version**
- TM1py 1.5.0

MariusWirtz:
solved with #326 

@cubewise-gng 
Unfortunately, the TM1 REST API does not provide the functionality to delete multiple edges from a hierarchy in one batch.  Therefore we must retrieve the full hierarchy to python, update it, and send it back. I imagine that this approach will be slower than calling the `}bedrock.hier.unwind` process in most cases.

PR #326, opened 2020-08-10
remove_edges_under_consolidation

rkvinoth:
Solution to https://github.com/cubewise-code/tm1py/issues/322#issue-675997568
Any ideas on refactoring?

### DIFF
--- a/TM1py/Services/HierarchyService.py
+++ b/TM1py/Services/HierarchyService.py
@@ -212,10 +212,10 @@ class HierarchyService(ObjectService):
         return self._rest.PATCH(url=url, data=json.dumps(body), **kwargs)
 
     def remove_edges_under_consolidation(self, dimension_name: str,  hierarchy_name: str,
-                                         consolidation_element: str, **kwargs) -> Response:
+                                         consolidation_element: str, **kwargs) -> List[Response]:
         """
         :param dimension_name: Name of the dimension
-        :param hierarchy_name: Nmae of the hierarchy
+        :param hierarchy_name: Name of the hierarchy
         :param consolidation_element: Name of the Consolidated element
         :return: response
         """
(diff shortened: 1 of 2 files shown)


TM1py/Services/HierarchyService.py
```diff
@@ -211,6 +211,30 @@ def remove_all_edges(self, dimension_name: str, hierarchy_name: str = None, **kw
...
+        elements_under_consolidations = element_service.get_members_under_consolidation(dimension_name, hierarchy_name,
+                                                                                        consolidation_element)
+        elements_under_consolidations.append(consolidation_element)
+        remove_edges = []
+        for element, _ in hierarchy.edges.items():
+            if element[0] in elements_under_consolidations and element[1] in elements_under_consolidations:
+                remove_edges.append(element)
+        hierarchy.remove_edges(remove_edges)
+
+        url = format_url("/api/v1/Dimensions('{}')/Hierarchies('{}')", dimension_name, hierarchy_name)
```
MariusWirtz, line 234: I think you can just use the `self.update` function here instead

TM1py/Services/HierarchyService.py
```diff
@@ -211,6 +211,30 @@ def remove_all_edges(self, dimension_name: str, hierarchy_name: str = None, **kw
...
+        :return: response
+        """
+        hierarchy = self.get(dimension_name, hierarchy_name)
+        from TM1py.Services import ElementService
+        element_service = ElementService(self._rest)
+        elements_under_consolidations = element_service.get_members_under_consolidation(dimension_name, hierarchy_name,
+                                                                                        consolidation_element)
+        elements_under_consolidations.append(consolidation_element)
+        remove_edges = []
+        for element, _ in hierarchy.edges.items():
```
MariusWirtz, line 229: I think it's better to write 
`for edge in hierarchy.edges:`
here. Or:
I think it's better to write 
`for (parent, component) in hierarchy.edges:`
here.

MariusWirtz:
Thanks for solving #322

I don't like that we have to retrieve the full hierarchy and send it back. It's going to be slow on big dimensions.
For the moment I see no better approach than the one you implemented though.

Once TM1 supports multi delete operations on collections, we could add a `delete_edges` function to this service and simplify (and speed up) this function.

TM1py/Services/HierarchyService.py
rkvinoth: nice..

TM1py/Services/HierarchyService.py
rkvinoth: agreed..

rkvinoth:
That's right.. even I was expecting a collection delete. 
Anyway let's use the available resources for now.

PR #329, opened 2020-08-10
skip tests in applications that fail on 11.3

scrambldchannel:
Another little change to skip a couple of further tests that fail on my setup.

Note I couldn't find anything in the release notes to pinpoint this so might just be a point release bug but the error I get is below for reference. It might just be down to the way all the TM1 artifacts are arranged for these tests (which I still want to clean up) but thought it makes sense to use the version skip for now, especially if it's not causing an issue for others.

Cheers
Alex

```
Tests/Application.py:218: in run_test_document_application
    self.tm1.applications.create(application=app, private=private)
TM1py/Services/ApplicationService.py:170: in create
    url = format_url(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

url = "/api/v1/Contents('Applications'){/Contents('TM1py_Tests_Applications_RootFolder')}/Contents('{name}.blob')/Document/Content", args = [], kwargs = {'name': 'TM1py_Tests_Applications_Document'}

    def format_url(url, *args: str, **kwargs: str) -> str:
        """ build url and escape single quotes in args and kwargs
    
        :param url: url with {} placeholders
        :param args: arguments to placeholders
        :return:
        """
        args = [build_url_friendly_object_name(arg) if isinstance(arg, str) else arg
                for arg
                in args]
    
        kwargs = {key: build_url_friendly_object_name(value) if isinstance(value, str) else value
                  for key, value
                  in kwargs.items()}
    
>       return url.format(*args, **kwargs)
E       KeyError: "/Contents('TM1py_Tests_Applications_RootFolder')"
```

### DIFF
--- a/Tests/Application.py
+++ b/Tests/Application.py
@@ -10,6 +10,8 @@ from TM1py import TM1Service, Element, ElementAttribute, Hierarchy, Dimension, C
 from TM1py.Objects.Application import CubeApplication, ApplicationTypes, ChoreApplication, DimensionApplication, \
     FolderApplication, LinkApplication, ProcessApplication, SubsetApplication, ViewApplication, DocumentApplication
 
+from .TestUtils import skip_if_insufficient_version
+
 # Hard coded stuff
 PREFIX = 'TM1py_Tests_Applications_'
 TM1PY_APP_FOLDER = PREFIX + "RootFolder"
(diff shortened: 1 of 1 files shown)


MariusWirtz:
I think you stumbled over this bug in the TM1 server:
https://community.ibm.com/community/user/businessanalytics/communities/community-home/digestviewer/viewthread?MessageKey=0dbb8f89-ad87-4f3c-9de5-14dc699d1f04&CommunityKey=8fde0600-e22b-4178-acf5-bf4eda43146b&tab=digestviewer#bm0dbb8f89-ad87-4f3c-9de5-14dc699d1f04

I just don't know in which version they addressed it (11.6 or 11.7 I guess). It only impacts private applications. Not sure if we should add the `require` decorator to the function itself.

MariusWirtz:
Either way, I will accept the merge. We do have to skip the tests in < 11.4.

scrambldchannel:
Thanks @MariusWirtz, I didn't see that. I guess I didn't search hard enough :)

PR #332, opened 2020-08-10
skip a test that fails <=11.7

scrambldchannel:
Another test to skip, as per comment here:

https://github.com/cubewise-code/tm1py/blob/e3e2c69a494d9cec5b2904c11ade3cd673ca9579/Tests/Chore.py#L127

It might just be the assert around the multiple/single commit mode so the test can possibly be split into two to test the rest of the functionality.

### DIFF
--- a/Tests/Chore.py
+++ b/Tests/Chore.py
@@ -9,6 +9,8 @@ from pathlib import Path
 from TM1py.Objects import Chore, ChoreStartTime, ChoreFrequency, ChoreTask, Process
 from TM1py.Services import TM1Service
 
+from .TestUtils import skip_if_insufficient_version
+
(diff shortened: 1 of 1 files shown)


MariusWirtz:
I agree.
Ideally we create two tests: `test_create_chore_with_dst_single_commit` and `test_create_chore_with_dst_multi_commit` and only one should fail in <= 11.7.

scrambldchannel:
I've added a separate test now using single commit and it passes for me. Duplicating the code like I've done maybe isn't the most elegant but at least it's now getting tested. What do you think?

MariusWirtz:
Thanks. The approach makes sense I think.

PR #333, opened 2020-08-11
Testing/class renaming and removing imports from __init__.py

scrambldchannel:
Hi

A little cleanup to give test classes unique names.

Also included is removing imports from Tests/__init__.py. Do we need these there? Removing them doesn't seem to impact my testing workflow (pytest from command line and within vscode). Removing them avoids processing the files and allows you to run the tests in eg Cells.py even if eg Annotations.py has errors. 

Cheers
Alex

### DIFF
--- a/TM1py/Services/ChoreService.py
+++ b/TM1py/Services/ChoreService.py
@@ -21,21 +21,41 @@ def deactivate_activate(func):
     """
 
     @functools.wraps(func)
-    def wrapper(self, chore: Chore):
-        # Get Chore
-        chore_old = self.get(chore.name)
+    def wrapper(self, *args, **kwargs):
+        # chore (type Chore) or chore_name (type str) is passed as first arg or kwargs
+        if 'chore' in kwargs:
+            chore = kwargs.get('chore').name
+        elif 'chore_name' in kwargs:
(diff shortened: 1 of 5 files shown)


MariusWirtz:
Thanks for addressing the naming issues. 

That's a fair point.
But if we remove those import, I can't right-click the folder in PyCharm and click `Run TM1py Tests` anymore 😄 
How do you execute it from command without the imports in the init file?

scrambldchannel:
Ah, OK, I thought it might be something to do with pycharm but wasn't sure. I tend to just run them from the command line, ie:

```sh
$ pytest -x Tests/*.py
```

I use vscode and the tests don't show up at all by default, it will only show tests with names like ```test_annotations``` so don't tend to run them that way. I'm not sure how pycharm does it. 

It's not a huge deal, I can redo the PR to leave __init__.py intact if you like?

MariusWirtz:
Thanks. 

I think it's best to leave the imports as is. 
Then everyone who uses PyCharm has an easy way to execute them and a visual interface.

scrambldchannel:
Sure, makes sense. Let me close this one and raise a new PR, there's a lot of noise after I tried to merge other changes

PR #339, opened 2020-08-18
Avoid execute_dataframe method to fail over 0 cell

MariusWirtz:
0 cells in REST API are returned as None.
We must not write None to CSV string, but "" in order to produce NaN in pandas dataframe


rkvinoth:
Nice! I wanted to do something about this as well.

PR #350, opened 2020-08-27
wip: Add manage_transaction_log decorator

### LINKED ISSUE #343: Deactive transaction log property for write_values method
After a brief discussion with @MariusWirtz on the write_values method we came up with the question if said method should be expanded with a property to deactivate the transaction log. It is common practice to deactivate the transaction log for bulk loads via TI. It could be done in tm1py as well by changing the value in the cube properties cube but a more convenient way would be a property attached to the write_value method which if true deactivates the transaction log and reactivates it again after the write operation is finished.

MariusWirtz:
Add transaction_log_is_active function

Fixes #343

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -744,11 +744,26 @@ def get_dimensions_from_where_clause(mdx: str) -> List[str]:
 
 
 def get_cube(mdx: str) -> str:
-    _, sub_mdx = mdx.upper().replace(" ", "").split("FROM")
-    if "WHERE" in sub_mdx:
-        sub_mdx, _ = sub_mdx.split("WHERE")
-    return sub_mdx.strip().strip("[").strip("]")
 
+    # replace tabs, line breaks, spaces
+    mdx = re.sub(r'\s+', '', mdx)
+
+    # happy case: cube name in square brackets
+    pattern = r"(?s)(?i)FROM\[(.*?)\]"
+    search_result = re.search(pattern, mdx)
+    if search_result:
+        return search_result.group(1)
+
+    # cut off where
+    pattern = r"(?s)(?i).*SELECT.*ON.*FROM.*WHERE\(.*"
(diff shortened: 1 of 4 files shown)


MariusWirtz:
@scrumthing
What do you think?

scrumthing:
I think that is the best way to do it. It returns _True_ or _False_ and with that the developer can work. :-)

MariusWirtz:
> I think that is the best way to do it. It returns _True_ or _False_ and with that the developer can work. :-)

Thanks @scrumthing 
actually I would like this decorator to work for the more efficient `write_values_through_cellset` function as well. Wouldn't that be nice? 
I Will leave this MR in wip state for the moment

TM1py/Utils/Utils.py
```diff
@@ -742,6 +743,18 @@ def get_dimensions_from_where_clause(mdx: str) -> List[str]:
...
+def get_cube(mdx: str) -> str:
+    _, sub_mdx = mdx.upper().replace(" ", "").split("FROM[")
+    return sub_mdx[: sub_mdx.find("]")]
+
+
+def resembles_mdx(mdx: str) -> bool:
+    pattern = r"(?s)(?i).*SELECT.*ON.*FROM.*\[.*].*"
+    if re.search(pattern=pattern, string=mdx):
+        return True
+    return False
```
MariusWirtz, line 755: @rkvinoth, @scrambldchannel 
Any thoughts or suggestions on this regex? Does it look valid? 
I am using it elsewhere to differentiate between a cube name and an MDX query.

rkvinoth: @MariusWirtz  I think `r"(?s)(?i).*SELECT.*ON.*FROM"` should be enough. Because we can pass the cube name without `[]` (you know the scenario).

MariusWirtz: @rkvinoth thanks for the hint! I will fix it.

Issue #363, opened 2020-09-19
Filter Message Log on Server Side


scrumthing:
I am thinking about the message log and how you can extract useful information while at the same time reducing the amount of data retrieved. We can filter the message log on the server side by time. But we don't actually know for example the timestamp from the last time the server was restarted. I can easily retrieve that information from the message log if I download the whole message log. But wouldn't it be easier to have a filter argument (or/and a specific function) to retrieve just the message log entries when the server was started and finished with the start and last shutdown? 
The rest api should allow to filter the message log server side for those specific log entries. Am I right?
With the result you had an overview of your startup times at hand and could use that information to a) filter only the server startup log entries or b) monitor everything that happens in between a finished start and a shutdown.

I know there are other ways to achieve it but I was thinking about an easy way within tm1py to get that information. What do I you think?

Issue #342, opened 2020-08-21
Improve Messagelog method


scrumthing:
When I prepared my presentation for tm1 horizon next week I realized that the current response for the message log method is not directly usable it being more or less a list of lines. For further querying it you need to convert it to a different object (in my case a pandas dataframe) and translate the timestamp properly and get more options for filtering it. 
As I am not an expert for the message log I want to use the issue to start a discussion on that topic. 
- What are your thoughts in that?
- What kind of useful information can you get from there? As I mentioned [here ](https://www.tm1forum.com/viewtopic.php?f=3&t=15467)I was thinking about feeder load time for example. 
- How should the customization of the log (via logproperties file) be handled?

``` Python
    def get_message_log_entries(self, reverse: bool = True, top: int = None, **kwargs) -> Dict:
        reverse = 'true' if reverse else 'false'
        url = '/api/v1/MessageLog(Reverse={})'.format(reverse)
        if top:
            url += '?$top={}'.format(top)
        response = self._rest.GET(url, **kwargs)
        return response.json()['value']```

PR #365, opened 2020-09-21
new filters for get_mesage_log_entries (updated)

adscheevel:
Added parameters to get_message_log_entries function to filter by logger, level, and substring in message. The msg_contains parameter can be a list or simple string. Suggesting for enhancement taken from issues #363 and #342 

EX: find all feeder processing statements for cube called "Income_Statement"
```python
with TM1Service() as tm1:
	logs = get_message_log_entries(msg_contains=['ProcessFeeders', "'Income_Statement'"])
```

EX: find all chore finishing times
```python
with TM1Service() as tm1:
	logs = get_message_log_entries(logger='TM1.Chore', msg_contains='finished executing')
```

EX: find all feeder finished processing statements:
```python
with TM1Service() as tm1:
	logs = get_message_log_entries(msg_contains='Done computing feeders')
```

EX: find 10 most recent TI process errors
```python
with TM1Service() as tm1:
	logs = get_message_log_entries(reverse=True, top=10, level='Error', logger='TM1.Process')
```


TM1py/Services/ServerService.py
```diff
@@ -74,36 +74,58 @@ def execute_message_log_delta_request(self, **kwargs) -> Dict:
...
                     until = self.utc_localize_time(until)
-                time_filters.append(format_url("TimeStamp le {}", until.strftime("%Y-%m-%dT%H:%M:%SZ")))
+                log_filters.append(format_url("TimeStamp le {}", until.strftime("%Y-%m-%dT%H:%M:%SZ")))
+
+            if logger:
+                log_filters.append(format_url("Logger eq '{}'", logger))
 
-            url += "&$filter={}".format(" and ".join(time_filters))
+            if level:
+                level_dict={'Error':1, 'Warning':2, 'Info':3, 'Debug':4, 'Unknown':5}
```
rkvinoth, line 111: Create a `CaseAndSpaceInsensitiveDict` here instead of normal `dict`. Add `whitespace` after `:` to follow **PEP 8**

TM1py/Services/ServerService.py
```diff
@@ -74,36 +74,58 @@ def execute_message_log_delta_request(self, **kwargs) -> Dict:
         return response.json()['value']
 
     def get_message_log_entries(self, reverse: bool = True, since: datetime = None,
-                                until: datetime = None, top: int = None, **kwargs) -> Dict:
+                                until: datetime = None, top: int = None, logger: str = None, level: str = None, msg_contains: str = None, **kwargs) -> Dict:
```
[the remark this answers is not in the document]
rkvinoth: Also break the line to follow **PEP 8**

TM1py/Services/ServerService.py
```diff
@@ -74,36 +74,58 @@ def execute_message_log_delta_request(self, **kwargs) -> Dict:
...
-                time_filters.append(format_url("TimeStamp le {}", until.strftime("%Y-%m-%dT%H:%M:%SZ")))
+                log_filters.append(format_url("TimeStamp le {}", until.strftime("%Y-%m-%dT%H:%M:%SZ")))
+
+            if logger:
+                log_filters.append(format_url("Logger eq '{}'", logger))
 
-            url += "&$filter={}".format(" and ".join(time_filters))
+            if level:
+                level_dict={'Error':1, 'Warning':2, 'Info':3, 'Debug':4, 'Unknown':5}
+                nLvl = level_dict[level]
```
rkvinoth, line 112: Instead of querying `level_dict[level]` use the `get` method to avoid `KeyError`... `level_dict.get(level)`.

Also, could you use a better variable name like `level_index` instead of `nLvl`?

on the whole, it would look something like this:

```PYTHON
level_index = level_dict.get(level)

if level_index :
    log_filters.append("Level eq {}".format(level_index))
```

TM1py/Services/ServerService.py
```diff
@@ -74,36 +74,58 @@ def execute_message_log_delta_request(self, **kwargs) -> Dict:
...
+                log_filters.append("Level eq {}".format(nLvl))
+
+            if msg_contains:
+                if isinstance(msg_contains, list):
+                    msg_filters = []
```
rkvinoth, line 117: Why not have a `list comprehension` here?

`msg_filters = [format_url("contains(Message,'{}')", wildcard) for wildcard in msg_contains]`

rkvinoth (changes requested):

MariusWirtz:
@adscheevel,
thank you for this MR. Very nice work and a useful feature!

Along with the changes @rkvinoth already requested, I have one minor suggestion w.r.t. the `message_contains` argument.

I think it would be nice if the expected argument would be of the type `Iterable` _(from collections.abc import Iterable)_.
This way we would expect `str`, `list`, `tuple`, or anything else that's iterable.

So instead of checking explicitly of the type `list`, you can check for the type `str`. If it is not of type `str`, you can assume that an iterable other than str is passed and you can consume it in the list comprehension as @rkvinoth suggested

TM1py/Services/ServerService.py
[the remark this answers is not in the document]
adscheevel: updated commit has line break

For the type hint, i set msg_contains: Iterable = None based on suggestion from @MariusWirtz

TM1py/Services/ServerService.py
adscheevel: Thanks for suggestion on CaseAndSpaceInsensitiveDict and I've added proper whitespace

TM1py/Services/ServerService.py
adscheevel: changed variable to level_index and now use level_dict.get(level)

TM1py/Services/ServerService.py
adscheevel: Ignorance on my part, appreciate the suggestion and have incorporated.

MariusWirtz:
All looking good from my side! Thanks @adscheevel and @rkvinoth  for the contributions.

PR #367, opened 2020-09-23
Properly handle verify as str or bool

MariusWirtz:


MariusWirtz:
@rkvinoth
what do you think?

rkvinoth:
Looks like a painful check and I don't see a way to optimize it further...
Please go ahead and merge it.

Issue #216, opened 2020-03-31
Support Sandboxes


MariusWirtz:
TM1py should offer features to manage TM1 Sandboxes

Tasks:
- Implement a `SandboxService` that offers `create`, `read`, `update`, `delete` methods for sandboxes.
Reference: https://www.ibm.com/developerworks/community/forums/html/threadTopic?id=77d13251-5276-47e0-b65a-1f6389b0b36d

- Fix tests that currently don't work with `EnableSandboxDimension=T`

- Add optional `sandbox` argument to `write_value`, `write_values`, `write_values_through_cellset` method

- `execute_mdx_...` and `execute_view_...` methods need an optional argument `skip_sandbox`. Which controls if the first element will be excluded from the result set if it is `[Sandboxes].[Base]`

MariusWirtz:
Implemented

PR #371, opened 2020-09-27
Sandbox

andreyea:
New sandbox service to create, delete, publish, reset, and merge sandboxes.
New sandbox object.
Updated all cell service methods that request data from TM1 or write data into TM1 cubes. All these methods have new optional sandbox_name string parameter. Whenever it is used, the request URL will have an extra parameter - !sandbox='<sandbox_name>'.


TM1py/Objects/Sandbox.py
```diff
@@ -0,0 +1,62 @@
...
+from TM1py.Objects.TM1Object import TM1Object
+from TM1py.Utils import format_url
+
+
+class Sandbox(TM1Object):
+    """ Abstraction of a TM1 Sandbox
+        
+    """
+
+    def __init__(self, name: str, includeInSandboxDimension: bool = True):
```
MariusWirtz, line 15: Please use `lower_case_with_underscores` for the argument

tst.py
```diff
@@ -0,0 +1,43 @@
+from TM1py.Services import TM1Service
```
MariusWirtz, line 1: We will need to create a dedicated test module for Sandboxes in line with the test modules for chores, cubes, etc. Then we can remove this prototype file.

I suppose we can do the tests in a separate MR since this MR is already quite big.

TM1py/Services/SandboxService.py
```diff
@@ -0,0 +1,109 @@
...
+    
+    """
+
+    def __init__(self, rest: RestService):
+        super().__init__(rest)
+
+    def get(self, sandbox_name: str, **kwargs) -> Sandbox:
+        """ get a sandbox from TM1 Server
+
+        :param cube_name:
```
MariusWirtz, line 23: It should say `sandbox_name` here

TM1py/Services/CellService.py
```diff
@@ -31,7 +31,11 @@ def wrapper(self, cellset_id, *args, **kwargs):
             return func(self, cellset_id, *args, **kwargs)
         finally:
             if kwargs.get("delete_cellset", True):
-                self.delete_cellset(cellset_id=cellset_id)
+                sandbox_name=kwargs.get("sandbox_name", None)
+                if sandbox_name != None:
```
MariusWirtz, line 35: The more pythonic way would be 
``` python
if sandbox_name is not None:
``` 
https://stackoverflow.com/a/3965165/4046584

TM1py/Services/CellService.py
```diff
@@ -180,15 +189,24 @@ def clear_with_mdx(self, cube: str, mdx: str, **kwargs):
...
 
         :param cellset_id:
         :param payload:
+        :param sandbox_name: str
         :param kwargs:
         :return:
         """
-        url = format_url("/api/v1/Cellsets('{}')/tm1.Update", cellset_id)
+        url = ""
+        if sandbox_name == None:
```
MariusWirtz, line 202: more pythonic would be 

``` python
if sandbox_name is None:
```

TM1py/Services/CellService.py
```diff
@@ -204,18 +222,27 @@ def get_dimension_names_for_writing(self, cube_name: str, **kwargs) -> List[str]
...
         :param dimensions: optional. Dimension names in their natural order. Will speed up the execution!
+        :paramm sandbox_name: str
         :return: response
         """
         if not dimensions:
             dimensions = self.get_dimension_names_for_writing(cube_name=cube_name)
-        url = format_url("/api/v1/Cubes('{}')/tm1.Update", cube_name)
+        url = ""
+        if sandbox_name == None:
+            url = format_url("/api/v1/Cubes('{}')/tm1.Update", cube_name)
```
MariusWirtz, line 239: @andreyea,
perhaps it's easier if we add an optional `sandbox_name` argument to the `format_url` function.
Then we can outsource the check in the function. What do you think?

TM1py/Services/CellService.py
```diff
@@ -847,47 +931,88 @@ def extract_cellset_raw(
...
+                "Cells($select={cell_properties}{top_cells}{skip_cells})" \
+                "&!sandbox={sandbox_name}".format(cellset_id=cellset_id,
+                        top_rows=f";$top={top}" if top and not skip else "",
+                        cell_properties=",".join(cell_properties),
+                        filter_axis=filter_axis,
+                        select_member_properties=select_member_properties,
+                        expand_elem_properties=expand_elem_properties,
+                        top_cells=f";$top={top}" if top else "",
+                        skip_cells=f";$skip={skip}" if skip else "",
+                        sandbox_name=format_url(sandbox_name))
```
MariusWirtz, line 963: I think it's a legacy issue, but shouldn't we be using the `format_url` here instead of calling the `format` function on the string?
If we do enhance the `format_url` function to support sandboxing, we won't need to repeat the full formatting block.

TM1py/Services/CellService.py
andreyea: I guess, it's an option, but it feels a little ambiguous to append extra parameter in there.

MariusWirtz:
Hi @andreyea,

thanks a lot for your valuable contribution! Very much appreciated.
A requested a few minor changes on the code. Mostly related to style.

The tests we can do in a separate MR. In fact, I wouldn't mind to write them myself. I it would be good opportunity for me to familiarize myself with the new functions.

I am not fully sure about the changes in the `CellService` though. The new code creates quite a bit of code duplication. I would love to make it a bit more efficient code wise. 
Instead of adding the sandbox_name to the URL, is it perhaps possible to pass the sandbox name as an HTTP header? That would be great because the HTTP methods in the RestService all support the optional `headers` argument. 
```python
    @tidy_cellset
    def extract_cellset_values(self, cellset_id: str, sandbox_name: str = None, **kwargs) -> List[Union[str, float]]:
        """ Extract cellset data and return only the cells and values

        :param cellset_id: String; ID of existing cellset
        :param sandbox_name: 
        :return: Raw format from TM1.
        """
        url = format_url("/api/v1/Cellsets('{}')?$expand=Cells($select=Value)", cellset_id)
        response = self._rest.GET(url=url, headers={"Sandbox": sandbox_name} if sandbox_name else None, **kwargs)
        return [cell["Value"] for cell in response.json()["Cells"]]
```

I think the second-best option would be to handle it in the `format_url` function.

What do you think?

MariusWirtz:
Fixes #216

andreyea:
Hi @MariusWirtz,

I will implement the changes you suggested above. 

In regards to the URL, I am not aware of a sandbox header in TM1, however, I can look into amending format_url function to add !sandbox if needed.

TM1py/Services/CellService.py
```diff
@@ -278,53 +292,58 @@ def write_values_through_cellset(self, mdx: str, values: List, **kwargs) -> Resp
         
         :param mdx: Valid MDX Expression.
         :param values: List of values. The Order of the List/ Iterable determines the insertion point in the cellset.
+        :param sandbox_name: str
         :return: 
         """
-        cellset_id = self.create_cellset(mdx, **kwargs)
-        return self.update_cellset(cellset_id=cellset_id, values=values, **kwargs)
+        cellset_id = self.create_cellset(mdx=mdx,sandbox_name=sandbox_name **kwargs)
```
MariusWirtz, line 298: We are lacking a comma here

TM1py/Utils/Utils.py
```diff
@@ -414,6 +416,23 @@ def get_seconds_from_duration(time_str: str) -> int:
     seconds = (int(d) * 86400) + (int(h) * 3600) + (int(m) * 60) + int(s)
     return seconds
 
+def url_parameters_add(url, **kwargs: str) -> str:
```
MariusWirtz, line 419: I think it may a good investment to add some unit tests for this function.

As a reference, you can check out the `format_url` tests in the MdxUtils Tests:
https://github.com/cubewise-code/tm1py/blob/master/Tests/MDXUtils.py

MariusWirtz:
> Hi @MariusWirtz,
> 
> I will implement the changes you suggested above.
> 
> In regards to the URL, I am not aware of a sandbox header in TM1, however, I can look into amending format_url function to add !sandbox if needed.

Nice one. Thanks. I like the new `url_parameters_add` function.

I see two topics left. 
1. Add Testing. I'm happy to add this is in a separate MR.
2. Solve Conflicts. I understand your branch is based on a version of the master from mid-July. There are quite a few conflicts that need to be solved before we can merge. If you don't mind I can take care of this since I am familiar with the changes we did over the last weeks.

andreyea:
Closes in favor of https://github.com/cubewise-code/tm1py/pull/374

PR #376, opened 2020-09-30
Add tests for Sandbox and SandboxService classes

MariusWirtz:


MariusWirtz:
@andreyea 
Please double-check my changes.

I didn't add tests for all existing `write` and `execute` functions. It seems disproportionate. Would you agree?

PR #377, opened 2020-09-30
Cleaned up cells tests

scrambldchannel:
Hi Marius

I had a go at tidying up the tests for the cells service. 

* Moved global vars to class properties
* Re-structured fixtures to ensure idempotency

Some of the tests were failing sporadically when I randomised the test order (I use the pytest-randomly plugin for this, which is simple and effective for flagging up these issues). The fixtures could probably do with some further housekeeping but I think this is a good start. Let me know what you think.

Cheers
Alex

(relates to https://github.com/cubewise-code/tm1py/issues/301)

### DIFF
--- a/Tests/Cell.py
+++ b/Tests/Cell.py
@@ -125,6 +125,8 @@ class TestCellMethods(unittest.TestCase):
         # For tests on string related methods
         cls.build_string_cube()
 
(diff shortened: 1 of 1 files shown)


MariusWirtz:
Thanks. Nice work. Good idea to randomize the test order.

We are getting there slowly but surely :-)

PR #384, opened 2020-10-05
Functions to extract specific bits from Updateable property on cells.

andreyea:
Two new functions in Utils:

**extract_bit** extracts specified bit from a decimal value.

**cell_is_updateable** checks if a cells has Updateable property and returns if the cell can be updated.


TM1py/Utils/Utils.py
```diff
@@ -801,3 +801,40 @@ def wrap_in_curly_braces(expression: str) -> str:
...
+    1 - security restriction
+    2 - updatable using UPDATE CUBE
+    3 - rule is applied
+    4 - picklist exists
+    5 - sandbox value is different to base
+    9 - no spreading hold
+    10 - leaf hold
+    11 - consolidation spreading hold
+    12 - temporary spreading hold
+    29 - cell is not updateable
```
MariusWirtz, line 821: Do you think we can outsource these numbers into an Enum?

andreyea:
Yes, if missing values is not an issue. Do you know if any other bits are
being used?

TM1py/Utils/Utils.py
```diff
@@ -801,3 +801,40 @@ def wrap_in_curly_braces(expression: str) -> str:
...
+    bit = (decimal_value & (1 << position - 1)) != 0
+    return bit
+
+
+def cell_is_updateable(cell: dict) -> bool:
+    """ Function checks if the cell can be updated
+    :param cell: dict cell including Updateable property
+    :return: bool
+    """
+    if cell.get("Updateable", False):
```
rkvinoth, line 837: `, False` is not needed.

MariusWirtz:
Hi @andreyea,

Thanks for your contribution! I just realized that there are conflicts that need to be resolved before I can merge.

PR #388, opened 2020-10-13
Split subsets tests into two and a couple of other things

scrambldchannel:
Hi Marius

I was looking at the subset code and tried a few things:

* split tests into Subset.py (just tests the objects) and SubsetService.py (tests the service)
* aligned the Subset from_dict method with the constructor to support missing values for optional args
* renamed the class in ```Tests/Sandbox.py```
* updated the ```Tests/__init__.py``` to include a few new files

Let me know what you think  :)

Cheers
Alex


MariusWirtz:
Hi @scrambldchannel,

nice work! The code looks all good.
Thanks for the contribution and the fixes on my sandbox tests 👍

PR #418, opened 2020-11-10
support row_axis_only MDX queries

rkvinoth:
code to test:

```python


tm1 = TM1Service()

mdx = """
        SELECT NON EMPTY
        {TM1SUBSETALL( [}Clients] )} * {TM1SUBSETALL( [}Groups] )}
        ON 0
        FROM [}ClientGroups]"""

csv = tm1.cubes.cells.execute_mdx_csv(mdx)

print(csv)
```

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -1395,7 +1395,7 @@ class CellService(ObjectService):
         rows, titles, columns = [], [], []
         if len(response_json["Axes"]) == 1:
             if response_json["Axes"][0]["Hierarchies"]:
-                rows = [hierarchy["UniqueName"] for hierarchy in response_json["Axes"][0]["Hierarchies"]]
+                columns = [hierarchy["UniqueName"] for hierarchy in response_json["Axes"][0]["Hierarchies"]]
         else:
             if response_json["Axes"][0]["Hierarchies"]:
                 columns = [hierarchy["UniqueName"] for hierarchy in response_json["Axes"][0]["Hierarchies"]]
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -85,12 +85,12 @@ def manage_transaction_log(func):
 
 class CellService(ObjectService):
     """ Service to handle Read and Write operations to TM1 cubes
-    
+
     """
 
     def __init__(self, tm1_rest: RestService):
         """
-        
+
         :param tm1_rest: instance of RestService
         """
         super().__init__(tm1_rest)
@@ -98,7 +98,7 @@ class CellService(ObjectService):
     def get_value(self, cube_name: str, element_string: str, dimensions: List[str] = None, sandbox_name: str = None,
                   **kwargs) -> Union[str, float]:
         """ Element_String describes the Dimension-Hierarchy-Element arrangement
-            
+
         :param cube_name: Name of the cube
         :param element_string: "Hierarchy1::Element1 && Hierarchy2::Element4, Element9, Element2"
             - Dimensions are not specified! They are derived from the position.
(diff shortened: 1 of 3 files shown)


MariusWirtz:
HI @rkvinoth 

the code looks good, but `On 0` means `on COLUMNS`. Some variable names in the changes should be different I think.

Tests/Cell.py
```diff
@@ -977,6 +977,45 @@ def test_execute_mdx_csv(self):
             self.total_value,
             sum(values))
 
+    def test_execute_mdx_csv_row_only(self):
```
MariusWirtz, line 980: should be `test_execute_mdx_csv_column_only`

Tests/Cell.py
```diff
@@ -1126,6 +1165,35 @@ def test_execute_mdx_dataframe(self):
             self.total_value,
             sum(values))
 
+    @skip_if_no_pandas
+    def test_execute_mdx_dataframe_row_only(self):
```
MariusWirtz, line 1169: Should be `test_execute_mdx_dataframe_column_only`

TM1py/Services/CellService.py
```diff
@@ -1393,10 +1393,14 @@ def extract_cellset_composition(self, cellset_id: str, sandbox_name: str = None,
         cube = response_json["Cube"]["Name"]
 
         rows, titles, columns = [], [], []
-        if response_json["Axes"][0]["Hierarchies"]:
-            columns = [hierarchy["UniqueName"] for hierarchy in response_json["Axes"][0]["Hierarchies"]]
-        if response_json["Axes"][1]["Hierarchies"]:
-            rows = [hierarchy["UniqueName"] for hierarchy in response_json["Axes"][1]["Hierarchies"]]
+        if len(response_json["Axes"]) == 1:
+            if response_json["Axes"][0]["Hierarchies"]:
+                rows = [hierarchy["UniqueName"] for hierarchy in response_json["Axes"][0]["Hierarchies"]]
```
MariusWirtz, line 1398: If only one axis is defined, it must be the columns, don't you think?

Tests/Cell.py
rkvinoth: Oh yeah! got confused while fixing it.

Tests/Cell.py
rkvinoth: Oh yeah! got confused while fixing it.

TM1py/Services/CellService.py
rkvinoth: Fixed it just now.

Tests/Cell.py
```diff
@@ -1599,7 +1667,7 @@ def test_execute_view_csv(self):
 
         # check if coordinates are the same
         self.assertTrue(coordinates.issubset(self.target_coordinates))
-        values = [float(record.split(',')[3]) for record in records if record != '']
+        values = [float(record.split(',')[4]) for record in records if record != '']
```
rkvinoth, line 1670: I'm seeing **index** getting returned part of the `execute_mdx_csv` method...

MariusWirtz: I'm sorry I don't understand 🤔

MariusWirtz: I think there is an issue here. The original test was alright. 

I just checked out this branch and debugged the test. Now the produced CSV doesn't map to the view data anymore.

image

TM1py/Utils/Utils.py
```diff
@@ -226,19 +226,21 @@ def build_csv_from_cellset_dict(
...
+        if column_axis and row_axis:
+            index_rows = ordinal // column_axis['Cardinality'] % row_axis['Cardinality']
+            csv_entry.extend(extract_element_names_from_members(row_axis['Tuples'][index_rows]['Members']))
+            index_columns = ordinal % column_axis['Cardinality']
+            csv_entry.extend(extract_element_names_from_members(column_axis['Tuples'][index_columns]['Members']))
         if column_axis:
-            index_columns = ordinal % row_axis['Cardinality']
-            csv_entry.extend(extract_element_names_from_members(row_axis['Tuples'][index_columns]['Members']))
+            index_rows = ordinal % column_axis['Cardinality']
+            csv_entry.extend(extract_element_names_from_members(column_axis['Tuples'][index_rows]['Members']))
```
MariusWirtz, line 243: I think the issue from the tests may originate here. I think we are adding the members from the column axis to the `csv_entry` twice. Should it perhaps say `elif` instead of `if`?

MariusWirtz:
Nice one! Thanks for the MR.

PR #436, opened 2020-12-07
Add add_component function to Hierarchy class

### LINKED ISSUE #433: Add ElementComponentAdd type of function
I want to add elements to consolidation. In TM1 I would use DimensionElementComponentAdd and not having too much experience with tm1py that's what I was looking for. 
It would be convenient to have a function that would combine
hierarchy.add_element and hierarchy.add_edge to get same result as mentioned TM1 function.

N/A

**Version**
- TM1py 1.5.0
- TM1 Server Version: 2.0.9.3

**Additional context**
N/A

MariusWirtz:
Add `add_component` function

Fixes #433

### DIFF
--- a/TM1py/Objects/Hierarchy.py
+++ b/TM1py/Objects/Hierarchy.py
@@ -142,6 +142,19 @@ class Hierarchy(TM1Object):
 
         self._elements[element_name] = Element(name=element_name, element_type=element_type)
 
+    def add_component(self, parent_name: str, component_name: str, weight: int):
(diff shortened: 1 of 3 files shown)


MariusWirtz:
Hi @macmaj,

please upgrade to this branch and check if the function works as expected.
you can upgrade with pip like this:
`pip install https://github.com/cubewise-code/tm1py/archive/feature/add-component.zip --upgrade`

Issue #443, opened 2020-12-22
Message log filter does not take list as parameter for msg_contains


scrumthing:
**Describe the bug**
The function **get_message_log_entries** excepts Iterable as variable. But if the example below is used no entries are returned. If only one of them is used (depending on the language of the server) entries are found. 

**To Reproduce**
```python
with TM1Service(**TM1_PARAMETERS) as tm1:
    messagelog = tm1.server.get_message_log_entries(reverse=False, logger= 'TM1.Server' , msg_contains= ['TM1 Server is ready','TM1 Server ist bereit,'])
```

**Expected behavior**
Entries should show up if only on of the entries in the list is valid

**Version**
Current GitHub Master
TM1 Build Number:  11.8.00200.24

rkvinoth:
The above PR will solve the issue and I believe this is how it should work instead of searching for all the wildcards in a single `Message` row.

scrumthing:
Sounds good for me! :-)

PR #444, opened 2020-12-23
msg_contains should filter rows containing the wildcards

rkvinoth:
#443 fix

wildcards in `msg_contains` should be applied on multiple rows.

Currently it is building `and` conditions and looking for all the wildcards in the same row.

### DIFF
--- a/TM1py/Services/ServerService.py
+++ b/TM1py/Services/ServerService.py
@@ -121,10 +121,11 @@ class ServerService(ObjectService):
 
             if msg_contains:
                 if isinstance(msg_contains, str):
-                    log_filters.append(format_url("contains(Message,'{}')", msg_contains))
+                    log_filters.append(format_url("contains(toupper(Message),toupper('{}'))", msg_contains))
                 else:
-                    msg_filters = [format_url("contains(Message,'{}')", wildcard) for wildcard in msg_contains]
-                    log_filters.append("({})".format(" and ".join(msg_filters)))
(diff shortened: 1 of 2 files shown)


TM1py/Services/ServerService.py
```diff
@@ -121,10 +121,11 @@ def get_message_log_entries(self, reverse: bool = True, since: datetime = None,
...
             if msg_contains:
                 if isinstance(msg_contains, str):
-                    log_filters.append(format_url("contains(Message,'{}')", msg_contains))
+                    log_filters.append(format_url("contains(toupper(Message),toupper('{}'))", msg_contains))
                 else:
-                    msg_filters = [format_url("contains(Message,'{}')", wildcard) for wildcard in msg_contains]
-                    log_filters.append("({})".format(" and ".join(msg_filters)))
+                    msg_filters = [format_url("contains(toupper(Message),toupper('{}'))", wildcard) 
+                                   for wildcard in msg_contains]
+                    log_filters.append("({})".format(" or ".join(msg_filters)))
```
MariusWirtz, line 128: Now we are not in line with the docstring of the function (`list of substrings will be queried as AND statement`) and the change may break compatibility.

Perhaps we can add an optional argument (something like `operator='AND'`) to the function to control if AND or OR should be used. What do you think?

rkvinoth: Yeah that should be useful!

MariusWirtz:
Created a new MR based on this one #449

PR #449, opened 2021-01-07
Issue/msg contains

### LINKED ISSUE #443: Message log filter does not take list as parameter for msg_contains

MariusWirtz:
Fixes #443

### DIFF
--- a/TM1py/Services/ServerService.py
+++ b/TM1py/Services/ServerService.py
@@ -78,7 +78,8 @@ class ServerService(ObjectService):
     @require_admin
     def get_message_log_entries(self, reverse: bool = True, since: datetime = None,
                                 until: datetime = None, top: int = None, logger: str = None,
-                                level: str = None, msg_contains: Iterable = None, **kwargs) -> Dict:
+                                level: str = None, msg_contains: Iterable = None, msg_contains_operator: str = 'and',
+                                **kwargs) -> Dict:
         """
         :param reverse: Boolean
         :param since: of type datetime. If it doesn't have tz information, UTC is assumed.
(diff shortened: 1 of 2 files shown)


TM1py/Services/ServerService.py
```diff
@@ -87,10 +88,15 @@ def get_message_log_entries(self, reverse: bool = True, since: datetime = None,
         :param logger: string, eg TM1.Server, TM1.Chore, TM1.Mdx.Interface, TM1.Process
         :param level: string, ERROR, WARNING, INFO, DEBUG, UNKNOWN
         :param msg_contains: iterable, find substring in log message; list of substrings will be queried as AND statement
+        :param msg_contains_operator: 'and' or 'or'
 
         :param kwargs:
         :return: Dict of server log
         """
+        msg_contains_operator = msg_contains_operator.strip().lower()
+        if msg_contains_operator not in ("and", "or"):
```
rkvinoth, line 97: Can we make this case insensitive?

MariusWirtz: it is :)

rkvinoth: haha didn't notice the previous line :D

rkvinoth: looks good from my end.

MariusWirtz: Thanks

PR #450, opened 2021-01-08
Feature enable sandbox unbound ti writeback

rclapp:
enabled unbound TI writeback method to support sandboxes.


MariusWirtz:
@rclapp Thanks for this change! Very much appreciated. I didn't know the sandbox could be controlled from TI.

I will add a couple of comments to the code now.

TM1py/Services/CellService.py
```diff
@@ -458,18 +466,21 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
-        process = Process(name="", prolog_procedure="\r".join(chunk))
+        process = Process(name="", prolog_procedure = enable_sandbox, epilog_procedure="\r".join(chunk))
         success, _, _ = self.execute_unbound_process(process, **kwargs)
         successes.append(success)
 
         if not all(successes):
             raise TM1pyException(f"{len(successes) - sum(successes)} out of {len(successes)} unbound processes failed")
 
+        if not all(successes):
+            raise TM1pyException(f"{len(successes) - sum(successes)} out of {len(successes)} unbound processes failed")
```
MariusWirtz, line 482: This is duplicated

TM1py/Services/CellService.py
```diff
@@ -411,7 +411,15 @@ def write(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str]
     def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, increment: bool = False,
                                       sandbox_name: str = None, precision=8, **kwargs):
         if sandbox_name:
-            raise NotImplementedError("Function does not support writing to sandboxes yet")
+            enable_sandbox = f"IF(ServerSandboxExists('{sandbox_name}')=1); " \
```
MariusWirtz, line 414: I am thinking that perhaps we could do this check through the `SandboxService` if a sandbox exists, while still in python.
Then we wouldn't need to do it in every TI and we could throw a descriptive ValueError or something in python.

I am thinking this because I am currently working on a small change to add 16K CellPut lines to the prolog and another 16K into the epilog.
Our changes could be merged more harmoniously if the sandbox checks were not happening in the prolog.

MariusWirtz:
@rclapp 
I merged our efforts in #451 please review and approve :)

TM1py/Services/CellService.py
MariusWirtz: Moved to #451

PR #452, opened 2021-01-11
support for sandboxes in mdx clear

rclapp:
Working on some data simulation projects, as I find methods that need sandbox support, I'll add it.

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -83,6 +83,26 @@ def manage_transaction_log(func):
 
     return wrapper
 
+def manage_changeset(func):
+    """ Control the start and end of change sets which goups write events together in the TM1 transaction log.
+
+    Decorated function working with all non-TI based writing methods
+    """
+
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        if not kwargs.get("use_changeset", False):
+            try:
+                changeset = self.begin_changeset()
+                kwargs["changeset"]=changeset
+                return func(self, *args, **kwargs)
+            finally:
+                self.end_changeset(changeset)
+        else:
+            return func(self, *args, **kwargs)
+
+    return wrapper
+
 
 class CellService(ObjectService):
     """ Service to handle Read and Write operations to TM1 cubes
@@ -312,9 +332,11 @@ class CellService(ObjectService):
         dimensions = cube_service.get_dimension_names(cube_name, True, **kwargs)
         return dimensions
 
+
     @require_pandas
     def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iterable[str] = None,
                         sandbox_name: str = None,
+                        use_changeset: bool = True,
                         **kwargs) -> Response:
         """
         Function expects same shape as `execute_mdx_dataframe` returns.
@@ -340,7 +362,7 @@ class CellService(ObjectService):
         cells = build_cellset_from_pandas_dataframe(data)
 
         return self.write_values(cube_name=cube_name, cellset_as_dict=cells, dimensions=dimensions,
-                                 sandbox_name=sandbox_name, **kwargs)
+                                 sandbox_name=sandbox_name, use_changeset=use_changeset, **kwargs)
 
     def write_value(self, value: Union[str, float], cube_name: str, element_tuple: Iterable,
                     dimensions: Iterable[str] = None, sandbox_name: str = None, **kwargs) -> Response:
@@ -367,9 +389,10 @@ class CellService(ObjectService):
         data = json.dumps(body_as_dict, ensure_ascii=False)
         return self._rest.POST(url=url, data=data, **kwargs)
 
+
     def write(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str] = None, increment: bool = False,
               deactivate_transaction_log: bool = False, reactivate_transaction_log: bool = False,
-              sandbox_name: str = None, use_ti=False, **kwargs):
+              sandbox_name: str = None, use_ti=False, use_changeset: bool=True, **kwargs):
         """ Write values to a cube
 
         Same signature as `write_values` method, but faster since it uses `write_values_through_cellset`
@@ -416,6 +439,7 @@ class CellService(ObjectService):
             deactivate_transaction_log=deactivate_transaction_log,
             reactivate_transaction_log=reactivate_transaction_log,
             sandbox_name=sandbox_name,
+            use_changeset=True,
             **kwargs)
 
     @require_admin
@@ -507,9 +531,10 @@ class CellService(ObjectService):
 
         return process_service.execute_process_with_return(process, **kwargs)
 
+    @manage_changeset
     @manage_transaction_log
     def write_values(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str] = None,
-                     sandbox_name: str = None, **kwargs) -> Response:
+                     sandbox_name: str = None, use_changeset:bool = True, **kwargs) -> Response:
         """ Write values to a cube
 
         For cellsets with > 1000 cells look into `write` or `write_values_through_cellset`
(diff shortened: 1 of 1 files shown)


MariusWirtz:
@rclapp 
I just realized that when the `ServerActiveSandboxSet`  function is called, the effect persists after the completion of the process. So subsequent processes would possibly be working on a sandbox unless they explicitly state otherwise in the code.

I think it may make sense to change the active sandbox back to `base` at the end of the epilog.
What do you think?

TM1py/Services/CellService.py
```diff
@@ -1800,7 +1834,40 @@ def activate_transactionlog(self, *args: str, **kwargs) -> Response:
...
+
+    def begin_changeset(self) -> str:
+        """ begin a change set
+
+        :return: Change set ID
+        """
+
+        url = "/api/v1/BeginChangeSet"
+        return self._rest.POST(url).json()['value']
+        return change_set_id
```
MariusWirtz, line 1847: code can't be reached

TM1py/Services/CellService.py
```diff
@@ -1800,7 +1834,40 @@ def activate_transactionlog(self, *args: str, **kwargs) -> Response:
...
+    def end_changeset(self, change_set: str) -> Response:
+        """end a change set
+
+        :return: Change set ID
+        """
+
+        url = "/api/v1/EndChangeSet"
+        data = {"ChangeSetID":change_set}
+        return self._rest.POST(url, data=json.dumps(data, ensure_ascii=False))
+        return response
```
MariusWirtz, line 1858: can't be reached

TM1py/Services/CellService.py
```diff
@@ -357,9 +389,10 @@ def write_value(self, value: Union[str, float], cube_name: str, element_tuple: I
         data = json.dumps(body_as_dict, ensure_ascii=False)
         return self._rest.POST(url=url, data=data, **kwargs)
 
+
     def write(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str] = None, increment: bool = False,
               deactivate_transaction_log: bool = False, reactivate_transaction_log: bool = False,
-              sandbox_name: str = None, use_ti=False, **kwargs):
+              sandbox_name: str = None, use_ti=False, use_changeset: bool=True, **kwargs):
```
MariusWirtz, line 395: I think it would be good to throw a ValueError here if `use_ti` is `True `and `use_changeset` is `True`

TM1py/Services/CellService.py
```diff
@@ -497,9 +531,10 @@ def execute_unbound_process(self, process: Process, **kwargs) -> Tuple[bool, str
 
         return process_service.execute_process_with_return(process, **kwargs)
 
+    @manage_changeset
     @manage_transaction_log
     def write_values(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str] = None,
-                     sandbox_name: str = None, **kwargs) -> Response:
+                     sandbox_name: str = None, use_changeset:bool = True, **kwargs) -> Response:
```
MariusWirtz, line 537: I understand that to use changesets the user would have to pass `use_changeset` as `True` and a random str value for `changeset` through `kwargs`. Is that correct? 

Perhaps instead of the additional boolean `use_changeset` argument, you could just check if the str `changeset` argument is not None. Do you think that would work?

rclapp:
Yes, I think that makes sense, but only for bound TIs. Unbound wont live beyond execution.

Sent from my mobile device.

On Jan 12, 2021 1:08 AM, Marius Wirtz <notifications@github.com> wrote:

I just realized that when the ServerActiveSandboxSet function is called, the effect persists after the completion of the process. So subsequent processes would possibly be working on a sandbox unless they explicitly state otherwise in the code.

I think it may make sense to change the active sandbox back to base at the end of the epilog.
What do you think?

-

TM1py/Services/CellService.py
```diff
@@ -83,6 +83,26 @@ def wrapper(self, *args, **kwargs):
...
+            try:
+                changeset = self.begin_changeset()
+                kwargs["changeset"]=changeset
+                return func(self, *args, **kwargs)
+            finally:
+                self.end_changeset(changeset)
+        else:
+            return func(self, *args, **kwargs)
+
+    return wrapper
```
MariusWirtz, line 104: Nice one! I like to use a decorator for this :)

rclapp:
The idea here is that typically users won't want to manage the change set on their own.   Functions that have the change set decorator and have change set set to true will auto generate a change set, complete the write back, and end the change set.

We could make it much simpler and just accept a change set Id as an optional parameter. However I was thinking of it being  closely related to Transaction log management, so I used function wrappers.

The write functions could use some clean up and that might help simplify this. For example dataframe and transaction log updates go through write vales, but the generic write uses write via cell sets. Streamlining the entry point for write operations might help since I struggled to land on a solid design.

Sent from my mobile device.

On Jan 12, 2021 1:21 AM, Marius Wirtz <notifications@github.com> wrote:

In TM1py/Services/CellService.py<https://github.com/cubewise-code/tm1py/pull/452#discussion_r555621371>:

>      @manage_transaction_log
     def write_values(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str] = None,
-                     sandbox_name: str = None, **kwargs) -> Response:
+                     sandbox_name: str = None, use_changeset:bool = True, **kwargs) -> Response:


I understand that to use changesets the user would have to pass use_changeset as True and a random str value for changeset through kwargs. Is that correct?

Perhaps instead of the additional boolean use_changeset argument, you could just check if the str changeset argument is not None. Do you think that would work?

-

MariusWirtz:
> Yes, I think that makes sense, but only for bound TIs. Unbound wont live beyond execution. 

Thanks. You are right. For unbound processes, it does not live beyond execution.

MariusWirtz:
> The write functions could use some clean up and that might help simplify this. For example dataframe and transaction log updates go through write vales, but the generic write uses write via cell sets. Streamlining the entry point for write operations might help since I struggled to land on a solid design.

Agree. We should try to get this right before the 1.6 release, which I wanted to release today :)
I think the new `write` function should be the new go-to function. In fact, we could add a deprecated warning to `write_values` function. It is slower and less flexible than the new `write` function anyway.
Since the `write` function hasn't been included in 1.5, we could set the `use_changeset=True` by default without breaking compatibility.


> The idea here is that typically users won't want to manage the change set on their own.   Functions that have the change set decorator and have change set set to true will auto generate a change set, complete the write back, and end the change set.

I see. That makes sense. I think we can use a changeset by default in the new `write` function but if we include it in `write_values_through_cellset` we may break compatibility.

> We could make it much simpler and just accept a change set Id as an optional parameter. However I was thinking of it being  closely related to Transaction log management, so I used function wrappers.

OK it all makes sense. I understand now that you get the changeset_id in the wrapper function.

TM1py/Services/CellService.py
```diff
@@ -330,7 +362,7 @@ def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iter
         cells = build_cellset_from_pandas_dataframe(data)
 
         return self.write_values(cube_name=cube_name, cellset_as_dict=cells, dimensions=dimensions,
```
MariusWirtz, line 364: I think it makes sense to call the new `write` function here instead. Then we could expose the `use_ti` argument also to the `write_dataframe` function.
What do you think?

TM1py/Services/CellService.py
MariusWirtz: Resolved

TM1py/Services/CellService.py
rclapp: Yes, that makes sense to me that would be better

TM1py/Services/CellService.py
rclapp: I will add that

rclapp:
Making a new pull request with cleaned up code

PR #453, opened 2021-01-12
Feature support changesets

rclapp:
Added a function wrapper to support change sets


MariusWirtz:
closed in favor of #452

I understand that both MR contain the same commits. Please reopen if I got something wrong.

Issue #372, opened 2020-09-29
Support for change sets


rclapp:
Enhancement

Add support for change sets to write functions and write methods for undo change sets

PR #454, opened 2021-01-13
Cleanup cellservice writeback changesets

rclapp:
Cell Service Clean Up
1) Added deprecation notice to write values, removed dependency on change set wrapper
2) Update transaction log methods to use write_value instead of write_values
3) Changed write_dataframe to use the new write function to support TI and changesets
4) Added exception when TI and Change sets are used at the same time

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -18,6 +18,7 @@ from TM1py.Services.ObjectService import ObjectService
 from TM1py.Services.RestService import RestService
 from TM1py.Services.SandboxService import SandboxService
 from TM1py.Services.ViewService import ViewService
+from TM1py.Services.ProcessService import ProcessService
 from TM1py.Utils import Utils, CaseAndSpaceInsensitiveSet, format_url, add_url_parameters
 from TM1py.Utils.Utils import build_pandas_dataframe_from_cellset, dimension_name_from_element_unique_name, \
     CaseAndSpaceInsensitiveDict, wrap_in_curly_braces, CaseAndSpaceInsensitiveTuplesDict, abbreviate_mdx, \
@@ -340,7 +341,7 @@ class CellService(ObjectService):
                         sandbox_name: str = None,
                         use_ti: bool = False,
                         use_changeset: bool = True,
-                        **kwargs) -> Response:
+                        **kwargs) -> Tuple[bool, list, str]:
         '''
         Function expects same shape as `execute_mdx_dataframe` returns.
         Column order must match dimensions in the target cube with an additional column for the values.
@@ -355,7 +356,7 @@ class CellService(ObjectService):
         :param use_ti:
         :param use_changeset:
         :param kwargs:
-        :return:
+        :return: Success: bool, Messages: list, ChangeSet: str
         '''
         if not isinstance(data, pd.DataFrame):
             raise ValueError("argument 'data' must of type DataFrame")
@@ -407,7 +408,7 @@ class CellService(ObjectService):
 
     def write(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str] = None, increment: bool = False,
               deactivate_transaction_log: bool = False, reactivate_transaction_log: bool = False,
-              sandbox_name: str = None, use_ti=False, use_changeset: bool=True, **kwargs):
+              sandbox_name: str = None, use_ti=False, use_changeset: bool=True, **kwargs) -> Tuple[bool, list, str]:
         """ Write values to a cube
 
         Same signature as `write_values` method, but faster since it uses `write_values_through_cellset`
@@ -424,8 +425,9 @@ class CellService(ObjectService):
         :param reactivate_transaction_log: reactivate after writing
         :param sandbox_name: str
         :param use_ti: Use unbound process to write. Requires admin permissions. causes massive performance improvement.
-        :return:
+        :return: Success: bool, Message: list, ChangeSet: str
         """
+
         if use_ti:
             if use_changeset:
                 raise TM1pyException("Writeback via unbound TI processes do not support changesets")
@@ -449,20 +451,30 @@ class CellService(ObjectService):
             values.append(value)
         mdx = query.to_mdx()
 
-        self.write_values_through_cellset(
+        return self.write_values_through_cellset(
             mdx=mdx,
             values=values,
             increment=increment,
             deactivate_transaction_log=deactivate_transaction_log,
             reactivate_transaction_log=reactivate_transaction_log,
             sandbox_name=sandbox_name,
-            use_changeset=True,
+            use_changeset=use_changeset,
             **kwargs)
 
     @require_admin
     @manage_transaction_log
     def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, increment: bool = False,
-                                      sandbox_name: str = None, precision=8, **kwargs):
+                                      sandbox_name: str = None, precision=8, **kwargs) -> Tuple[bool, list, str]:
+        '''
+        Writes data back to TM1 via an unbound TI process
+        :param cube_name:
+        :param cellset_as_dict:
+        :param increment:
+        :param sandbox_name:
+        :param precision:
+        :param kwargs:
+        :return: Success: bool, Messages: list, ChangeSet: None
+        '''
         if sandbox_name:
             if not self.sandbox_exists(sandbox_name):
                 raise ValueError(f"Sandbox '{sandbox_name}' does not exist")
@@ -473,6 +485,8 @@ class CellService(ObjectService):
 
         successes = list()
         statements = list()
+        messages = list()
+        changeset = None
 
         cube_service = self.get_cube_service()
         element_service = self.get_element_service()
@@ -488,9 +502,10 @@ class CellService(ObjectService):
             # use default 'Numeric' so that not existing elements trigger minor error during TI execution
             element_type = measure_dimension_elements.get(coordinates[-1], 'Numeric')
 
-            if element_type == 'String':
+            if element_type == 'String' or isinstance(value, str):
                 function_str = 'CellPutS('
                 value_str = f"'{value}'"
+
 
             # by default assume numeric, to trigger minor errors on write operations to C elements
             else:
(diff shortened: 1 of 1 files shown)


rclapp:
@MariusWirtz I also wanted to standardize the return types of the 2 main Write functions (Dataframe, and Write). When using ChangeSets it would be useful to return the changeset id so a user could call UndoChangeSet. I was thinking of using a similar pattern to TI processers, where the functions return 3 obects
1) write successful: bool
2) changeset: str
3) response: Response

Before I get into it, what is your thoughts?

TM1py/Services/CellService.py
```diff
@@ -406,6 +456,7 @@ def write(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str]
             deactivate_transaction_log=deactivate_transaction_log,
             reactivate_transaction_log=reactivate_transaction_log,
             sandbox_name=sandbox_name,
+            use_changeset=True,
```
MariusWirtz, line 459: this should say `use_changeset=use_changeset` I think

MariusWirtz:
> @MariusWirtz I also wanted to standardize the return types of the 2 main Write functions (Dataframe, and Write). When using ChangeSets it would be useful to return the changeset id so a user could call UndoChangeSet. I was thinking of using a similar pattern to TI processers, where the functions return 3 obects
> 
> 1. write successful: bool
> 2. changeset: str
> 3. response: Response
> 
> Before I get into it, what is your thoughts?

I like this idea! Now is a good moment to implement it before the two functions are officially released. 

The `write` function doesn't provide a single response object to return if the `use_ti` option is used, but I guess then the response can just be None. Or how would you handle that case?

rclapp:
I was thinking of structuring the return as follows

WriteSuccess: bool
ChangeSetID: string/none
WriteMessages (for TI content of log file, or response body for a direct cell update): string/none

Sent from my mobile device.

On Jan 13, 2021 12:02 AM, Marius Wirtz <notifications@github.com> wrote:

@MariusWirtz<https://github.com/MariusWirtz> I also wanted to standardize the return types of the 2 main Write functions (Dataframe, and Write). When using ChangeSets it would be useful to return the changeset id so a user could call UndoChangeSet. I was thinking of using a similar pattern to TI processers, where the functions return 3 obects

  1.  write successful: bool
  2.  changeset: str
  3.  response: Response

Before I get into it, what is your thoughts?

I like this idea! Now is a good moment to implement it before the two functions are officially released.

The write function doesn't provide a single response object to return if the use_ti option is used, but I guess then the response can just be None. Or how would you handle that case?

-

scrumthing:
Very cool improvement @rclapp. Had no idea of changesets before, looks very promising. Now I wonder far you can go back in time to reverse a change via the changeset id.

MariusWirtz:
> I was thinking of structuring the return as follows 
> WriteSuccess: bool 
> ChangeSetID: string/none 
> WriteMessages (for TI content of log file, or response body for a direct cell update): string/none

Sounds good. As the third return value, we may want to return a list by default. When multiple unbound processes are used, behind the scenes, we may need to return a couple of ti-log file references.

rclapp:
Sounds good.

Sent from my mobile device.

On Jan 13, 2021 2:51 AM, Marius Wirtz <notifications@github.com> wrote:

I was thinking of structuring the return as follows
WriteSuccess: bool
ChangeSetID: string/none
WriteMessages (for TI content of log file, or response body for a direct cell update): string/none

Sounds good. As the third return value, we may want to return a list by default. When multiple unbound processes are used, behind the scenes, we may need to return a couple of ti-log file references.

-

MariusWirtz:
Fixes #372

TM1py/Services/CellService.py
rclapp: fixed

TM1py/Services/CellService.py
```diff
@@ -437,10 +501,11 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
             # use default 'Numeric' so that not existing elements trigger minor error during TI execution
             element_type = measure_dimension_elements.get(coordinates[-1], 'Numeric')
 
-            if element_type == 'String':
+            if element_type == 'String' or isinstance(value, str):
```
MariusWirtz, line 504: Why do we check with `isinstance` ?
Shouldn't the element type of the measure element determine whether it's CellPutN or CellPutS exclusively?

TM1py/Services/CellService.py
```diff
@@ -469,19 +534,25 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
-        success, _, _ = self.execute_unbound_process(process, **kwargs)
+        success, status, error_log_file = self.execute_unbound_process(process, **kwargs)
         successes.append(success)
+        if error_log_file:
+            message = self.get_error_log_file_content(file_name=error_log_file)
+            messages.append(message)
 
-        if not all(successes):
-            raise TM1pyException(f"{len(successes) - sum(successes)} out of {len(successes)} unbound processes failed")
+        return all(successes), messages, changeset
```
MariusWirtz, line 555: I imagine retrieving all error_log_files may slow down the function.
Perhaps you could just return the name of the error_log_files and leave it to the caller to read the details and understand what went wrong in detail.

MariusWirtz:
I find the new approach elegant, but I'm having second thoughts on not raising a `TM1pyException` when an operation fails 🤔

Normally in TM1py when any operation fails, a `TM1pyException` is raised.
This way failures can never go unnoticed and nothing fails silently. In the `execute_with_return` function we sort of moved away from that design pattern, but that was mostly because TI processes can be "halfway successful".

Now if a TM1py user uses the `write` function carelessly, and doesn't check explicitly for the `success` return value, fails may go unnoticed.

@rclapp 
What do you think about my concerns? 

Any third opinions on that matter?

TM1py/Services/CellService.py
rclapp: We can pass back a list of log files instead. I was trying to bring message parity to the different write methods. Messages from MDX updates don't need to be "unpacked"

TM1py/Services/CellService.py
rclapp: This was an interesting one. In my test I was writing back the word "Test" into a string cell. On the failure test case set the cellset to use an invalid measure, "Amounts" instead of "Amount". Then when it tries to retrieve the type of that element it can't, and instead returns numeric. Therefore the TI statements take the form "CellPutN(Test,Dim1,Dim2)". This statement results in a very odd TI error that says it can't find the variable "Test", because it is not wrapped in single quotes. By checking the instance of the value the TI reads "CellPutS('Test', Dim1, Dim2)" which returns a much clearer error KeyError can't find element Amounts in the dimension measures.

rclapp:
> I find the new approach elegant, but I'm having second thoughts on not raising a `TM1pyException` when an operation fails 🤔
> 
> Normally in TM1py when any operation fails, a `TM1pyException` is raised.
> This way failures can never go unnoticed and nothing fails silently. In the `execute_with_return` function we sort of moved away from that design pattern, but that was mostly because TI processes can be "halfway successful".
> 
> Now if a TM1py user uses the `write` function carelessly, and doesn't check explicitly for the `success` return value, fails may go unnoticed.
> 
> @rclapp
> What do you think about my concerns?
> 
> Any third opinions on that matter?

I understand the concern, if a write fails then it should be an exception. The challenge is that TI based writes can partially fail. Maybe we need a TM1Py exception that matches TIs exit codes (Failed, Aborted, Ran with messages, Ran with Minor Errors, Complete). That would bring execute with return inline with everything else. We could also make Write Exceptions, That way we could return them and the developer could handle them.

- MDX Write Success -> True,ChangeSetID
- MDX Write Fail -> TM1py Exception with the REST Response Error
- TI Write Success -> True,None
- TI Write All Failed -> TM1pyTIAbortedException(include the status and list of logs)
- TI Write Some Processes Failed -> TM1pyTIWriteIncompleteException (Status of failed writes and the logs)
- TI Write Some/All Processes had Minor Errors or Messages ->TM1pyTIWriteIncompleteException (Status of failed writes and the logs)

scrumthing:
I agree with your idea. Bring them up as exceptions sounds like a good approach. Especially for people not that familiar with tm1 it could be hard to tell the difference. Faced that with some colleagues from the sql department recently.

TM1py/Services/CellService.py
MariusWirtz: OK I see. I would prefer to just return the log_file_names to not lose performance.

TM1py/Services/CellService.py
MariusWirtz: OK. Makes sense. Then let's leave it like that :)

MariusWirtz:
> > I find the new approach elegant, but I'm having second thoughts on not raising a `TM1pyException` when an operation fails 🤔
> > Normally in TM1py when any operation fails, a `TM1pyException` is raised.
> > This way failures can never go unnoticed and nothing fails silently. In the `execute_with_return` function we sort of moved away from that design pattern, but that was mostly because TI processes can be "halfway successful".
> > Now if a TM1py user uses the `write` function carelessly, and doesn't check explicitly for the `success` return value, fails may go unnoticed.
> > @rclapp
> > What do you think about my concerns?
> > Any third opinions on that matter?
> 
> I understand the concern, if a write fails then it should be an exception. The challenge is that TI based writes can partially fail. Maybe we need a TM1Py exception that matches TIs exit codes (Failed, Aborted, Ran with messages, Ran with Minor Errors, Complete). That would bring execute with return inline with everything else. We could also make Write Exceptions, That way we could return them and the developer could handle them.
> 
> * MDX Write Success -> True,ChangeSetID
> * MDX Write Fail -> TM1py Exception with the REST Response Error
> * TI Write Success -> True,None
> * TI Write All Failed -> TM1pyTIAbortedException(include the status and list of logs)
> * TI Write Some Processes Failed -> TM1pyTIWriteIncompleteException (Status of failed writes and the logs)
> * TI Write Some/All Processes had Minor Errors or Messages ->TM1pyTIWriteIncompleteException (Status of failed writes and the logs)

Yeah, we should define specific Exceptions for these partial successes as you said. I like this idea.
Can I support you with the implementation?

rclapp:
Yes please, more hands will go faster

Sent from my mobile phone

On Jan 14, 2021 4:14 AM, Marius Wirtz <notifications@github.com> wrote:

I find the new approach elegant, but I'm having second thoughts on not raising a TM1pyException when an operation fails ?
Normally in TM1py when any operation fails, a TM1pyException is raised.
This way failures can never go unnoticed and nothing fails silently. In the execute_with_return function we sort of moved away from that design pattern, but that was mostly because TI processes can be "halfway successful".
Now if a TM1py user uses the write function carelessly, and doesn't check explicitly for the success return value, fails may go unnoticed.
What do you think about my concerns?
Any third opinions on that matter?

I understand the concern, if a write fails then it should be an exception. The challenge is that TI based writes can partially fail. Maybe we need a TM1Py exception that matches TIs exit codes (Failed, Aborted, Ran with messages, Ran with Minor Errors, Complete). That would bring execute with return inline with everything else. We could also make Write Exceptions, That way we could return them and the developer could handle them.

  *   MDX Write Success -> True,ChangeSetID
  *   MDX Write Fail -> TM1py Exception with the REST Response Error
  *   TI Write Success -> True,None
  *   TI Write All Failed -> TM1pyTIAbortedException(include the status and list of logs)
  *   TI Write Some Processes Failed -> TM1pyTIWriteIncompleteException (Status of failed writes and the logs)
  *   TI Write Some/All Processes had Minor Errors or Messages ->TM1pyTIWriteIncompleteException (Status of failed writes and the logs)

Yeah, we should define specific Exceptions for these partial successes as you said. I like this idea.
Can I support you with the implementation?

-

MariusWirtz:
> Yes please, more hands will go faster

OK. I will get my hands dirty now :)

TM1py/Services/CellService.py
MariusWirtz: The change is harmful if numbers come in scientific notation. We must rely exclusively on the type of the measure element.
See this test: https://github.com/cubewise-code/tm1py/blob/a44842d5012f61d970f2b1b8d873e417e32184fe/Tests/CellService.py#L440

MariusWirtz:
> Yes please, more hands will go faster

Based on this branch, I started working on raising dedicated exceptions. You can cherry-pick my commit and continue from there if you like 9ea56534b70067478c0bb61532e97ebc8cda36aa
Alternatively, I would try to add test cases and finished things up next week.

MariusWirtz:
@rclapp `write_values` supports the spreading shortcuts and the other write functions don't. Unfortunately, we can't deprecate it 😒

MariusWirtz:
closed in favor of #462

PR #462, opened 2021-01-15
Feature/cleanup cellservice writeback changesets

MariusWirtz:
Finish up support for changesets and sandboxes in CellService

### DIFF
--- a/TM1py/Exceptions/Exceptions.py
+++ b/TM1py/Exceptions/Exceptions.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 # TM1py Exceptions are defined here
-from typing import Mapping
+from typing import Mapping, List
 
 
 class TM1pyTimeout(Exception):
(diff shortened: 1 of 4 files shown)


MariusWirtz:
Fixes #463

MariusWirtz:
@rclapp please review and approve

rclapp (commented):
@rkvinoth did we implement any parallel threading for the TI based data writes?

rclapp:
Sorry for the delay, I just got a new laptop and I have been setting it up. Looks and tests well.

rclapp (approved):

rkvinoth:
> @rkvinoth did we implement any parallel threading for the TI based data writes?

we were testing it and the results are looking good with default config `worker=6`
It would be nice if we can implement it.

One thing to note is that we should raise any exception before triggering the `ThreadPoolExecutor` if not then the exceptions would go unnoticed (script will complete immediately without throwing any erros).

Issue #470, opened 2021-01-25
PROPERTIES in MDX does not work


yyzz1010:
**Describe what did you try to do with TM1py**
in cube view mdx query using PROPERTIES can show attribute name in column/row, but when using tm1py querying the same mdx, column/row name still showing principal name. is it possible for tm1y to show attribute name when querying cube view mdx?

**Describe what's not working the way you expect**
Didn't get the expected result? Describe:
1. I've got a mdx like this:
```
SELECT 
  NON EMPTY 
   {[Currency].[Currency].[1],[Currency].[Currency].[2]} 
   PROPERTIES [Currency].[Currency].[SALES_DESC_EN]  ON COLUMNS , 
  NON EMPTY 
   {[Year Month].[Year Month].[2015^2015-01],[Year Month].[Year Month].[2015^2015-02]} 
  ON ROWS 
FROM [Ex Rate Table] 
```
2. in cube view, column names are showing "SALES_DESC_EN" attribute
![image](https://user-images.githubusercontent.com/45536868/105676777-f3952a80-5f25-11eb-932c-cfcd39336c08.png)

3. but in tm1py still showing principal name
![image](https://user-images.githubusercontent.com/45536868/105676351-5afeaa80-5f25-11eb-8579-bfe1702f3a9d.png)

MariusWirtz:
The `execute_mdx_dataframe_shaped` function doesn't support the PROPERTIES syntax. Currently, it is simply ignored.
IMO this is something we should add to the function. I understand that in Arc the first property is displayed as the "Column Header". Is that how you would like it to work in the TM1py function as well?

For the time being, if you want to see attribute values on the column headers, you can do a small additional query like this:

``` python
from TM1py import TM1Service

with TM1Service(address="", port=12354, user="admin", password="apple", ssl=True) as tm1:
    mdx = """
    SELECT
    {Tm1SubsetAll([d1])} ON ROWS,
    {[d2].[e1], [d2].[e2], [d2].[e3]} ON COLUMNS
    FROM [c1]
    """

    df = tm1.cells.execute_mdx_dataframe_shaped(mdx)

    mdx = """
    SELECT
    {[}ElementAttributes_d2].[Alias]} ON ROWS,
    {[d2].[e1], [d2].[e2], [d2].[e3]} ON COLUMNS
    FROM [}ElementAttributes_d2]
    """
    values = list(tm1.cells.execute_mdx_values(mdx))

    df.columns = ["d1"] + values

[...]

PR #479, opened 2021-02-03
Support MDX PROPERTIES in execute_mdx_dataframe

MariusWirtz:
use `display_attribute=True` and user PROPERTIES in MDX query

Related to #470

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -992,7 +992,7 @@ class CellService(ObjectService):
     @require_pandas
     def execute_mdx_dataframe(self, mdx: str, top: int = None, skip: int = None, skip_zeros: bool = True,
                               skip_consolidated_cells: bool = False, skip_rule_derived_cells: bool = False,
-                              sandbox_name: str = None, display_attribute: bool = False,
+                              sandbox_name: str = None, include_attributes: bool = False,
                               **kwargs) -> 'pd.DataFrame':
         """ Optimized for performance. Get Pandas DataFrame from MDX Query.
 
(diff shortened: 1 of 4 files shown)


MariusWirtz:
Prototype for MDX PORPERTIES support in `execute_mdx_dataframe` functions:


``` python
from TM1py import TM1Service

with TM1Service(address="", port=12354, user="admin", password="apple", ssl=True) as tm1:
    mdx = """
    SELECT
    {Tm1SubsetAll([d1])} PROPERTIES [d1].[Attribute Something] ON 0,
    {Tm1SubsetAll([d2])} PROPERTIES [d2].[Number] ON 1
    FROM [c1]
    """

    data = tm1.cells.execute_mdx_dataframe(mdx, display_attribute=True)

    print(data)
```

rclapp:
@MariusWirtz this works, but behaves differently than the old content endpoint. When properties appear on the rows, the display_attributes replaces the element which is problematic when a query includes multiple attributes. I think the the desired behavior is to add a new column for each row property, then users could drop the original member name if desired.

For this query:
`SELECT 
      NON EMPTY 
       {[Scenario].[Scenario].[Planning]} 
      ON COLUMNS , 
      NON EMPTY 
       {[Product].[Product].[C1],[Product].[Product].[C3],[Product].[Product].[C4]} 
       DIMENSION PROPERTIES [Product].[Product].[Usage], [Product].[Product].[Reporting Name]
      ON ROWS 
    FROM [Test] `

The structure would look something like this
image

MariusWirtz:
Thanks @rclapp. I will adjust it accordingly.

MariusWirtz:
@rclapp I changed it accordingly. The only problem I see now is that if you don't provide `PROPERTIES` for any dimension in the MDX, it will retrieve all attributes for that dimension by default.

Do you know a "dummy" `PROPERTY` of a dimension, that users could specify in the MDX in order to avoid all attributes being loaded?

rclapp:
Yes, you could include "DIMENSION PROPERTIES MEMBER_NAME"

It works on all axis. Ideally include attributes would be false by default. When set to true the user should have a well defined query to prevent getting too many attributes.

Sent from my mobile phone

On Feb 4, 2021 8:19 AM, Marius Wirtz <notifications@github.com> wrote:

@rclapp<https://github.com/rclapp> I changed it accordingly. The only problem I see now is that if you don't provide PROPERTIES for any dimension in the MDX, it will retrieve all attributes for that dimension by default.

Do you know a "dummy" PROPERTY of a dimension, that users could specify in the MDX in order to avoid all attributes being loaded?

-

MariusWirtz:
Nice one. `PROPERTIES MEMBER_NAME`  does the trick.  Yes, `include_attributes` is `False` by default. 
As you say, it's up to the user now to use this feature with care.

For the `execute_mdx_dataframe_shaped`, I suggest we use a different approach and a different argument name: `display_attribute`. If it's passed as `True` we would show the (first) attribute value instead of the element name. That's also how it works in Arc and what @yyzz1010 asked for in #470.

rclapp:
Agree 100%, Shaped frames don’t need any more complexity.

PR #493, opened 2021-02-16
Introduce new cancel_at_timeout argument

### LINKED ISSUE #489: Cancel MDX query when timeout argument is specified in execute_mdx type of functions
I would like the timeout argument for execute_mdx to also cancel the thread on tm1 side.

Currently timeout argument only cancels execute_mdx function and moves on in the python script, but the MDX query is still running in the background

**Version**
- TM1py 1.6.0
- TM1 Server Version: 11.0.8

**Additional context**
If you encounter an error, please add the error message and the stack trace

MariusWirtz:
`cancel_at_timeout` can be passed to any function in TM1py.
Controls if operation in TM1 is aborted when timeout is reached.

Fixes #489

### DIFF
--- a/TM1py/Services/MonitoringService.py
+++ b/TM1py/Services/MonitoringService.py
@@ -83,6 +83,14 @@ class MonitoringService(ObjectService):
         response = self._rest.POST(url, **kwargs)
         return response
 
+    def get_active_session_threads(self, exclude_idle: bool = True, **kwargs):
+        url = "/api/v1/ActiveSession/Threads?$filter=Function ne 'GET /api/v1/ActiveSession/Threads'"
+        if exclude_idle:
+            url += " and State ne 'Idle'"
+
+        response = self._rest.GET(url, **kwargs)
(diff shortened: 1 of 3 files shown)


MariusWirtz:
``` python
from TM1py import TM1Service
from TM1py.Exceptions import TM1pyTimeout

with TM1Service(address="", port=12354, ssl=True, user="admin", password="apple", async_requests_mode=True) as tm1:
    try:
        data = tm1.cells.execute_view("Big Cube", "Default", timeout=1, cancel_at_timeout=True)

    except TM1pyTimeout:
        pass

```

``` python
from TM1py import TM1Service
from TM1py.Exceptions import TM1pyTimeout

with TM1Service(address="", port=12354, ssl=True, user="admin", password="apple", async_requests_mode=False) as tm1:
    try:
        data = tm1.cells.execute_view("Big Cube", "Default", timeout=1, cancel_at_timeout=True)

    except TM1pyTimeout:
        pass

```

PR #506, opened 2021-03-10
Make install with pandas work with zsh

beckyconning:


MariusWirtz:
Hi @beckyconning,
thanks for the MR. Why do we need the double quotes around `tm1py[pandas]`?

MariusWirtz:
Understood. Apparently "zsh uses square brackets for globbing / pattern matching."
Thanks!

PR #521, opened 2021-04-06
Added GitService to TM1Service and __init__ files

harveyca307:
I noticed that the Git service was not available on the context menu.  I added the service to TM1Service.py and __init__.py under Services and TM1Py


MariusWirtz:
Thanks @harveyca307

PR #543, opened 2021-05-28
Add write_dataframe_async function and tests

macsir:
Hi, @MariusWirtz 
Finally, got some time to tidy this up. I have done some testing. Please review and test it again.
Happy to discuss further. Have a nice weekend! :)

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -386,9 +386,16 @@ class CellService(ObjectService):
         :param increment: increment or update cell values. Defaults to False.
         :return: the Future’s result or raise exception.
         """
+        if not isinstance(data, pd.DataFrame):
+            raise ValueError("argument 'data' must of type DataFrame")
+
+        dimensions = self.get_dimension_names_for_writing(cube_name=cube_name)
+
+        if not len(data.columns) == len(dimensions) + 1:
+            raise ValueError("Number of columns in 'data' DataFrame must be number of dimensions in cube + 1")
+
         def chunks(data):
-            slice_list = [data.iloc[i:i + slice_size_of_dataframe] for i in range(0, data.shape[0], slice_size_of_dataframe)]
-            return slice_list
+            return [data.iloc[i:i + slice_size_of_dataframe] for i in range(0, data.shape[0], slice_size_of_dataframe)]
 
         def write(self, chunk):
             return self.write_dataframe(cube_name=cube_name, data=chunk, increment=increment, use_ti=True)
@@ -404,15 +411,7 @@ class CellService(ObjectService):
                     outcomes.append(await future)
 
             return outcomes
-
-        if not isinstance(data, pd.DataFrame):
-            raise ValueError("argument 'data' must of type DataFrame")
-
-        dimensions = self.get_dimension_names_for_writing(cube_name=cube_name)
-
-        if not len(data.columns) == len(dimensions) + 1:
-            raise ValueError("Number of columns in 'data' DataFrame must be number of dimensions in cube + 1")
-
+            
         return asyncio.run(write_async(self, data))
 
     def write_value(self, value: Union[str, float], cube_name: str, element_tuple: Iterable,
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -391,10 +391,7 @@ class CellService(ObjectService):
             return slice_list
 
         def write(self, chunk):
-            try:
-                return self.write_dataframe(cube_name=cube_name, data=chunk, increment=increment, use_ti=True)
-            except TM1pyException as e:
-                raise e
+            return self.write_dataframe(cube_name=cube_name, data=chunk, increment=increment, use_ti=True)
 
         async def write_async(self, data):
             loop = asyncio.get_event_loop()
@@ -416,10 +413,7 @@ class CellService(ObjectService):
         if not len(data.columns) == len(dimensions) + 1:
             raise ValueError("Number of columns in 'data' DataFrame must be number of dimensions in cube + 1")
 
-        try:
-            return asyncio.run(write_async(self, data))
-        except TM1pyException as e:
-            raise e
+        return asyncio.run(write_async(self, data))
 
     def write_value(self, value: Union[str, float], cube_name: str, element_tuple: Iterable,
                     dimensions: Iterable[str] = None, sandbox_name: str = None, **kwargs) -> Response:
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -4,9 +4,11 @@ import functools
 import json
 import uuid
 import warnings
+import asyncio
 from collections import OrderedDict
 from io import StringIO
 from typing import List, Union, Dict, Iterable, Tuple, Optional
+from concurrent.futures.thread import ThreadPoolExecutor
 
 from mdxpy import MdxHierarchySet, MdxBuilder, Member
 from requests import Response
@@ -367,6 +369,57 @@ class CellService(ObjectService):
                           use_ti=use_ti,
                           use_changeset=use_changeset,
                           **kwargs)
+
+    @require_pandas
+    @manage_transaction_log
+    def write_dataframe_async(self, cube_name: str, data: 'pd.DataFrame',
+                              slice_size_of_dataframe: int, max_workers: int,
+                              increment: bool = False):
+        """ Write DataFrame into a cube using unbound TI processes in a multi-threading way. Requires admin permissions.
+        For a DataFrame with > 1,000,000 rows, this function will at least save half of runtime compared with `write_dataframe` function.
+        Column order must match dimensions in the target cube with an additional column for the values.
+        Column names are not relevant.
+        :param cube_name:
+        :param data: Pandas Data Frame
+        :param slice_size_of_dataframe: Number of rows for each DataFrame slice, e.g. 10000
+        :param max_workers: Max number of threads, e.g. 14
+        :param increment: increment or update cell values. Defaults to False.
+        :return: the Future’s result or raise exception.
+        """
+        def chunks(data):
+            slice_list = [data.iloc[i:i + slice_size_of_dataframe] for i in range(0, data.shape[0], slice_size_of_dataframe)]
+            return slice_list
+
+        def write(self, chunk):
+            try:
+                return self.write_dataframe(cube_name=cube_name, data=chunk, increment=increment, use_ti=True)
+            except TM1pyException as e:
+                raise e
+
+        async def write_async(self, data):
+            loop = asyncio.get_event_loop()
+            outcomes = []
+
+            with ThreadPoolExecutor(max_workers) as executor:
+                futures = [loop.run_in_executor(executor, write, self, chunk) for chunk in chunks(data)]
+
+                for future in futures:
+                    outcomes.append(await future)
+
+            return outcomes
+
+        if not isinstance(data, pd.DataFrame):
+            raise ValueError("argument 'data' must of type DataFrame")
+
+        dimensions = self.get_dimension_names_for_writing(cube_name=cube_name)
+
+        if not len(data.columns) == len(dimensions) + 1:
+            raise ValueError("Number of columns in 'data' DataFrame must be number of dimensions in cube + 1")
+
+        try:
+            return asyncio.run(write_async(self, data))
+        except TM1pyException as e:
+            raise e
 
     def write_value(self, value: Union[str, float], cube_name: str, element_tuple: Iterable,
                     dimensions: Iterable[str] = None, sandbox_name: str = None, **kwargs) -> Response:
--- a/Tests/CellService.py
+++ b/Tests/CellService.py
@@ -645,6 +645,26 @@ class TestCellService(unittest.TestCase):
         self.assertEqual(list(df["Value"]), values)
 
     @skip_if_no_pandas
+    def test_write_dataframe_async(self):
+        df = pd.DataFrame({
+            self.dimension_names[0]: ["element 1", "element 1", "element 1"],
+            self.dimension_names[1]: ["element 1", "element 2", "element 3"],
+            self.dimension_names[2]: ["element 5", "element 5", "element 5"],
+            "Value": [1, 2, 3]})
+        self.tm1.cubes.cells.write_dataframe_async(self.cube_name, df, 1, 3)
+
+        mdx = MdxBuilder.from_cube(self.cube_name) \
+            .add_hierarchy_set_to_column_axis(MdxHierarchySet.member(Member.of(self.dimension_names[0], "element 1"))) \
+            .add_hierarchy_set_to_row_axis(MdxHierarchySet.members([
+            Member.of(self.dimension_names[1], "element 1"),
+            Member.of(self.dimension_names[1], "element 2"),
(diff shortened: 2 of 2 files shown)


MariusWirtz:
Hi @macsir,

thanks for the PR. Very much appreciated. 

The code looks great. It's nice that you added tests too! 
I will add one small comment on something that isn't clear to me.

TM1py/Services/CellService.py
```diff
@@ -368,6 +370,57 @@ def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iter
...
+        """
+        def chunks(data):
+            slice_list = [data.iloc[i:i + slice_size_of_dataframe] for i in range(0, data.shape[0], slice_size_of_dataframe)]
+            return slice_list
+
+        def write(self, chunk):
+            try:
+                return self.write_dataframe(cube_name=cube_name, data=chunk, increment=increment, use_ti=True)
+            except TM1pyException as e:
+                raise e
```
MariusWirtz, line 397: I think lines 394, 396, and 397 are obsolete. 
I mean if you didn't add lines 394, 396, and 397 wouldn't the code do the same thing?

TM1py/Services/CellService.py
```diff
@@ -368,6 +370,57 @@ def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iter
...
+
+        dimensions = self.get_dimension_names_for_writing(cube_name=cube_name)
+
+        if not len(data.columns) == len(dimensions) + 1:
+            raise ValueError("Number of columns in 'data' DataFrame must be number of dimensions in cube + 1")
+
+        try:
+            return asyncio.run(write_async(self, data))
+        except TM1pyException as e:
+            raise e
```
MariusWirtz, line 422: Same as in the comment above. I think lines 419, 421, and 422 are obsolete.
Am I missing something?

TM1py/Services/CellService.py
macsir: Hi, @MariusWirtz 
Yes, agreed and I have submitted another commit to clean them.

TM1py/Services/CellService.py
macsir: Same as above.

TM1py/Services/CellService.py
```diff
@@ -368,6 +370,57 @@ def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iter
                           use_changeset=use_changeset,
                           **kwargs)
 
+    @require_pandas
+    @manage_transaction_log
+    def write_dataframe_async(self, cube_name: str, data: 'pd.DataFrame',
+                              slice_size_of_dataframe: int, max_workers: int,
```
rkvinoth, line 376: `max_workers` can be made a default argument.

TM1py/Services/CellService.py
```diff
@@ -368,6 +370,57 @@ def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iter
...
+                              increment: bool = False):
+        """ Write DataFrame into a cube using unbound TI processes in a multi-threading way. Requires admin permissions.
+        For a DataFrame with > 1,000,000 rows, this function will at least save half of runtime compared with `write_dataframe` function.
+        Column order must match dimensions in the target cube with an additional column for the values.
+        Column names are not relevant.
+        :param cube_name:
+        :param data: Pandas Data Frame
+        :param slice_size_of_dataframe: Number of rows for each DataFrame slice, e.g. 10000
+        :param max_workers: Max number of threads, e.g. 14
```
rkvinoth, line 385: We could add a note on the limit for max threads.

TM1py/Services/CellService.py
rkvinoth, line 376: `slice_size_of_dataframe` could also be converted into a default argument and if nothing is passed, the method should intelligently split the DataFrame into multiple chunks especially when the number of records are less than 16380.

TM1py/Services/CellService.py
```diff
@@ -368,6 +370,57 @@ def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iter
...
+    def write_dataframe_async(self, cube_name: str, data: 'pd.DataFrame',
+                              slice_size_of_dataframe: int, max_workers: int,
+                              increment: bool = False):
+        """ Write DataFrame into a cube using unbound TI processes in a multi-threading way. Requires admin permissions.
+        For a DataFrame with > 1,000,000 rows, this function will at least save half of runtime compared with `write_dataframe` function.
+        Column order must match dimensions in the target cube with an additional column for the values.
+        Column names are not relevant.
+        :param cube_name:
+        :param data: Pandas Data Frame
+        :param slice_size_of_dataframe: Number of rows for each DataFrame slice, e.g. 10000
```
rkvinoth, line 384: We could add a note here saying the `slice_size_of_dataframe` should be less than 16381 for the highest performance.

TM1py/Services/CellService.py
```diff
@@ -368,6 +370,57 @@ def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iter
...
+        :param max_workers: Max number of threads, e.g. 14
+        :param increment: increment or update cell values. Defaults to False.
+        :return: the Future’s result or raise exception.
+        """
+        def chunks(data):
+            slice_list = [data.iloc[i:i + slice_size_of_dataframe] for i in range(0, data.shape[0], slice_size_of_dataframe)]
```
rkvinoth, line 390: We could remove this variable declaration to save a line of code.

TM1py/Services/CellService.py
```diff
@@ -368,6 +370,57 @@ def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iter
...
+
+            with ThreadPoolExecutor(max_workers) as executor:
+                futures = [loop.run_in_executor(executor, write, self, chunk) for chunk in chunks(data)]
+
+                for future in futures:
+                    outcomes.append(await future)
+
+            return outcomes
+
+        if not isinstance(data, pd.DataFrame):
```
rkvinoth, line 408: This could be moved to the top this method.

TM1py/Services/CellService.py
```diff
@@ -368,6 +370,57 @@ def write_dataframe(self, cube_name: str, data: 'pd.DataFrame', dimensions: Iter
...
+            raise ValueError("argument 'data' must of type DataFrame")
+
+        dimensions = self.get_dimension_names_for_writing(cube_name=cube_name)
+
+        if not len(data.columns) == len(dimensions) + 1:
```
rkvinoth, line 413: This could also be moved to the top.

rkvinoth (commented):
Suggesting few minor changes.

TM1py/Services/CellService.py
MariusWirtz: I agree it would be nice to have default values, or `None` as default values and intelligence inside the function to determine sensible values considering the size of the passed data frame.

The value for `slice_size_of_dataframe` should be 32k or a multiple of 32k IMO. Since we squeeze 16k in the prolog and 16k in the epilog in the `write_through_unbound_process` function. So effectively TM1 consumes chunks of 32k.

macsir: Hi, @rkvinoth 
Thanks. The reason I don't want to put default values for the slice_size_of_dataframe and max_workers is that they are really dependent on the actual running environment. e.g. memory, cpu, network etc. After testing, I reckon it would be good to let users decide to get the most optimum combination. Obviously, if putting the too many workers, it would cause high cpu usage and freeze the system possibly.

TM1py/Services/CellService.py
macsir: Same as above.

TM1py/Services/CellService.py
macsir: Same as above.

TM1py/Services/CellService.py
macsir: Yes, agreed and will change it.

TM1py/Services/CellService.py
MariusWirtz: I think I disagree. I think it would be most efficient to break the data frame into multiple data frames of each 32k records or into n data frames with each a multiple of 32k records.

Like if a data frame of 320k records is passed and max_workers is 5, I think there are 2 sensible approaches:

A: break data frame into 10 dataframes of 32k records each -> the `write_dataframe` function is called 10 times. 10 unbound processes will be executed with each 32k statements.

B: break data frame into 5 dataframes of 64k records each -> the `write_dataframe` function is called 5 times. 10 unbound processes will be executed with each 32k statements.

TM1py/Services/CellService.py
macsir: Yes, agreed.

TM1py/Services/CellService.py
macsir: Yes, agreed.

TM1py/Services/CellService.py
MariusWirtz: Fair enough. `max_workers` it totally depending on the environment.

However, for the `slice_size` I do believe we should provide a sensible default value since a casual TM1py user may not be aware of the fact that TM1 consumes records in chunks of 32k (since this is the max number of CellPut statements we can squeeze into an unbound process)

rkvinoth: `max_workers` by default takes the best possible number of threads. If users need optimization (or reach the threshold) they can reduce it as required.
`ThreadPoolExecutor` has the `max_workers` as a default argument and I think we should respect that.

with <3.8 the default count is cpu_count * 5
with >3.8 the default count is cpu_count + 4

TM1py/Services/CellService.py
rkvinoth: There should be another approach as well when `max_workers` is not provided:
A DataFrame with 320K records can be divided as follows:
1. 10 threads - 32K each
2. 20 threads - 16K each (This will be the fastest way for bigger orgs with tanks and bazooka)

Spawning threads will take extra time and I believe TM1py should handle these stuff by default as we can't expect everybody to be a tech geek. 
People with skills can still optimize it further by playing with the `max_workers`.

IMHO **threadpool** is very less effective compared to **processpool** as what we are doing here is a CPU expensive operation.

If we can also implement another method that can write using **ProcessPoolExecutor** that would be awesome.

TM1py/Services/CellService.py
MariusWirtz: It is not guaranteed that Python runs on the same machine as TM1. 
TM1py can not know for sure how many CPUs are available to TM1.

rkvinoth: > It is not guaranteed that Python runs on the same machine as TM1.

If python is not running on the same machine as TM1, we shouldn't even worry about the number of threads.

> TM1py can not know for sure how many CPUs are available to TM1.

TM1py is just triggering the ThreadPoolExecutor.... ThreadPoolExecutor in turn uses `os` module to handle the thread count when it is not provided.
If you think ThreadPoolExecutor could take up more threads and not provide enough for TM1, I would argue that users could do the same. They can just go berserk and it is their responsibility to keep things in check.

MariusWirtz: > If python is not running on the same machine as TM1, we shouldn't even worry about the number of threads.

Thanks for your thoughts and feedback @rkvinoth. I think I disagree. As TM1 admins or developers, I think we do care about not overloading the TM1 server with work.
I mean when TM1py sends those CellPut statements to the TM1 Server, the heavy lifting is done by TM1. Python is actually just waiting for a response from the server most of the time. TM1 is digesting the values and updating the cube structures. 
On the client-side python can use 4 or 10 threads even if you have just one free CPU. That's not problematic because most of the (python) threads will be waiting most of the time anyways for TM1 to finish digesting the API requests.
 
On the server-side, I would expect to see (in Arc, Pulse or tm1top) 4 or 10 threads processing almost at every moment.
So on the server side, we are causing TM1 to be busy. If the server has let's say 16 CPUs, I would argue we shouldn't use all of them for writing because then if TM1 users open views at this moment there are limited CPU resources to cater for their session and they may experience delays.

rkvinoth: I agree with what you have in mind on not burdening TM1 server. My point is the same, TM1 developer/admin should know what he is doing.
If you think developers will not provide the thread count if we leave it as default then let's not keep it default. I wanted to respect what `ThreadPoolExecutor` offers.

rkvinoth: As you know, using all the available threads will not provide better results. Developers still have to play with the thread count to arrive at the optimal number.

MariusWirtz: Yeah maybe let's not provide a default for `max_workers`. It feels wrong to me to default to something that python suggests for the environment in which python runs. 
I think it's reasonable to ask the TM1py user to provide a value for `max_workers`.

> As you know, using all the available threads will not provide better results. Developers still have to play with the thread count to arrive at the optimal number.

Yes!

TM1py/Services/CellService.py
MariusWirtz: Interesting thoughts! It would be interesting to quantify how the `ProcessPoolExecutor` performs compared to `ThreadPoolExecutor`. 

Perhaps we can enhance the function (now or in a separate PR) to support both executors and let the user choose through an optional function argument?

TM1py/Services/CellService.py
macsir: Totally agree. I have same feeling when I was writing a huge dataframe on Cloud down to a TM1 server. I can simply put a very high number for the function to make the TM1 server "dead" as the CPU was exhausted.

So, do you guy want me to leave the max_workers without any default value as it is?

> Yeah maybe let's not provide a default for max_workers. It feels wrong to me to default to something that python suggests for the environment in which python runs.
I think it's reasonable to ask the TM1py user to provide a value for max_workers.

> It is not guaranteed that Python runs on the same machine as TM1.
TM1py can not know for sure how many CPUs are available to TM1.

macsir (commented):
Thanks for your advice. 
My 2 cents. When I was testing on the Cloud, the 32k slice_size_of_dataframe doesn't give the best performance as the transferring speed of chunks are again limited by the network environment between the Cloud and the intranet. You won't see max_works run at the same time at the begining. If you run everything locally or even on the same server, this obviously can be increased a lot. That's why I leave it as it is so dynamic as well.

ProcessPoolExecutor came to my mind too while testing. However, my environments are more IO-bound as there is a network speed limit. And cellput,  in my opinion, in essence doesn't involve lots of CPU calculation and won't be benefit from that. Therefore, I didn't go down to that path. Happy to discuss further.

TM1py/Services/CellService.py
MariusWirtz: OK. So we don't put a default value for `max_workers`.

MariusWirtz:
Thanks @macsir and @rkvinoth for the excellent contributions and discussions. I will merge this as is now.

I will try to do some performance testing on my site. I would expect to see a positive impact on performance if a multiple of 32k is used, compared to an arbitrary number like 10k or 50k. I will try to quantify this in the next couple of days and potentially commit a change to the function, without breaking compatibility.

@rkvinoth feel free to make Executor dynamic in a separate PR. I think it would be a nice enhancement to the function if the user could choose if threads or processes shall be used for the parallelization.

PR #545, opened 2021-05-31
Utilise MDX to return list of elements

jrobinsonAG:
Replacing REST GET with execute_set_mdx to return list of element names filtered by attribute value negating use cases where attribute names contain special characters.


jrobinsonAG:
Closing as incorrectly merged subsequent commits

PR #546, opened 2021-06-01
Utilise MDX for attribute filter

jrobinsonAG:
Replace REST call with set MDX to mitigate special characters in attribute names.

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -350,7 +350,11 @@ class ElementService(ObjectService):
         :param attribute_value:
         :return: List of element names
         """
-
+        if not self.exists(f'}}ElementAttributes_{dimension_name}',
+                f'}}ElementAttributes_{dimension_name}',
+                attribute_name):
+            raise RuntimeError(f"Attribute '{attribute_name}' does not exist in Dimension '{dimension_name}'.")
+            
         if isinstance(attribute_value, str):
             mdx = (
                 f'{{FILTER({{TM1SUBSETALL([{dimension_name}].[{hierarchy_name}])}},'
@@ -362,7 +366,11 @@ class ElementService(ObjectService):
                 f'[{dimension_name}].[{hierarchy_name}].[{attribute_name}] = {attribute_value})}}'
             )
 
-        elems =  self.execute_set_mdx(mdx)
+        elems =  self.execute_set_mdx(
+            mdx=mdx,
+            member_properties=["Name"],
+            parent_properties=None,
+            element_properties=None)
         return [elem[0]['Name'] for elem in elems]
 
     def create_element_attribute(self, dimension_name: str, hierarchy_name: str, element_attribute: ElementAttribute,
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -350,19 +350,20 @@ class ElementService(ObjectService):
         :param attribute_value:
         :return: List of element names
         """
-        attribute_name = attribute_name.replace(" ", "")
+
         if isinstance(attribute_value, str):
-            url = format_url("/api/v1/Dimensions('{}')/Hierarchies('{}')?$expand=Elements($filter = "
-                             "Attributes/{} eq '{}';$select=Name)",
-                             dimension_name, hierarchy_name, attribute_name, attribute_value)
-
+            mdx = (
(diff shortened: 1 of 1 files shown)


rkvinoth:
@jrobinsonAG 
Thank you for the PR, its nice! The only problem I see here is that the method fails when the attribute name is incorrect and the error message is not very helpful for debugging.

> TM1py.Exceptions.Exceptions.TM1pyRestException: Text: '{"error":{"code":"248","message":"Could not execute MDX expression. Error Code:rte51. Error:Get cell value: Cannot resolve cube context"}}' - Status Code: 400 - Reason: 'Bad Request' - Headers: {'Content-Length': '133', 'Connection': 'keep-alive', 'Content-Encoding': 'gzip', 'Cache-Control': 'no-cache', 'Content-Type': 'application/json; charset=utf-8', 'OData-Version': '4.0'}

Do you think we could add a check for the attribute and raise a custom error?

TM1py/Services/ElementService.py
```diff
@@ -350,19 +350,20 @@ def get_elements_filtered_by_attribute(self, dimension_name: str, hierarchy_name
...
-                             "Attributes/{} eq {};$select=Name)",
-                             dimension_name, hierarchy_name, attribute_name, attribute_value)
+            mdx = (
+                f'{{FILTER({{TM1SUBSETALL([{dimension_name}].[{hierarchy_name}])}},'
+                f'[{dimension_name}].[{hierarchy_name}].[{attribute_name}] = {attribute_value})}}'
+            )
 
-        response = self._rest.GET(url, **kwargs)
-        return [elem['Name'] for elem in response.json()['Elements']]
+        elems =  self.execute_set_mdx(mdx)
```
MariusWirtz, line 365: Since we are returning only a list of element names, I think it would make sense to explicitly only retrieve the element names in the `execute_set_mdx` function. That should speed up the execution. Kinda like this:

``` python
from TM1py import TM1Service

with TM1Service(address='localhost', port=12354, ssl=True, user="admin", password="apple") as tm1:
    mdx = "{Tm1SubsetAll([d1])}"
    elements = tm1.elements.execute_set_mdx(
        mdx=mdx,
        member_properties=["Name"],
        parent_properties=None,
        element_properties=None)

    print(elements)
```

jrobinsonAG: @MariusWirtz  updated as suggested.

jrobinsonAG:
@rkvinoth given there is currently no other custom TM1py exception (not imported) I've just added a custom runtime exception to catch the unknown attribute as suggested. 

How does this look?

rkvinoth:
Just what I had in my mind. But why not raise a TM1pyException?

jrobinsonAG:
@rkvinoth no reason why the tm1py exception module couldn't be imported and utilised. 

Given there was no existing utilisation in ElementService (also checked handful of other objectservices) to date, I was reluctant to add without further input as to the preferred package exception handling conventions. Also did a quick review of the Exceptions module and there are no defined Element or MDX related customs exceptions.

MariusWirtz:
Hi,

Currently we kinda have three conventions that we mostly stick to in TM1py:

- If a TM1 operation fails (e.g. invalid MDX, attempting to write a string to a numeric cell) a `TM1pyException` is raised

- If an illegal argument is passed to a TM1py function we raise a `ValueError` following python's good practices 

https://github.com/cubewise-code/tm1py/blob/f751edff429e264b2d8954680cfd4499dbacdbe7/TM1py/Objects/Dimension.py#L99

https://github.com/cubewise-code/tm1py/blob/f751edff429e264b2d8954680cfd4499dbacdbe7/TM1py/Services/ChoreService.py#L40

- If for whatever reason, we reach an illegal state in the code, we raise a `RuntimeError`.

https://github.com/cubewise-code/tm1py/blob/f751edff429e264b2d8954680cfd4499dbacdbe7/TM1py/Services/DimensionService.py#L36

With that background, the `RuntimeError` or `ValueError` both seem appropriate to me.

Perhaps it would be nice in the long run to have more dedicated TM1pyExceptions like a TM1pyElementDoesNotExist Exception or something?

jrobinsonAG:
Thanks for clarifying @MariusWirtz. Agree longer term a custom exception would likely be more appropriate, especially given the common issue of missing elements in operations against objects; I've left the runtime error for now.

I've made another commit to clean up the string formatting after reviewing some more repo examples.

PR #549, opened 2021-06-10
Create new server on adminhost

jrobinsonAG:
Create a new TM1 instance on the AdminHost via POST method. Requires a valid AdminHost string, and the server configuration. Return instance of TM1py.Server.

Example use:
```python
from TM1py import TM1Service, Utils


with TM1Service(address="", port=12354, ssl=True, user="admin", password="apple") as tm1:
    admin_server = tm1.server.get_admin_host()
    server_as_dict = {
        "Name":"MyModel1",
        "IPAddress":"172.20.10.10",
        "IPv6Address":None,
        "PortNumber":12345,
        "UsingSSL": True,
        "ClientMessagePortNumber":61098,
        "HTTPPortNumber":12999,
        "ClientExportSSLSvrCert":True,
        "ClientExportSSLSvrKeyID":"whateverExportSSLSvrKeyID",
        "AcceptingClients":True
    }

    new_server = Utils.create_server_on_adminhost(
                    adminhost=admin_server,
                    server_as_dict=server_as_dict
                    )
    servers = Utils.get_all_servers_from_adminhost(adminhost=admin_server)
```

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -8,7 +8,7 @@ from contextlib import suppress
 from enum import Enum, unique
 from typing import Any, Dict, List, Tuple, Iterable, Optional, Generator
 
-from TM1py.Exceptions.Exceptions import TM1pyVersionException, TM1pyNotAdminException, TM1pyRestException
+from TM1py.Exceptions.Exceptions import TM1pyVersionException, TM1pyNotAdminException
 
 try:
     import pandas as pd
@@ -100,7 +100,7 @@ def create_server_on_adminhost(adminhost='localhost', server_as_dict={}):
     response_as_dict = json.loads(response)
 
     return Server(response_as_dict)
-    
+
 def build_url_friendly_object_name(object_name: str) -> str:
     return object_name.replace("'", "''").replace('%', '%25').replace('#', '%23')
 
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -74,7 +74,7 @@ def get_all_servers_from_adminhost(adminhost='localhost') -> List:
         servers.append(server)
     return servers
 
-def create_server_on_adminhost(adminhost='localhost', server_as_dict={}):
+def create_server_on_adminhost(adminhost='localhost', server_as_dict=None):
     from TM1py.Objects import Server
     """  Create new TM1 instance on Adminhost
     :param adminhost: IP or DNS Alias of the adminhost
@@ -92,7 +92,8 @@ def create_server_on_adminhost(adminhost='localhost', server_as_dict={}):
                 "AcceptingClients":True }
     :return: instance of TM1py.Server
     """
-
+    if server_as_Dict is None: 
+        server_as_dict = {}
     conn = http_client.HTTPConnection(adminhost, 5895)
     request = '/api/v1/Servers'
     conn.request('POST', request, body=json.dumps(server_as_dict), headers={'Content-Type':'application/json'})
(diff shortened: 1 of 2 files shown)


wimgielis:
Nice :-)
I see that you don't specify config directory; data directory and logging
directory. What does TM1 take then ?


------

Best regards / Beste groeten,

Wim Gielis
MS Excel MVP 2011-2014
https://www.wimgielis.com <http://www.wimgielis.be>


Op vr 11 jun. 2021 om 00:20 schreef Josh.R ***@***.***>:

jrobinsonAG:
Hi @wimgielis 

The example code is replicating the IBM example notes. The server_as_dict parameter should handle and valid endpoint content , acting as a passthrough.

scrumthing:
I thought that this only works on cloud environments. I second @wimgielis questions btw.

TM1py/Utils/Utils.py
```diff
@@ -74,6 +74,32 @@ def get_all_servers_from_adminhost(adminhost='localhost') -> List:
         servers.append(server)
     return servers
 
+def create_server_on_adminhost(adminhost='localhost', server_as_dict={}):
```
MariusWirtz, line 77: default values in functions should be immutable. Like you can set it to `None` instead. 

https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

TM1py/Utils/Utils.py
```diff
@@ -74,6 +74,32 @@ def get_all_servers_from_adminhost(adminhost='localhost') -> List:
...
+    :return: instance of TM1py.Server
+    """
+
+    conn = http_client.HTTPConnection(adminhost, 5895)
+    request = '/api/v1/Servers'
+    conn.request('POST', request, body=json.dumps(server_as_dict), headers={'Content-Type':'application/json'})
+    response = conn.getresponse().read().decode('utf-8')
+    response_as_dict = json.loads(response)
+
+    return Server(response_as_dict)
```
MariusWirtz, line 102: I think we should enhance the existing Server class. 
The class should support all fields that are returned in the response dictionary.

TM1py/Utils/Utils.py
jrobinsonAG: Update to immutable.

TM1py/Utils/Utils.py
jrobinsonAG: I've added all returned properties. I haven't been able to verify all possible outcomes, i.e. local v hosted.

jrobinsonAG:
> I thought that this only works on cloud environments. I second @wimgielis questions btw.

First pass of this was applied on local installation. 

I wasn't trying to provide a non-answer to @wimgielis , though I did. I've run into a bit of a blocker (self imposed) given my available environments.

TM1py/Utils/Utils.py
```diff
@@ -74,6 +74,33 @@ def get_all_servers_from_adminhost(adminhost='localhost') -> List:
...
+                "PortNumber":12345,
+                "UsingSSL": True,
+                "ClientMessagePortNumber":61098,
+                "HTTPPortNumber":12999,
+                "ClientExportSSLSvrCert":True,
+                "ClientExportSSLSvrKeyID":"whateverExportSSLSvrKeyID",
+                "AcceptingClients":True }
+    :return: instance of TM1py.Server
+    """
+    if server_as_Dict is None:
```
MariusWirtz, line 95: typo: `server_as_dict` instead of `server_as_Dict`

MariusWirtz:
@jrobinsonAG 
I added a commit on top of yours and created a new Pull Request here #571.
Since one can only update/delete a server that has been manually added to the Admin Server through the REST API I think a delete method is much less concerning and I included it in the new Pull Request.

Please review and approve #571. Then I will merge the changes into the master.

PR #553, opened 2021-06-22
Fix mismatch df header

jrobinsonAG:
Overriding default row header in dataframe returned by MDX to match the dimension name used in the drop duplicate operation.


MariusWirtz:
Nice one! Thanks

Issue #564, opened 2021-07-06
get_active_session_threads not working


macsir:
**Describe the bug**
get_active_session_threads not returning anything even if there is a process running on the server. exclude_idle flag doesn't work either.

**Expected behavior**
It should return the active thread ID or session ID

**Additional context**
I tested the call " url = "/api/v1/ActiveSession/Threads?$filter=Function ne 'GET /api/v1/ActiveSession/Threads'"" and looks like it always return nothing. Is that due to my server version (11.8)? If removing **ActiveSession**, it only returns part of the sessions and threads.

rclapp:
Are you an admin user? Non admin accounts can only see their own thread.

Sent from my mobile phone

On Jul 6, 2021 3:40 PM, tm1sir ***@***.***> wrote:

Describe the bug
get_active_session_threads not returning anything even if there is a process running on the server. exclude_idle flag doesn't work either.

Expected behavior
It should return the active thread ID or session ID

Additional context
I tested the call " url = "/api/v1/ActiveSession/Threads?$filter=Function ne 'GET /api/v1/ActiveSession/Threads'"" and looks like it always return nothing. Is that due to my server version (11.8)? If removing ActiveSession, it only returns part of the sessions and threads.

-

macsir:
Hi, @rclapp 
Yes, I am using the admin account.

MariusWirtz:
Hi @macsir,

the `get_active_session_threads` function only returns threads that are running in the scope of your current active session.
So anything that's executed by other users or by yourself in another session is excluded by design.

It is really only interesting if you are doing stuff asynchronously. E.g. execute a process without waiting for the response.

macsir:
Thanks. Marius. Would it be useful to check if there is any session or thread running on the server? E.g. Just a simple check before running any other process to avoid potential lock contention?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

MariusWirtz:
Agree. We are missing a function to get all threads that are not idle.

currently, you can only use the `get_threads` function and filter in python.

PR #566, opened 2021-07-07
Add get_active_threads

jrobinsonAG:
Add get_active_threads function to filter TM1 threads which are non-idle directly via REST end-point.

### DIFF
--- a/TM1py/Services/MonitoringService.py
+++ b/TM1py/Services/MonitoringService.py
@@ -28,11 +28,11 @@ class MonitoringService(ObjectService):
         return response.json()['value']
 
     def get_active_threads(self, **kwargs):
-        '''Return a list of non-idle threads from the TM1 Server
+        """Return a list of non-idle threads from the TM1 Server
 
             :return:
                 list: TM1 threads as dict
-        '''
+        """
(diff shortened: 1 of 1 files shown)


jrobinsonAG:
@MariusWirtz is this in line with what you were looking for in relation to #564 ?

TM1py/Services/MonitoringService.py
```diff
@@ -27,6 +27,16 @@ def get_threads(self, **kwargs) -> List:
         response = self._rest.GET(url, **kwargs)
         return response.json()['value']
 
+    def get_active_threads(self, **kwargs):
+        '''Return a list of non-idle threads from the TM1 Server
```
MariusWirtz, line 31: Triple double-quotes should be used for docstrings
https://www.python.org/dev/peps/pep-0257/

MariusWirtz:
Thanks @jrobinsonAG 

Code looks good! I could only find a small violation on double quotes and single quotes conventions :)

jrobinsonAG:
@MariusWirtz sorry was obviously not paying attention to my finger, pre-coffee :)

MariusWirtz:
Thanks for the contribution! Very much appreciated

PR #576, opened 2021-07-26
empty strings suppression in extract_cellset_raw

tomasfelcman:
Adding zero suppression for empty strings.
Instead of filtering only 0 and null values, '' have been added.


MariusWirtz:
Nice one! Thanks @tomasfelcman

PR #578, opened 2021-07-27
Improve `write_dataframe_async` function

MariusWirtz:
introduce new `write_async` function

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import functools
+import itertools
 import json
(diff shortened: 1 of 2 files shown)


MariusWirtz:
@macsir 
I did a few small changes to the `write_dataframe_async` function and added a new `write_async` function for dictionaries.
Can you please double-check that my changes don't break your code?

macsir:
Hi, @MariusWirtz  sorry for the late reply. Yes, it looks good to me. Thanks.

MariusWirtz:
Thanks for the confirmation!

Issue #609, opened 2021-09-16
Update Applications


cubewise-gng:
**Describe what did you try to do with TM1py**
Would like to update existing Application document (workbooks). But cannot find any update functions currently to support this. Does IBM support it yet?

**Describe what's not working the way you expect**
There is the create and create_document_from_file function but would fail if the application document already exist.

**Version**
- TM1py 1.7.0

jrobinsonAG:
Hi,

There is no update_or_create method for applications currently from what I can see. 

You could check if the application exists and then remove it, before adding it again?

``` python
with TM1Service(config) as tm1:
    application_path = "Finance"
    application_type = "Document"
    file_path = "C:\\Users\\User\\Reconciliation.xlsx"
    application_name = "Reconciliation.xlsx"
    private = False
    # Check if already exists
    if tm1.applications.exists(path=application_path,
                               application_type=application_type,
                               name=application_name,
                               private=private):
        # Remove Existing application
        tm1.applications.delete(path=application_path,
                                application_type=application_type,
                                name=application_name,
                                private=private)
    # Add Application
    tm1.applications.create_document_from_file(path_to_file=file_path,
                                               application_path=application_path,
                                               application_name=application_name,
                                               private=private)
```

MariusWirtz:
@jrobinsonAG 
perhaps we should offer a `update_or_create_document_from_file` function, based on your code sample.
Could you post it as a Pull Request?

jrobinsonAG:
@MariusWirtz  sorry been a bit busy lately. PR submitted with basic implementation.

PR #632, opened 2021-10-21
add update_or_create for document from file

josrobins:
#609 adding basic implementation of update_or_create_docuument_from_file. 

Usage
```python

application_path = "Finance/Reports"
file_path = "C:\\Users\\user\\Flash.xlsx"
application_name = "Flash.xlsx"
private = False

with TM1Service(address="", port=12354, ssl=True, user="admin", password="apple") as tm1:

    response = tm1.applications.update_or_create_document_from_file(path=application_path,
                                                                    name=application_name,
                                                                    path_to_file=file_path,
                                                                    private=private)
    if response.status_code == 204:
        print("Success")
    else:
        print(f"Completed with Response Code:{response.status_code}")
```

### DIFF
--- a/TM1py/Services/ApplicationService.py
+++ b/TM1py/Services/ApplicationService.py
@@ -190,7 +190,7 @@ class ApplicationService(ObjectService):
                                             path_to_file: str, private: bool = False, **kwargs) -> Response:
         """Update or create application from file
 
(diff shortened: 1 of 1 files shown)


MariusWirtz:
Nice one! Thanks

PR #657, opened 2021-12-17
Add `skip_non_updateable` arg to write function

### LINKED ISSUE #651: Skip rule calculated cells when writing
When the TM1py script writes the data to a cube with tm1.cells.**write** but some cells are rule calculated, the scripts fails with an error.
It would be useful to have an argument to the **write** method to skip the blocked cells (like **SkipNonUpdateable=True**) and do a check similar to **CellIsUpdateable**.
If the argument is False or not provided, the script should act as of now.

**Version**
TM1 Version: 11.4.00003.8

MariusWirtz:
Handy new argument to skip cells that are not updateable.

Fixes #651

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -27,7 +27,8 @@ from TM1py.Utils.Utils import build_pandas_dataframe_from_cellset, dimension_nam
     CaseAndSpaceInsensitiveDict, wrap_in_curly_braces, CaseAndSpaceInsensitiveTuplesDict, \
     abbreviate_mdx, \
     build_csv_from_cellset_dict, require_version, require_pandas, build_cellset_from_pandas_dataframe, \
-    case_and_space_insensitive_equals, get_cube, resembles_mdx, require_admin, extract_compact_json_cellset
+    case_and_space_insensitive_equals, get_cube, resembles_mdx, require_admin, extract_compact_json_cellset, \
+    cell_is_updateable, build_mdx_from_cellset, build_mdx_and_values_from_cellset
 
 try:
     import pandas as pd
@@ -570,7 +571,7 @@ class CellService(ObjectService):
     def write(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str] = None, increment: bool = False,
               deactivate_transaction_log: bool = False, reactivate_transaction_log: bool = False,
               sandbox_name: str = None, use_ti=False, use_changeset: bool = False, precision: int = 8,
-              **kwargs) -> Optional[str]:
+              skip_non_updateable: bool = False, **kwargs) -> Optional[str]:
         """ Write values to a cube
 
         Same signature as `write_values` method, but faster since it uses `write_values_through_cellset`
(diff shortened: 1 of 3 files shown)


MariusWirtz:
Performance impact of new argument quantified below:

> Scenario1: 100k updateable cells + 100k non updateable cells

|use_ti|skip_non_updateable|Elapsed Time|
|------|-------------------|------------|
|TRUE  |TRUE               |00:00:33    |
|FALSE |TRUE               |00:01:25    |
|TRUE  |FALSE              |00:01:25    |
|FALSE |FALSE              |00:01:25    |


> Scenario2: 200k updateable cells

|use_ti|skip_non_updateable|Elapsed Time|
|------|-------------------|------------|
|TRUE  |TRUE               |00:00:36    |
|FALSE |TRUE               |00:01:57    |
|TRUE  |FALSE              |00:00:14    |
|FALSE |FALSE              |00:00:52    |



All tests run on localhost.

MariusWirtz:
@dsfomichev, @gbryant-dev
Any thoughts on this approach?

gbryant-dev:
I like it. I've left a comment just to get your thoughts on one approach over the other 🙂

The performance impact when testing with the first scenario and `skip_non_updateable` is `False` is interesting compared to second scenario when all cells are updateable. Presumably, the write failures are slowing the processing down.

TM1py/Services/CellService.py
```diff
@@ -680,12 +702,24 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
                 else:
                     value_str = format(value, f'.{precision}f')
 
+            comma_separated_elements = ",".join("'" + element.replace("'", "''") + "'" for element in coordinates)
+
+            if skip_non_updateable:
+                cell_is_updateable_pre = f"IF(CellIsUpdateable('{cube_name}', {comma_separated_elements})=1,"
+                cell_is_updateable_post = ",0);"
+            else:
```
rkvinoth, line 710: why not make the statements in else condition default? (saves a line of code).

rkvinoth:
The performance metrics looks very useful and provides an idea of when to use this parameter. 
Thanks @MariusWirtz

TM1py/Services/CellService.py
```diff
@@ -602,19 +604,16 @@ def write(self, cube_name: str, cellset_as_dict: Dict, dimensions: Iterable[str]
...
 
-        values = []
-        query = MdxBuilder.from_cube(cube_name)
-        for coordinates, value in cellset_as_dict.items():
-            members = (Member.of(dimension, element) for dimension, element in zip(dimensions, coordinates))
-            query.add_member_tuple_to_columns(*members)
-            values.append(value)
-        mdx = query.to_mdx()
+        if skip_non_updateable:
+            cellset_as_dict = self.drop_non_updateable_cells(cellset_as_dict, cube_name, dimensions)
```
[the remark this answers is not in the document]
MariusWirtz: Thanks. I agree it should be handled one level deeper. The `write_values_through_cellset` function already takes an MDX and a list of values but the current approach to filter out the non-updateable requires a cellset dictionary structure.

I introduced a new function: `write_through_cellset` that is called from `write` and takes a cellset dictionary instead of an MDX and calls the `write_values_through_cellset`.

TM1py/Services/CellService.py
MariusWirtz: OK

MariusWirtz:
Thanks for the feedback @gbryant-dev and @rkvinoth  

I implemented your suggestions.

PR #658, opened 2021-12-21
Cube and Process Search functions

adscheevel:
**Process Search:** added 2 functions to search for string in process body and to search string in process name
_search_string_in_body_ will look for string anywhere in prolog, metadata, data epilog. UseCase: find TI process(es) that imports a csv file with name "x.csv"
_search_string_in_name_ will search for string anywhere in process name; two parameters: name_startswith and name_contains. name_contains can be an iterable list of strings to search with AND operator. UseCase: find TI process(es) that start with "Load" and have string "GL" anywhere in process name.

**Cube Search:** added 2 functions to find list of cubes that contain specified dimension or wildcard search on dimension name
_get_names_with_dimension_ will return list of cubes that contain specified dimension, optional parameter to return model cubes only. UseCase: find all cubes that use dimension "Account" , a dimension I'm planning a major reconstruction of
_get_names_with_dimension_wildcard_ will return dictionary of cubes that contain a dimension whose name contains a wildcard search criteria. The dimensions returned for a given cube in this function are only the dimensions that match the wildcard search criteria. UseCase: find all cubes that use dim that has the word "time" in its name so I can work on harmonizing them.

### DIFF
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -189,6 +189,42 @@ class CubeService(ObjectService):
             return dimension_names[1:]
         return dimension_names
 
+    def get_names_with_dimension(self, dimension_name: str, model_cubes_only: bool = False,
+                                 **kwargs) -> List[str]:
+        """ Ask TM1 Server for list of cube names that contain specific dimension
+
+        :param dimension_name: string, valid dimension name (case insensitive)
+        :param model_cubes_only: bool, True will filter result to model cubes only
+        """
+
+        cube_prefix = 'Cubes' if model_cubes_only == False else 'ModelCubes()'
+        url = format_url(
+            "/api/v1/{}?$select=Name&$filter=Dimensions/any(d: toupper(d/Name) eq toupper('{}'))",
+            cube_prefix, dimension_name
+            )
+        response = self._rest.GET(url, **kwargs)
+        cubes = list(entry['Name'] for entry in response.json()['value'])
+        return cubes
+
+
+    def get_names_with_dimension_wildcard(self, wildcard: str, model_cubes_only: bool = False,
+                                          **kwargs) -> List[str]:
+        """ Ask TM1 Server for list of cube names that contain dimension whose name contains wildcard
+
+        :param wildcard: string, wildcard to search for in dim name
+        :param model_cubes_only: bool, True will filter result to model cubes only
+        """
+        cube_prefix = 'Cubes' if model_cubes_only == False else 'ModelCubes()'
+        url = format_url(
+            "/api/v1/{}?$select=Name&$filter=Dimensions/any(d: contains(toupper(d/Name),toupper('{}')))" \
+            "&$expand=Dimensions($select=Name;$filter=contains(toupper(Name),toupper('{}')))",
+            cube_prefix, wildcard, wildcard
+            )
+        response = self._rest.GET(url, **kwargs)
+        cube_dict = {entry['Name']:[dim['Name'] for dim in entry['Dimensions']] for entry in response.json()['value']}
+        return cube_dict
+    
+    
     @require_version(version="11.4")
     def get_storage_dimension_order(self, cube_name: str, **kwargs) -> List[str]:
         """ Get the storage dimension order of a cube
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -196,7 +196,6 @@ class CubeService(ObjectService):
         :param dimension_name: string, valid dimension name (case insensitive)
         :param model_cubes_only: bool, True will filter result to model cubes only
         """
-
         cube_prefix = 'Cubes' if model_cubes_only == False else 'ModelCubes()'
         url = format_url(
             "/api/v1/{}?$select=Name&$filter=Dimensions/any(d: toupper(d/Name) eq toupper('{}'))",
@@ -223,8 +222,7 @@ class CubeService(ObjectService):
         response = self._rest.GET(url, **kwargs)
         cube_dict = {entry['Name']:[dim['Name'] for dim in entry['Dimensions']] for entry in response.json()['value']}
         return cube_dict
-    
-    
+        
     @require_version(version="11.4")
     def get_storage_dimension_order(self, cube_name: str, **kwargs) -> List[str]:
         """ Get the storage dimension order of a cube
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -189,14 +189,14 @@ class CubeService(ObjectService):
             return dimension_names[1:]
         return dimension_names
 
-    def get_names_with_dimension(self, dimension_name: str, model_cubes_only: bool = False,
+    def get_names_with_dimension(self, dimension_name: str, skip_control_cubes: bool = False,
                                  **kwargs) -> List[str]:
         """ Ask TM1 Server for list of cube names that contain specific dimension
 
         :param dimension_name: string, valid dimension name (case insensitive)
-        :param model_cubes_only: bool, True will filter result to model cubes only
+        :param skip_control_cubes: bool, True will exclude control cubes from result
         """
-        cube_prefix = 'Cubes' if model_cubes_only == False else 'ModelCubes()'
+        cube_prefix = 'Cubes' if skip_control_cubes == False else 'ModelCubes()'
         url = format_url(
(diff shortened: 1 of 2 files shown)


TM1py/Services/CubeService.py
```diff
@@ -189,6 +189,40 @@ def get_dimension_names(self, cube_name: str, skip_sandbox_dimension: bool = Tru
             return dimension_names[1:]
         return dimension_names
 
+    def get_names_with_dimension(self, dimension_name: str, model_cubes_only: bool = False,
```
rkvinoth, line 192: Does `get_cube_names_with_dimension` sound better?

TM1py/Services/CubeService.py
```diff
@@ -189,6 +189,40 @@ def get_dimension_names(self, cube_name: str, skip_sandbox_dimension: bool = Tru
...
+        url = format_url(
+            "/api/v1/{}?$select=Name&$filter=Dimensions/any(d: toupper(d/Name) eq toupper('{}'))",
+            cube_prefix, dimension_name
+            )
+        response = self._rest.GET(url, **kwargs)
+        cubes = list(entry['Name'] for entry in response.json()['value'])
+        return cubes
+
+
+    def get_names_with_dimension_wildcard(self, wildcard: str, model_cubes_only: bool = False,
```
rkvinoth, line 209: Does `get_cube_names_with_dimension_wildcard` sound better?

rkvinoth:
Useful functions and very efficiently handled via REST instead of python.
@adscheevel Nice one!

Could you please add some test cases?

TM1py/Services/CubeService.py
adscheevel: I originally considered that, but felt the function name was too long and that the "cubes" was implied since it's in the cube service

TM1py/Services/CubeService.py
adscheevel: similar to above, felt that "cube" was implied

TM1py/Services/CubeService.py
```diff
@@ -189,6 +189,40 @@ def get_dimension_names(self, cube_name: str, skip_sandbox_dimension: bool = Tru
             return dimension_names[1:]
         return dimension_names
 
+    def get_names_with_dimension(self, dimension_name: str, model_cubes_only: bool = False,
+                                 **kwargs) -> List[str]:
+        """ Ask TM1 Server for list of cube names that contain specific dimension
+
+        :param dimension_name: string, valid dimension name (case insensitive)
+        :param model_cubes_only: bool, True will filter result to model cubes only
```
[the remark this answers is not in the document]
adscheevel: model cubes aligns with IBM's naming convention. It's not a required parameter and i've defaulted the value to False so the function will check all cubes if the user is not aware of the parameter.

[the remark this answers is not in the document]
rkvinoth: @MariusWirtz what do you think?

adscheevel:
I won't be able to upload screenshots from my tm1 environment and might need someone else to test it. Funny story, I got a subscription of PA OnDemand yesterday to have a private container in order to take screenshots of test cases before immediately realizing rest api functionality isn't available with OnDemand or trial.

TM1py/Services/CubeService.py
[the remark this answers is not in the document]
MariusWirtz: I think the term "control cubes" is more familiar to most TM1 guys than "model cubes", but that's just my impression.

In the `CubeService` class we already have some optional arguments like `skip_sandbox_dimension`. Perhaps `skip_control_cubes` Would be more in harmony with other parts of TM1py like the `CellService` where we recently added a couple of optional `skip_something` arguments to functions.

TM1py/Services/CubeService.py
MariusWirtz: I also kinda think it's implied

TM1py/Services/CubeService.py
```diff
@@ -189,6 +189,40 @@ def get_dimension_names(self, cube_name: str, skip_sandbox_dimension: bool = Tru
...
+        """
+        cube_prefix = 'Cubes' if model_cubes_only == False else 'ModelCubes()'
+        url = format_url(
+            "/api/v1/{}?$select=Name&$filter=Dimensions/any(d: toupper(d/Name) eq toupper('{}'))",
```
MariusWirtz, line 201: Does this filter take into account TM1's inherent space insensitivity?

Perhaps we need to include an odata string replace within the odata lambda.
Maybe check this function as a reference
https://github.com/cubewise-code/tm1py/blob/f19f58bcfdb65f2ae293762c50cb850d8e4a0bee/TM1py/Services/ObjectService.py#L25

Also, I think it's safer to use `tolower` rather than `toupper`. I learned the other day that the German `ß` is translated to `SS` in some libraries.
https://github.com/cubewise-code/mdxpy/issues/22

MariusWirtz:
Thanks for the contribution! The functions will be useful. I added a few comments to the code. 

No worries. If you don't have an accessible TM1 instance available I can write the tests myself after the merge.

TM1py/Services/ProcessService.py
```diff
@@ -82,6 +82,48 @@ def get_all_names(self, **kwargs) -> List[str]:
         processes = list(process['Name'] for process in response.json()['value'])
         return processes
 
+    def search_string_in_body(self, search_string: str, **kwargs) -> List[str]:
```
MariusWirtz, line 85: Actually, we are not looking the full body of a process (parameters, variables, etc.) but only at the code tabs.

Perhaps it would be more precise to call the function something like `search_string_in_code_tabs`

TM1py/Services/ProcessService.py
```diff
@@ -82,6 +82,48 @@ def get_all_names(self, **kwargs) -> List[str]:
...
+                         "or contains(toupper(MetadataProcedure),toupper('{}')) " \
+                         "or contains(toupper(DataProcedure),toupper('{}')) " \
+                         "or contains(toupper(EpilogProcedure),toupper('{}'))",
+                         search_string, search_string, search_string, search_string
+            )
+        response = self._rest.GET(url, **kwargs)
+        processes = list(process['Name'] for process in response.json()['value'])
+        return processes
+
+    def search_string_in_name(self, name_startswith: str = None, name_contains: Iterable = None,
```
MariusWirtz, line 101: I think it would be a nice feature if it was configurable whether to search with a logical `and` or a logical `or` like what you did in https://github.com/cubewise-code/tm1py/commit/5b3386e14a381f277ee0f2859cf093a476f88edd

TM1py/Services/CubeService.py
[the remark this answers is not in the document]
adscheevel: I like `skip_control_cubes`, will update to that over thee weekend and resolve

TM1py/Services/CubeService.py
adscheevel: I will need to test and confirm

TM1py/Services/ProcessService.py
adscheevel: agreed, i can make that update

TM1py/Services/ProcessService.py
adscheevel: yes good point, i can update to `search_string_in_code` or possibly update the function to search the full body of the process

TM1py/Services/CubeService.py
adscheevel: After testing, I think the functions are working as we'd intend them to, but perhaps I didn't test right.

I tested this by creating a dim "astest" and placing that dim in a cube and then attempted to search for the dim with both cube search functions looking for "as test" and it did not return any results. I then tried the reverse with a dim called "as test" and attempted to search based on dim "astest" and did not have any results returned.

TM1py/Services/CubeService.py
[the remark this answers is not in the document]
adscheevel: updated: https://github.com/cubewise-code/tm1py/pull/658/commits/f2ee1afe6c1b021031243822b2d8addc743dfea6

TM1py/Services/ProcessService.py
adscheevel: updated: https://github.com/cubewise-code/tm1py/pull/658/commits/3a372f9acc16215e47fbbb7d82f4c39d56799a8b

TM1py/Services/ProcessService.py
adscheevel: updated: https://github.com/cubewise-code/tm1py/pull/658/commits/07ace1f769e3e412b0e343e5faa065c39df61921

PR #663, opened 2022-01-11
Chore search methods

adscheevel:
Added two functions for searching for chores based on process name or defined parameter value(not parameter name and not the parameter default set in the TI process) in the chore configuration. I took a stab at some test cases as well.

### DIFF
--- a/TM1py/Services/ChoreService.py
+++ b/TM1py/Services/ChoreService.py
@@ -129,6 +129,43 @@ class ChoreService(ObjectService):
         url = format_url("/api/v1/Chores('{}')", chore_name)
         return self._exists(url, **kwargs)
 
+     def search_for_process_name(self, process_name: str, matching_tasks_only: bool=False, **kwargs) -> List[Chore]:
+        """ Return chore details for any/all chores that contain specified process name in tasks
+
+        :param process_name: string, a valid ti process name; spaces will be elimniated
+        :param matching_tasks_only: bool, True will only return tasks whose process name matches process_name parameter.
+            Be very careful with this, calling chores.update on a filterd task list of a chore will change the structure of the chore
+        """
+        url = format_url(
+            "/api/v1/Chores?$filter=Tasks/any(t: replace(tolower(t/Process/Name), ' ', '') eq '{}')" \
+            "&$expand=Tasks($expand=*,Chore($select=Name),Process($select=Name)" \
+            + (format_url(";$filter=replace(tolower(Process/Name), ' ', '') eq '{}')", process_name.lower().replace(' ', '')
+                         ) if matching_tasks_only else ")"
+              ),
+            process_name.lower().replace(' ', '')
+        )
+        response = self._rest.GET(url, **kwargs)
+        return [Chore.from_dict(chore_as_dict) for chore_as_dict in response.json()['value']]
+
+    def search_for_parameter_value(self, parameter_value: str, matching_tasks_only: bool=False, **kwargs) -> List[Chore]:
+        """ Return chore details for any/all chores that have a specified value set in the chore parameter settings
+            *this will NOT check the process parameter default, rather the defined parameter value saved in the chore
+
+        :param parameter_value: string, will search wildcard for string in parameter value using Contains(string)
+        :param matching_tasks_only: bool, True will only return tasks that that have the defined string in the parameter value
+            Be very careful with this, calling chores.update on a filterd task list of a chore will change the structure of the chore
+        """
+        url = format_url(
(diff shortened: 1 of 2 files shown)


Tests/ChoreService_test.py
```diff
@@ -400,6 +400,22 @@ def test_exists(self):
...
         self.assertTrue(self.tm1.chores.exists(self.chore_name3))
         self.assertFalse(self.tm1.chores.exists(uuid.uuid4()))
+        
+    def test_search_for_process_name_no_match(self):
+        chore_names = self.tm1.chores.search_for_process_name(process_name="NotAProcessName")
+        self.assertEqual([], chore_names)
+
+    def test_search_for_process_name_happy_case(self):
+        chore_names = self.tm1.chores.search_for_process_name(process_name=process_name1)
+        self.assertEqual(self.chore_name1, chore_names[0].name)
```
MariusWirtz, line 410: Maybe also assert that chore_names has a length of 1

Tests/ChoreService_test.py
```diff
@@ -400,6 +400,22 @@ def test_exists(self):
...
+
+    def test_search_for_parameter_value_no_match(self):
+        chore_names = self.tm1.chores.search_for_parameter_value(parameter_value='NotAParamValue')
+        self.assertEqual([], chore_names)
+
+    def test_search_for_parameter_value_happy_case(self):
+        chore_names = self.tm1.chores.search_for_parameter_value(parameter_value='UK')
+        self.assertEqual(self.chore_name1, chore_names[0].name)
```
MariusWirtz, line 418: Same here. I think it would be good to assert that `len(chore_names) == 1`

TM1py/Services/ChoreService.py
```diff
@@ -129,6 +129,43 @@ def exists(self, chore_name: str, **kwargs) -> bool:
         url = format_url("/api/v1/Chores('{}')", chore_name)
         return self._exists(url, **kwargs)
 
+     def search_for_process_name(self, process_name: str, matching_tasks_only: bool=False, **kwargs) -> List[Chore]:
+        """ Return chore details for any/all chores that contain specified process name in tasks
+
+        :param process_name: string, a valid ti process name; spaces will be elimniated
+        :param matching_tasks_only: bool, True will only return tasks whose process name matches process_name parameter.
+            Be very careful with this, calling chores.update on a filterd task list of a chore will change the structure of the chore
```
MariusWirtz, line 137: I don't understand this last part of the docstring about the `chores.update` function. I this comment perhaps out of place or am I missing something?

MariusWirtz: Oh I understand now. If `matching_tasks_only` is True the function returns `Chore` objects that are not equal to the actual chores in TM1.
I think that's dangerous indeed. Have you thought about returning dictionaries instead?

MariusWirtz: Or perhaps we could skip the `matching tasks_only` flag. 
Do you have a use case in mind in which the non-matched tasks would be disturbing?

adscheevel: Dangerous indeed. I liked the option to return a filtered list, particularly when dealing with a large chore. I considered a dictionary, but wanted to utilize the chore class object due to it's familiarity. I think for now, I'd rather remove the matching_tasks_only flag and simply return the full chore details. If that's ok with you, I'll make that update.

MariusWirtz: OK. Thanks!

Tests/ChoreService_test.py
MariusWirtz: Yes. You are right. process_name1 appears in the tasks for both chores.

MariusWirtz: I think this assertion would be fairly complete.
```  
        chore_names = self.tm1.chores.search_for_process_name(process_name=process_name1)
        self.assertEqual(2, len(chore_names))
        self.assertEqual(self.chore_name1, chore_names[0].name)
        self.assertEqual(self.chore_name2, chore_names[1].name)
```

adscheevel: Thanks! i've updated the assertions with this commit 0c1c0d84bdc3bfd8532c6ba9898e41df5740ae9b

Tests/ChoreService_test.py
adscheevel: assertions updated with this commit 0c1c0d84bdc3bfd8532c6ba9898e41df5740ae9b

MariusWirtz:
Nice one @adscheevel 

Thanks!

PR #678, opened 2022-01-27
Use csv module to create csv strings

jordanjeremy:
Current csv string creation can produce invalid strings in the event of something like the delimiter being present in a value.  Propose using the csv module to create these strings instead of just concatenating values together, to ensure correct strings are created.

Adds argument for csv_dialect to allow user to provide settings for all possible formatting options.  If not provided will create dialect based on existing value_separator and line_separator arguments, so they will be used in the event that a user has passed custom values.

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import asyncio
+import csv
 import functools
 import itertools
 import json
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Nice work @jordanjeremy. Thank you.

I will do testing tonight and provide feedback if I find anything.

TM1py/Services/CellService.py
```diff
@@ -2273,9 +2291,9 @@ def extract_cellset_csv_iter_json(
...
 
                 if current_axes == 0:
-                    axes0_list[current_tuple] += ('' if axes0_list[current_tuple] == '' else value_separator) + value
+                    axes0_list[current_tuple].append(value)
                 else:
-                    axes1_list[current_tuple] += ('' if axes1_list[current_tuple] == '' else value_separator) + value
+                    axes1_list[current_tuple].append(value)
 
                 # Add header entry for attribute if necessary
                 if current_tuple == 0:
```
MariusWirtz, line 2299: I think we also need changes in this block _(old lines 2282 to 2292)_

MariusWirtz:
There is a small difference between the new implementation and the old one. 
The new one includes a linebreak in the last row. The old one didn't. 

If it's a small change it would be nice to comply with the old implementation.
Or is there perhaps a good reason to deviate from the old way?

By the way, we have some test cases defined for the `CellService` class. You can find them here:
https://github.com/cubewise-code/tm1py/blob/master/Tests/CellService_test.py
You can find a small readme on the tests here:
https://github.com/cubewise-code/tm1py/blob/master/Tests/_readme.md

Currently, 7 tests fail due to the two issues I raised above.

image

jordanjeremy:
@MariusWirtz that you for the feedback.  I will look into the issues identified and follow up.

MariusWirtz:
The changes look good and all tests pass.

Thanks for the contribution @jordanjeremy!

PR #686, opened 2022-02-09
Fix set_time of ChoreStartTime class

samuelko123:
It attempts to fix 2 issues:
1. We cannot set minute to 0, because "if minute:" will evaluate to False. The same goes for other parameters.
2. We would get "ValueError: day is out of range for month" if we set month to 2 while the original day is 31, vice versa.


MariusWirtz:
Thanks for the fix!

PR #699, opened 2022-03-18
Add `get_parents` function to `ElementService`

### LINKED ISSUE #697: How do I get the parent of a given element?
I need to get the parent of a given element. 

I checked the methods for elements and hierarchies but couldn't find a direct way to do so. I checked the hierarchy.edges but I cannot iterate through it since I don't have the parent name. The rest of the script is not quite relevant, so I feel like it is sufficient to mention that it retrieves an element name. I need to retrieve the immediate parent of that element.

Thanks in advance,

Elçin

MariusWirtz:
Fixes #697

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -626,3 +626,12 @@ class ElementService(ObjectService):
         body = [element_attribute.body_as_dict for element_attribute in element_attributes]
 
         return self._rest.POST(url=url, data=json.dumps(body), **kwargs)
+
+    def get_parents(self, dimension_name: str, hierarchy_name: str, element_name: str, **kwargs) -> List[str]:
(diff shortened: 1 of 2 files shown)


MariusWirtz:
@elcinsayin
please upgrade to this branch and test the `get_parents` function:

```
pip uninstall tm1py
pip install https://github.com/cubewise-code/tm1py/archive/refs/heads/feature/get-parents.zip
```

Let me know how it goes

elcinsayin:
Thank you so much, will try it next week.

elcinsayin:
I tried it. It retrieves all the parents when it's under different consolidated elements and it retrieves the immediate parent when it's under a consolidated element which is also under another consolidated element. Thank you so much.

PR #708, opened 2022-04-04
handle duplicates in write_dataframe

MariusWirtz:


TM1py/Utils/Utils.py
```diff
@@ -696,6 +696,12 @@ def build_cellset_from_pandas_dataframe(df: 'pd.DataFrame') -> 'CaseAndSpaceInse
     """
     if isinstance(df.index, pd.MultiIndex):
         df.reset_index(inplace=True)
+
+    # handle duplicate intersections
```
MariusWirtz, line 700: add optional increment argument

MariusWirtz: treat str values differently

PR #712, opened 2022-04-07
Handle duplictes in df to dict transformation

MariusWirtz:
duplicates in string values are accepted (last one wins)
duplicates in numeric values are aggregated


MariusWirtz:
@Kevin-Dekker 
I'm curious about your thoughts on the data frame operations. Anything we could do more efficiently?

PR #716, opened 2022-05-11
Use WithProcess in write_through_unbound_ti

### LINKED ISSUE #709: Optimize `write_through_unbound_process` through use of `WithProcess`
Instead of calling n unbound processes sequentially from python, we could call one process with n subprocesses.

The `ExecuteProcessWithReturn` action allows to pass a collection of the additional unbound process with your main unbound process through the parameter `WithProcess`.

The processes passed in the WithProcess collection can be called from the main unbound process through `ExecuteProcess`.

Effectively this would allow TM1py to send more than 32k statements at once. This should reduce unnecessary back and forth trips and increase the write performance.

``` XML
<Action Name="ExecuteProcessWithReturn">
	<Annotation Term="Core.Description">
		<String>Runs a process defined in Turbo Integrator and returns back the status.</String>
	</Annotation>
	<Annotation Term="Core.Revisions">
		<Collection>
			<Record>
				<PropertyValue Property="Version">
					<String>11.3.0</String>
				</PropertyValue>
				<PropertyValue Property="Kind">
					<EnumMember>Core.RevisionKind/Added</EnumMember>
				</PropertyValue>
			</Record>
		</Collection>
	</Annotation>
	<Parameter Name="Process" Type="tm1.Process" Nullable="false">
		<Annotation Term="Core.Description">
			<String>Specifies the process to be executed.</String>
		</Annotation>
		<Annotation Term="Core.LongDescription">
			<String>This parameter can either reference an existing process or define a transient process that will only exist for the duration of the execution of such process.</String>
		</Annotation>
	</Parameter>
	<Parameter Name="Parameters" Type="Collection(tm1.NameValuePair)">
		<Annotation Term="Core.Description">
			<String>Override default values (numeric or string) defined in the model.</String>
		</Annotation>
	</Parameter>
	<Parameter Name="WithProcess" Type="Collection(tm1.Process)">
		<Annotation Term="Core.Description">
			<String>Optional. A collection of temporary processes available to the main, transient, process being executed.</String>
		</Annotation>
		<Annotation Term="Core.LongDescription">
			<String>This parameter can be used to specify a collection of temporary processes that will only be available to the main, transient, process being executed. If the process being executed is not a transient process, this parameter will be ignored. Note that these temporary processes can only be executed using the ExecuteProcess function!</String>
		</Annotation>
		<Annotation Term="Core.OptionalParameter">
			<Record></Record>
		</Annotation>
	</Parameter>
	<ReturnType Type="tm1.ProcessExecuteResult" Nullable="false"/>
</Action>
```

MariusWirtz:
Fixes #709

### DIFF
--- a/TM1py/Exceptions/Exceptions.py
+++ b/TM1py/Exceptions/Exceptions.py
@@ -80,21 +80,17 @@ class TM1pyRestException(TM1pyException):
 
 class TM1pyWriteFailureException(TM1pyException):
 
-    def __init__(self, statuses: List[str], error_log_files: List[str]):
-        self.statuses = statuses
+    def __init__(self, error_log_files: List[str]):
         self.error_log_files = error_log_files
 
-        message = f"All {len(self.statuses)} write operations failed. Details: {self.error_log_files}"
-        super(TM1pyWriteFailureException, self).__init__(message)
+        message = f"Write operation failed. Details: '{self.error_log_files}'"
+        super(TM1pyWriteFailureException, self).__init__([message])
 
 
 class TM1pyWritePartialFailureException(TM1pyException):
 
-    def __init__(self, statuses: List[str], error_log_files: List[str], attempts: int):
-        self.statuses = statuses
+    def __init__(self, error_log_files: List[str]):
         self.error_log_files = error_log_files
-        self.attempts = attempts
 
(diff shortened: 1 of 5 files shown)


MariusWirtz:
closed in favor of #774

PR #721, opened 2022-05-21
Search cube rules function

adscheevel:
Adding function to search rules of cube for substring. Similar to prior work searching TI process for substring or searching cubes with dim or dim substring. The function will return list of Cube objects, including full rules file. I debated only returning matching rule lines, but thought that might be dangerous if someone were to inadvertently update a cubes rules with only the matched lines. The function includes the parameter "skip_control_cubes" similar to the other search functions.


adscheevel:
closing as these commits are captured in downstream branch commit

PR #723, opened 2022-05-21
Cube Rule search function and new skip_control parameters

adscheevel:
Sorry for the multiple pull requests that were subsequently closed. I was working on separated, nested branches and created a pull request for each individually before simply merging them all into one here.

1. new function search_for_rule_substring will return all cubes with substring in rules
2. added skip_control_cubes parameter to multiple CubeService functions e.g. get_all_names, get_number_of_cubes
3. added skip_control_dims parameter to DimensionService functions get_all_names and get_number_of_dimensions
4. attempted update to test cases

### DIFF
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -84,11 +84,19 @@ class CubeService(ObjectService):
         cubes = [Cube.from_dict(cube_as_dict=cube) for cube in response.json()['value']]
         return cubes
 
-    def get_number_of_cubes(self, **kwargs) -> int:
-        url = format_url(
-            "/api/v1/Cubes/$count")
-        response = self._rest.GET(url, **kwargs)
-        return int(response.text)
+    def get_number_of_cubes(self, skip_control_cubes: bool = False, **kwargs) -> int:
+        """ Ask TM1 Server for count of cubes
+
+        :skip_control_cubes: bool, True will exclude control cubes from count
+        :return: int, count
+        """ 
+        if skip_control_cubes:
+            response = len(self.get_all_names(skip_control_cubes=True, **kwargs))
+        else:
+            response = int(self._rest.GET(url=format_url("/api/v1/Cubes/$count"), **kwargs).text)
+    
+        return response
+
 
     def get_measure_dimension(self, cube_name: str, **kwargs) -> str:
         url = format_url(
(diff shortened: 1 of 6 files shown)


adscheevel:
added skip_control_processes parameter to functions in ProcessService too

TM1py/Services/CubeService.py
```diff
@@ -84,11 +84,19 @@ def get_control_cubes(self, **kwargs) -> List[Cube]:
...
-        response = self._rest.GET(url, **kwargs)
-        return int(response.text)
+    def get_number_of_cubes(self, skip_control_cubes: bool = False, **kwargs) -> int:
+        """ Ask TM1 Server for count of cubes
+
+        :skip_control_cubes: bool, True will exclude control cubes from count
+        :return: int, count
+        """ 
+        if skip_control_cubes:
+            response = len(self.get_all_names(skip_control_cubes=True, **kwargs))
```
MariusWirtz, line 94: I imagine it would be more efficient if you used `ModelCubes()/$count` here

TM1py/Services/CubeService.py
```diff
@@ -224,6 +250,21 @@ def search_for_dimension_substring(self, substring: str, skip_control_cubes: boo
...
+    
+    def search_for_rule_substring(self, substring: str, skip_control_cubes: bool = False, **kwargs) -> List[Cube]:
+        """ get all cubes from TM1 Server as TM1py.Cube instances where rules for given cube contain specified substring
+
+        :param substring: string to search for in rules
+        :param skip_control_cubes: bool, True will exclude control cubes from result
+        :return: List of TM1py.Cube instances
+        """
+        url = format_url(
+            "/api/v1/{}?$filter=Rules ne null and contains(toupper(Rules),toupper('{}'))&$expand=Dimensions($select=Name)",
```
MariusWirtz, line 262: I wonder if we would like the `contains` check to be space insensitive as well. 
Any thoughts?

TM1py/Services/DimensionService.py
```diff
@@ -128,23 +128,36 @@ def exists(self, dimension_name: str, **kwargs) -> bool:
...
+        """Ask TM1 Server for number of dimensions
 
+        :skip_control_dims: bool, True to exclude control dims from count
         :return: Number of dimensions
         """
-        response = self._rest.GET(url='/api/v1/Dimensions/$count', **kwargs)
-        return int(response.text)
+
+        if skip_control_dims:
+            response = len(self.get_all_names(skip_control_dims=True, **kwargs))
```
MariusWirtz, line 156: I imagine it would be more efficient if you used `ModelDimensions()/$count` here

TM1py/Services/ProcessService.py
```diff
@@ -48,11 +48,14 @@ def get(self, name_process: str, **kwargs) -> Process:
...
         return Process.from_dict(response.json())
 
-    def get_all(self, **kwargs) -> List[Process]:
+    def get_all(self, skip_control_processes: bool = False, **kwargs) -> List[Process]:
         """ Get a processes from TM1 Server
     
+        :param skip_control_processes: bool, True to exclude processes that begin with "}" or "{"
         :return: List, instances of the TM1py.Process
         """
+        model_process_filter = "&$filter=startswith(Name,'}') eq false and startswith(Name,'{') eq false"
```
MariusWirtz, line 57: I think here it should say `or` instead of `and`

Tests/CubeService_test.py
```diff
@@ -135,6 +137,12 @@ def test_get_all_names(self):
 
         self.tm1.cubes.delete(cube_name)
 
+        cube = self.tm1.cubes.get_control_cubes()[0]
```
MariusWirtz, line 140: tests should be non-invasive in a way that they only manipulate objects that they created themself and not interact with the rest of the model. 

Please create a dummy control cube instead

MariusWirtz:
Hi @adscheevel,

thanks for the PR. Nice work!
I added a few comments to the code. Mostly related to style. Let me know what you think :)

TM1py/Services/CubeService.py
adscheevel: It certainly would be more efficient except `ModelCubes()/$count`  or `ModelCubes/$count` don't work like `Cubes/$count` does unfortunately. I got lazy and leveraged len on the response of the get_all_names function. I've got an alternate method that grabs the @odata.count value from a response that I'll update today.

Tests/CubeService_test.py
adscheevel: That's good feedback, I'll update.

adscheevel: updated with https://github.com/cubewise-code/tm1py/pull/723/commits/3ae357020d0793f6bbdf595af7137860c3169ae9

TM1py/Services/CubeService.py
adscheevel: I think that makes sense and will update the odata filter accordingly. Makes me wonder then if we should update the search_string_in_code function in ProcessService to also be space insensitive?

MariusWirtz: Thanks.

Yeah. I also think it would also be nice to keep it consistent.
Let's do this in a separate PR before the next release.

MariusWirtz:
I will merge this feature as is. Thanks again!

I think for the case and space insensitive search through code (rules and TI) we should look at this in a separate Pull Request. Perhaps it makes sense to pass an additional optional argument (something like `case_and_space_insensitive`) to the functions to control the behaviour.

PR #726, opened 2022-05-24
Minor changes and adjustments

MariusWirtz:
@adscheevel
I made a few small changes. Please review and approve.


adscheevel (approved):
Changes look good, thank you for those improvements! I think I'm done for now with search functions, but I promise next time to use "tolower" instead of my default "toupper", ha.

MariusWirtz:
Thanks!! Haha no worries :)

Only thing missing is the optional space sensitivity on the process functions. It would be nice to include it before the next release I think.

PR #727, opened 2022-05-24
ProcessDebugBreakpoint: Initial commit

adscheevel:
Initial commit with ProcessDebugBreakpoint object and additional debug actions in ProcessService from the existing StepOver and StepOut.

I've also added a function `ti_formula()` to replicate the evaluate option of the TI variable editor in architect/perspectives. It's not a light-weight operation as it involves creating a temporary TI in the same fashion as the `execute_ti_code()` function. It then starts a debug session, sets a breakpoint, and runs until the breakpoint is hit, returning the variable value at that breakpoint before cleaning up the temp TI. The idea for this came from a colleague asking me how to evaluate a variable formula in the PAW TI editor like they frequently do in architect. I wasn't aware of similar functionality in PAW TI editor so created this as a stop-gap until IBM builds something in natively.

I'm expecting to be out on parental leave any day, so won't have time to focus on this for a bit. Below are some initial cleanup ideas/questions that I'd like help with.

1. The object ProcessDebugBreakpoint is a mouthful. If the word "breakpoint" is unique to a process debug context, we could rename the object to simply Breakpoint or DebugBreakpoint
2. Current debug session actions (ie continue, step_out, step_over) return the fully expanded response of the request URL. Does this fit with how we imagine people will use the methods?
3. New functions need docstrings
4. The breakpoint_type parameter takes the full string used for the url e.g. ProcessDebugDataBreakpoint, when this could be refactored to simply be Data, Line, Lock and add the necessary prefix/suffix when building the URL.
5. Validation of breakpoint_type parameter values needed
6. HitMode, LockMode, and LockScope are enumerated types. I don't know exactly how they should be considered differently or any validations that may be needed around their use.

### DIFF
--- a/TM1py/Objects/ProcessDebugBreakpoint.py
+++ b/TM1py/Objects/ProcessDebugBreakpoint.py
@@ -1,0 +1,186 @@
+# -*- coding: utf-8 -*-
+import json
+import re
+from typing import Optional, Iterable, Dict, List, Union
+
+from TM1py.Objects.TM1Object import TM1Object
+
+
+class ProcessDebugBreakpoint(TM1Object):
+    """ Abstraction of a TM1 Process Debug Breakpoing.
+
+    """
+
+
+    def __init__(self,
+               breakpoint_type: str = '',
+               breakpoint_id: int = 1,
+               enabled: bool = True,
+               hitmode: str = '',
+               hitcount: int = 0,
+               expression: str = '',
+               variable_name: str = '',
+               process_name: str = '',
+               procedure: str = '',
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Thanks for the MR. I will merge your PR into the [cubewise-code:feature/process-debug-breakpoint](https://github.com/cubewise-code/tm1py/tree/feature/process-debug-breakpoint) branch. 

Let's continue the collaboration there.
I'm not sure I fully grasp all the ideas yet. I will play with the code I try to understand.

PR #735, opened 2022-06-01
Feature/process debug breakpoint

MariusWirtz:
Based on #727

### DIFF
--- a/TM1py/Objects/ProcessDebugBreakpoint.py
+++ b/TM1py/Objects/ProcessDebugBreakpoint.py
@@ -52,10 +52,10 @@ class ProcessDebugBreakpoint(TM1Object):
 
     def __init__(
             self,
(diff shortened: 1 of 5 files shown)


MariusWirtz:
@adscheevel 
I made a few small changes. In order to avoid conflicts with the built-in name `breakpoint,` I suggest we name the variable `break_point` in py.
I wonder if we should perhaps rename the `ti_formula` to something similar to the existing `execute_ti_code` because it kinda does the same thing. Except that it only executes one "statement" and returns the result of the statement back to py.

What do you think about `evaluate_ti_statement`? Or is that too long perhaps?

adscheevel:
I like the evaluate_ti_expression you went with for the function. It is long, but best describes the functinality.

Thank you for all the improvements you've added and test cases!

PR #749, opened 2022-06-16
Alignment of float numbers when writing back through unbound process

pbuncik:
This code allows for automatic alignment of float numbers when writing back through unbound process to avoid "Number too big" errors during writeback.


MariusWirtz:
Thanks @pbuncik

Issue #738, opened 2022-06-07
Subset search functions for native cube views


MariusWirtz:
It would be nice to have `search` functions in the `ViewService` to find all views that reference a subset.

Kinda similar to the `search` functions in the `ProcessService`
https://github.com/cubewise-code/tm1py/blob/7358f7bf7d372b18fb82f93a36336aa79304c8cf/TM1py/Services/ProcessService.py#L93

MariusWirtz:
closed with #752

PR #751, opened 2022-06-17
View subset search

adscheevel:
Method for finding views that contain specified dim/subset as requested in this issue: https://github.com/cubewise-code/tm1py/issues/738#issue-1263244896

I included an enhancement to the get_all function in View Service to specify whether or not the returned list of views should include elements. By excluding the element list, the get_all function leveraged by this new search method runs 10x faster.

I put the subset search function in the cube service as I needed to leverage the get_names_with_dimension function and didn't want to mess with the dependencies.

Ideally, there would be a good odata query to do all of this right in the URL, but that's proving somewhat complex.

### DIFF
--- a/TM1py/Services/ViewService.py
+++ b/TM1py/Services/ViewService.py
@@ -208,3 +208,57 @@ class ViewService(ObjectService):
         url = format_url("/api/v1/Cubes('{}')/{}('{}')", cube_name, view_type, view_name)
         response = self._rest.DELETE(url, **kwargs)
         return response
+
+    def search_subset_in_view(self, dimension_name: str = None, subset_name: str = None, cube_name: str = None, 
+                              include_elements: bool = True, **kwargs) -> Tuple[List[View], List[View]]:
+        """ Get all public and private views that utilize specified dimension subset
+
+        :param dimension_name: string, valid dimension name with subset to query
+        :param subset_name: string, valid subset name to search for in views
+        :param cube_name: str, optionally specify cube to search, otherwise will search all cubes
+        :param include_elements: false to return view details without elements, faster
+        :return: 2 Lists of TM1py.View instances: private views, public views
+        """
+
+        element_filter = ";$top=0" if not include_elements else ""
+        cube_filter = "('{}')".format(cube_name) if cube_name else ""
+        
+        private_views, public_views = [], []
+        for view_type in ('PrivateViews', 'Views'):
+            url = format_url(
+                "/api/v1/Cubes{}?$expand={}("
+                "$filter=isof(tm1.NativeView) and"
+                "("
+                "(tm1.NativeView/Rows/any (r: r/Subset/Name eq '{}' and r/Subset/Hierarchy/Dimension/Name eq '{}'))"
+                "or"
+                "(tm1.NativeView/Columns/any (c: c/Subset/Name eq '{}' and c/Subset/Hierarchy/Dimension/Name eq '{}')) "
+                "or"
+                "(tm1.NativeView/Titles/any (t: t/Subset/Name eq '{}' and t/Subset/Hierarchy/Dimension/Name eq '{}'))"
+                ");"
+                "tm1.NativeView/Rows/Subset($expand=Hierarchy($select=Name;"
+                "$expand=Dimension($select=Name)),Elements($select=Name{});"
+                "$select=Expression,UniqueName,Name, Alias),  "
+                "tm1.NativeView/Columns/Subset($expand=Hierarchy($select=Name;"
+                "$expand=Dimension($select=Name)),Elements($select=Name{});"
+                "$select=Expression,UniqueName,Name,Alias), "
+                "tm1.NativeView/Titles/Subset($expand=Hierarchy($select=Name;"
+                "$expand=Dimension($select=Name)),Elements($select=Name{});"
+                "$select=Expression,UniqueName,Name,Alias), "
+                "tm1.NativeView/Titles/Selected($select=Name))",
+                cube_filter, view_type, subset_name, dimension_name, subset_name, dimension_name, subset_name, dimension_name, 
+                element_filter, element_filter, element_filter)
+
+            response = self._rest.GET(url, **kwargs)
+            response_as_list = response.json()['value']
+            for cube in response_as_list:
+                if cube[view_type]:
+                    for view_as_dict in cube[view_type]:
+                        if view_as_dict['@odata.type'] == '#ibm.tm1.api.v1.MDXView':
+                            view = MDXView.from_dict(view_as_dict, cube_name)
+                        else:
+                            view = NativeView.from_dict(view_as_dict, cube_name)
+                        if view_type == "PrivateViews":
+                            private_views.append(view)
+                        else:
+                            public_views.append(view)
+        return private_views, public_views
--- a/TM1py/Services/ViewService.py
+++ b/TM1py/Services/ViewService.py
@@ -235,7 +235,7 @@ class ViewService(ObjectService):
                 "or"
                 "(tm1.NativeView/Titles/any (t: t/Subset/Name eq '{}' and t/Subset/Hierarchy/Dimension/Name eq '{}'))"
                 ");"
-                "tm1.NativeView/Rows/Subset($expand=Hierarchy($select=Name;"
+                "$expand=tm1.NativeView/Rows/Subset($expand=Hierarchy($select=Name;"
                 "$expand=Dimension($select=Name)),Elements($select=Name{});"
                 "$select=Expression,UniqueName,Name, Alias),  "
                 "tm1.NativeView/Columns/Subset($expand=Hierarchy($select=Name;"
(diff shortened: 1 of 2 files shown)


gbryant-dev:
Something like this seems to work for a OData query to find any views that contains a given dim and subset. Note the type filter using `isof` to restrict to native views before attempting to filter on nested properties. Probably needs more testing but appears to do the trick. 

```
Cubes?$select=Name&$expand=Views($select=Name;
  $filter=isof(tm1.NativeView) and 
    (
        (tm1.NativeView/Rows/any (r: r/Subset/Name eq 'Default' and r/Subset/Hierarchy/Dimension/Name eq 'Version'))
        or 
       (tm1.NativeView/Columns/any (r: r/Subset/Name eq 'Default' and r/Subset/Hierarchy/Dimension/Name eq 'Version')) 
        or 
       (tm1.NativeView/Titles/any (r: r/Subset/Name eq 'Default' and r/Subset/Hierarchy/Dimension/Name eq 'Version'))
     )
)
```

Note sure i'd necessarily pick it over the approach you've taken but it's good to know it's possible through OData.

MariusWirtz:
Hi,

nice work @adscheevel and @gbryant-dev. Thanks!

I assume the odata query does the filtering more efficiently. I think we should implement the search function based on the query.

In any case, I think we should merge the change that the view skips the element names. That's a useful feature.

adscheevel:
Thanks @gbryant-dev! I agree an odata query would be preferred and almost surely more efficient. The only issue I have with this query is that it will return a list of all cubes and the matching views for each cube. The result will be a bunch of cubes with null results. This can easily be cleaned up on the python side but it would be nice to have the cube list filtered too, that is a filter both inside and outside of the $expand statement.

gbryant-dev:
It's possible but I can't find a better way of doing it without duplicating the filter like this:

```Cubes?$select=Name&$filter=Views/any (v:v/isof(tm1.NativeView) and ((tm1.NativeView/Rows/any (r: r/Subset/Name eq 'Default' and r/Subset/Hierarchy/Name eq 'Version')) or (tm1.NativeView/Columns/any (r: r/Subset/Name eq 'Default' and r/Subset/Hierarchy/Name eq 'Version')) or (tm1.NativeView/Titles/any (r: r/Subset/Name eq 'Default' and r/Subset/Hierarchy/Name eq 'Version'))))&$expand=Views($select=Name;$filter=isof(tm1.NativeView) and ((tm1.NativeView/Rows/any (r: r/Subset/Name eq 'Default' and r/Subset/Hierarchy/Name eq 'Version')) or (tm1.NativeView/Columns/any (r: r/Subset/Name eq 'Default' and r/Subset/Hierarchy/Name eq 'Version')) or (tm1.NativeView/Titles/any (r: r/Subset/Name eq 'Default' and r/Subset/Hierarchy/Name eq 'Version'))))```

**Comment by adscheevel** [comment]
I was afraid of that, knew the filter would need to be present in both places but wasn't sure how messy it'd be. My latest commit https://github.com/cubewise-code/tm1py/pull/751/commits/4da336cc80120d36b461c49153a82d1f39543659 splits the difference and uses @gbryant-dev's filter in the expand statement and then filters out the null view collections via python. I removed my original function in cube service via this commit https://github.com/cubewise-code/tm1py/pull/751/commits/4591636958d664f252e0dad7c24a4aeb62e436c8.

**Comment by MariusWirtz** [comment]
Nice work! I will review and add tests

File: TM1py/Services/ViewService.py [hunk 1]
```diff
@@ -205,3 +208,57 @@ def delete(self, cube_name: str, view_name: str, private: bool = False, **kwargs
...
+                "$select=Expression,UniqueName,Name, Alias),  "
+                "tm1.NativeView/Columns/Subset($expand=Hierarchy($select=Name;"
+                "$expand=Dimension($select=Name)),Elements($select=Name{});"
+                "$select=Expression,UniqueName,Name,Alias), "
+                "tm1.NativeView/Titles/Subset($expand=Hierarchy($select=Name;"
+                "$expand=Dimension($select=Name)),Elements($select=Name{});"
+                "$select=Expression,UniqueName,Name,Alias), "
+                "tm1.NativeView/Titles/Selected($select=Name))",
+                cube_filter, view_type, subset_name, dimension_name, subset_name, dimension_name, subset_name, dimension_name, 
+                element_filter, element_filter, element_filter)
```
gbryant-dev, line 249: This might work better as an f string as the single quotes around the cube name for `cube_filter` are going to be escaped by the `format_url` function resulting in something like this `Cubes(''General Ledger'')` which is not valid

TM1py/Services/ViewService.py
```diff
@@ -205,3 +208,57 @@ def delete(self, cube_name: str, view_name: str, private: bool = False, **kwargs
...
+        :return: 2 Lists of TM1py.View instances: private views, public views
+        """
+
+        element_filter = ";$top=0" if not include_elements else ""
+        cube_filter = "('{}')".format(cube_name) if cube_name else ""
+        
+        private_views, public_views = [], []
+        for view_type in ('PrivateViews', 'Views'):
+            url = format_url(
+                "/api/v1/Cubes{}?$expand={}("
```
gbryant-dev, line 229: Is it worth being more specific on the properties e.g Name only to retrieve for the cube if we're only interested in the views? By default, this is going to return the rules for each cube amongst other properties

TM1py/Services/ViewService.py
```diff
@@ -205,3 +208,57 @@ def delete(self, cube_name: str, view_name: str, private: bool = False, **kwargs
...
+                "$select=Expression,UniqueName,Name,Alias), "
+                "tm1.NativeView/Titles/Subset($expand=Hierarchy($select=Name;"
+                "$expand=Dimension($select=Name)),Elements($select=Name{});"
+                "$select=Expression,UniqueName,Name,Alias), "
+                "tm1.NativeView/Titles/Selected($select=Name))",
+                cube_filter, view_type, subset_name, dimension_name, subset_name, dimension_name, subset_name, dimension_name, 
+                element_filter, element_filter, element_filter)
+
+            response = self._rest.GET(url, **kwargs)
+            response_as_list = response.json()['value']
```
gbryant-dev, line 252: I don't think this is going to work for a single cube as the `value` property won't exist?

TM1py/Services/ViewService.py
```diff
@@ -205,3 +208,57 @@ def delete(self, cube_name: str, view_name: str, private: bool = False, **kwargs
...
+            for cube in response_as_list:
+                if cube[view_type]:
+                    for view_as_dict in cube[view_type]:
+                        if view_as_dict['@odata.type'] == '#ibm.tm1.api.v1.MDXView':
```
gbryant-dev, line 256: I don't think you need to check for a MDX view here as `$filter=isof(tm1.NativeView)` will only return native views. With that in mind, a better suited function name might be `search_subset_in_native_view` so it's clearer? I think you'd have to rely on regex to find subsets in an MDX view - sounds like a future PR 😉

TM1py/Services/ViewService.py
```diff
@@ -205,3 +208,57 @@ def delete(self, cube_name: str, view_name: str, private: bool = False, **kwargs
...
+        for view_type in ('PrivateViews', 'Views'):
+            url = format_url(
+                "/api/v1/Cubes{}?$expand={}("
+                "$filter=isof(tm1.NativeView) and"
+                "("
+                "(tm1.NativeView/Rows/any (r: r/Subset/Name eq '{}' and r/Subset/Hierarchy/Dimension/Name eq '{}'))"
+                "or"
+                "(tm1.NativeView/Columns/any (c: c/Subset/Name eq '{}' and c/Subset/Hierarchy/Dimension/Name eq '{}')) "
+                "or"
+                "(tm1.NativeView/Titles/any (t: t/Subset/Name eq '{}' and t/Subset/Hierarchy/Dimension/Name eq '{}'))"
```
gbryant-dev, line 236: Should these be case and space insensitive?

gbryant-dev (commented):
I've left some comments 🙂

TM1py/Services/ViewService.py
adscheevel: good catch, I had $select=name in an earlier test and somehow dropped it.

TM1py/Services/ViewService.py
adscheevel: You're right, I had copied from the `get_all` function. i've removed it .

i've renamed the function per your suggestion with https://github.com/cubewise-code/tm1py/pull/751/commits/95da2b6ffd933eedcd4a272f7ed5aae0bfc915f9

TM1py/Services/ViewService.py
adscheevel: yes i think so. I've updated with https://github.com/cubewise-code/tm1py/pull/751/commits/c5c2e5f0813dee39d822f4353dfbc6b98d57952e

TM1py/Services/ViewService.py
adscheevel: Yep, wasn't going to work when cube_name was specified. the update in this commit should now work both ways: https://github.com/cubewise-code/tm1py/pull/751/commits/7966e1dcda7e15c3bc9b2332b247428ab13751eb

TM1py/Services/ViewService.py
adscheevel: I didn't test the cube_name parameter outside of postman, shame on me. this is fixed with https://github.com/cubewise-code/tm1py/pull/751/commits/be1fdd346c8f66751b5f48af13aa8762b05326c1

adscheevel:
Thank you @gbryant-dev for the suggestions and review of my poorly tested code!

MariusWirtz:
Closed in favor of #752 

Hi @adscheevel, @gbryant-dev,

I add some tests to the code. Please approve #752 and I will merge the changes.

PR #752, opened 2022-06-28
Feature/view subset search

MariusWirtz:
successor of #751


adscheevel:
Updated search function looks good and works when I test in my environment, thanks!

PR #760, opened 2022-07-12
Update Hierarchy.py

skriptmeister42:
As of today, information on parent-child relationships (i.e. edges), including consolidation weight, of a hierarchy tree is missing.
The method "get_descendant_edges" addresses this gap.


MariusWirtz:
Thanks!

Issue #748, opened 2022-06-16
Update Applications


cubewise-gng:
**Describe the bug**
The current function update_or_create_document_from_files(), is essentially deleting and recreating the application.
So is not really doing an update.
By uploading a new file to TM1 it is assigned a new system name so any reference to the old name will not work. eg used as menu links.

Could there also be a update() function by passing in the Application object?

**Expected behavior**
Update existing applications but retain the existing system name.

**Version**
TM1py 1.9.1

MariusWirtz:
This issue should be addressed with #762.

@cubewise-gng please confirm that it has been resolved for you.

cubewise-gng:
Thanks guys the update function works well.

PR #762, opened 2022-07-18
Add update to patch file

jrobinsonLOR:
@MariusWirtz 

Re #748; I had a go at creating an update method for the application with some additional DocumentApplication methods to patch the document on the server rather than delete and recreate.


MariusWirtz:
Nice work @jrobinsonLOR. Thank you!

PR #767, opened 2022-07-20
Feature element functions

rclapp:
New features to resolve https://github.com/cubewise-code/tm1py/issues/758


MariusWirtz:
Nice one! Thanks, @rclapp!

Let's see how we can bring this together with #770

PR #770, opened 2022-07-21
Feature element ancestors

Kevin-Dekker:
For large/deep dimensions, it could be more efficient to use the descendants-based method. Especially when checking non-leaf level elements for ancestors this method might have performance benefits.

image

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -3,7 +3,7 @@ import json
 from typing import List, Union, Iterable, Optional, Dict, Tuple
 
 from requests import Response
-
+from mdxpy import MdxHierarchySet
 from TM1py.Exceptions import TM1pyRestException
 from TM1py.Objects import ElementAttribute, Element
 from TM1py.Services.ObjectService import ObjectService
@@ -648,3 +648,19 @@ class ElementService(ObjectService):
     def get_element_principal_name(self, dimension_name: str, hierarchy_name: str, element_name: str, **kwargs) -> str:
         element = self.get(dimension_name, hierarchy_name, element_name, **kwargs)
         return element.name
+
+    def element_is_descendant_of(self, dimension_name: str, hierarchy_name: str, descendant, ancestor,
+                                   max_depth: int = None):
+        if not self.get(dimension_name, hierarchy_name, ancestor).element_type == Element.Types.CONSOLIDATED:
+            raise ValueError('parent must be a consolidated element.')
+
+        return descendant in self.get_members_under_consolidation(dimension_name, hierarchy_name, ancestor, max_depth)
+
+    def element_is_parent_of(self, dimension_name: str, hierarchy_name: str, parent: str, child: str):
+        return parent in self.get_parents(dimension_name, hierarchy_name, child)
+
+    def element_is_ancestor_of(self, dimension_name: str, hierarchy_name: str, ancestor: str, member: str, mode: str):
+        return member in self.execute_set_mdx(
+            MdxHierarchySet(dimension_name, hierarchy_name).ancestors(member).to_mdx())
+
+    # mdx variant & TI process variant, get parents until root variant, drilldown vs drillup, get_ancestors in REST
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -3,7 +3,7 @@ import json
 from typing import List, Union, Iterable, Optional, Dict, Tuple
 
 from requests import Response
-from mdxpy import MdxHierarchySet
+from mdxpy import MdxHierarchySet, Member
 from TM1py.Exceptions import TM1pyRestException
 from TM1py.Objects import ElementAttribute, Element
 from TM1py.Services.ObjectService import ObjectService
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Thanks @Kevin-Dekker!

I understand that calculating the distance and using the precise DESCENDANTS function is faster in most scenarios.
Except for the last scenario "Layer 9 Product 3". I understand that this is because the relationship between the ancestor and child is simple and the initial overhead of retrieving the element levels in the DESCENDANTS approach doesn't pay off in terms of performance.

You mentioned that if the `ancestor` or `child` doesn't exist, the function returns an error. 
I tend to think that that is a good thing, but perhaps it's nice if we make this configurable.

For the `element_is_parent_of` function I guess @rclapp's approach is superior. Retrieving all parents could be expensive in a dimension where elements roll up into lots of consolidation. (e.g. date dimensions with YTD, MAT rollups).

I think it would be good to use mdxpy for the MDX building. It should take care of some of the edge cases like having a `]` in an element name.

MariusWirtz:
I am not sure if we should offer both approaches in the function if the DESCENDANTS approach is faster in most cases.
As a user, I think I would rather choose if potential not-existing elements raise an error.
What do you think @Kevin-Dekker, @rclapp?

rclapp:
In my testing I found that descendants was almost always slower than tm1drilldownmember (using a dimension of 2M elements). Regardless of what mdx function we use I would tend toward the simple implementation, the closer to TI the better. If users desire a more robust option it's easy for them to write it.

The exception path is an interesting one. In TI there seems to be no errors thrown even if the entire set of parameters refer to non existent objects. I started down that path, but realized that it would be problematic if we suppressed api level exceptions. So, I ended up with a compromise. Set expressions that refer to non existent members still work, and just return an empty set, and therefore a false (this is default api behavior). However, queries that use a non existent dimension raise an exception. I tried to add member checking, but decided to remove it since the api call does actually work.

Sent from my mobile device



On Jul 21, 2022 8:05 AM, Marius Wirtz ***@***.***> wrote:

I am not sure if we should offer both approaches in the function if the DESCENDANTS approach is faster in most cases.
As a user, I think I would rather choose if potential not-existing elements raise an error.
What do you think @Kevin-Dekker<https://github.com/Kevin-Dekker>, @rclapp<https://github.com/rclapp>?

-

rclapp:
On the performance side, there is a gain using the cardinality to tell you if the relationship exists so I recommend we stick with that implementation.

On the MDX side, as you stated, it is the distance that drives the performance. Whenever the distance includes "BEFORE" the performance takes a real hit. The MDX defaults are the worse performing of the them all , SELF_BEFORE_AFTER. I am fine changing to Descendants if we know the permutation that preforms the best. 

TBH, I tend to avoid using it because it is so overloaded and not super intuitive.

MariusWirtz:
> On the performance side, there is a gain using the cardinality to tell you if the relationship exists so I recommend we stick with that implementation.

Agree!

> On the MDX side, as you stated, it is the distance that drives the performance. Whenever the distance includes "BEFORE" the performance takes a real hit. The MDX defaults are the worse performing of the them all , SELF_BEFORE_AFTER. I am fine changing to Descendants if we know the permutation that preforms the best.

Do you mean retrieve the (TM1) levels of the elements, calculate and distance and limit the expansion in the `DESCENDANTS` as @Kevin-Dekker suggested? 
I would be very interested to learn if this speeds things up in your test case as well.

I agree. We could probably use a flag that performs better than the "SELF_BEFORE_AFTER" since we already know the distance between ancestor and child.

rclapp:
I ran a few tests and DESCENDANTS is 2x longer when testing against the full structure of the tree. Level 2 vs Level 0.

- `INTERSECT({TM1DRILLDOWNMEMBER({[Big Dim].[Big Dim].[All Divisors]}, ALL, RECURSIVE)}, {[Big Dim].[Big Dim].[5]})` 7.9s

Taking the level of the element being tested
- `INTERSECT({DESCENDANTS([Big Dim].[Big Dim].[All Divisors], [Big Dim].[Big Dim].[5].level)}, {[Big Dim].[Big Dim].[5]}) `12.2s

Taking calculating the difference between the elements (= 2 - 0)

- `INTERSECT({DESCENDANTS([Big Dim].[Big Dim].[All Divisors], 2)}, {[Big Dim].[Big Dim].[5]}) ` 12.3s

However, when testing a 1 level drill Descendants is faster. 
- `INTERSECT({TM1DRILLDOWNMEMBER({[Big Dim].[Big Dim].[All Divisors]}, ALL, RECURSIVE)}, {[Big Dim].[Big Dim].[Div by 3]})` 6s

- `INTERSECT({DESCENDANTS([Big Dim].[Big Dim].[All Divisors], [Big Dim].[Big Dim].[div by 3].level)}, {[Big Dim].[Big Dim].[div by 3]})` 1.2s

rclapp:
Using the API to request levels takes about a second each as well.

Kevin-Dekker:
Interesting findings! I think they are in line with my tests & expectations; the descendant method is slower than the drilldown method on ancestor-element tuples where the element is leaf level, but faster when the element is above leaf level. 

I ran some extra tests using the same dimension as @rclapp. The tests (n=20 per element) confirm this:
image
_run time on x-axis_

Further, I agree with the objective to keep it simple for the user, but I kept the method choice in because the runtime of this function might be essential. 

If the descendants method is really less efficient to retrieve leaf level elements, we could retrieve the element level and in case it's leaf use drilldown else use descendants, this has the following test (n=20 per element) results:
image

Note that there are some outliers in the graphs due to the small sample size. 

The descendant method does seem to add value with big/deep dimensions when the element in question is not leaf. 
What are your thoughts?

rclapp:
I think exposing the method makes sense. We could pick a default so users that are not so performance focused can get started faster.

We can abstract the method used in the query by adding a parameter here https://github.com/cubewise-code/tm1py/pull/767/files#:~:text=def%20_build_drill_intersection_mdx(self%2C%20dimension_name%3A%20str%2C%20hierarchy_name%3A%20str%2C%20first_element_name%3A%20str%2C

Then just expose that option in the public method and compare it to an ENUM

class MDXDrillMethod(Enum):
    TM1DrillDownMember= "TM1DRILLDOWNMEMBER"
    Descendants = "DESCENDANTS"


From: Kevin-Dekker ***@***.***>
Date: Thursday, July 21, 2022 at 1:15 PM
Cc: "Clapp, Ryan" ***@***.***>, Mention ***@***.***>
Subject: Re: [cubewise-code/tm1py] Feature element ancestors (PR #770)


Interesting findings! I think they are in line with my tests & expectations; the descendant method is slower than the drilldown method on ancestor-element tuples where the element is leaf level, but faster when the element is above leaf level.

I ran some extra tests using the same dimension as @rclapp<https://github.com/rclapp>. The tests (n=20 per element) confirm this:
[image]<https://user-images.githubusercontent.com/101728567/180270570-7ff3ff14-19c4-4827-9328-6d16143504db.png>
run time on x-axis

Further, I agree with the objective to keep it simple for the user, but I kept the method choice in because the runtime of this function might be essential.

If the descendants method is really less efficient to retrieve leaf level elements, we could retrieve the element level and in case it's leaf use drilldown else use descendants, this has the following test (n=20 per element) results:
[image]<https://user-images.githubusercontent.com/101728567/180263313-1a75207c-b7dd-4805-a522-5fd9a09aea3c.png>

Note that there are some outliers in the graphs due to the small sample size.

The descendant method does seem to add value with big/deep dimensions when the element in question is not leaf.
What are your thoughts?

MariusWirtz:
I merged #767 already.
@Kevin-Dekker please rebase this one on the current master. 

An Enum to choose between the two approaches sounds reasonable.

If we offer two different MDX approaches, I think they should behave consistently regarding nonexisting elements.

I wonder if we should default to the approach that is performing better on leaves _(testing leaves seems the more common use case to me)_ or instead, check the type of the second element and then default to the approach that is more appropriate?

rclapp:
I'm working on the enum, but found that descendants needs some work in mdxpy so I am doing that too.

Sent from my mobile device

On Jul 22, 2022 8:20 AM, Marius Wirtz ***@***.***> wrote:

I merged #767<https://github.com/cubewise-code/tm1py/pull/767> already.
@Kevin-Dekker<https://github.com/Kevin-Dekker> please rebase this one on the current master.

An Enum to choose between the two approaches sounds reasonable.

If we offer two different MDX approaches, I think they should behave consistently regarding nonexisting elements.

I wonder if we should default to the approach that is performing better on leaves (testing leaves seems the more common use case to me) or instead, check the type of the second element and then default to the approach that is more appropriate?

-

MariusWirtz:
closed in favor of #771

PR #771, opened 2022-07-25
Update Element Parent and Ancestor methods

rclapp:
Added the ability to select different methods for the Element functions. This requires an update to MDXpy

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -4,13 +4,19 @@ from typing import List, Union, Iterable, Optional, Dict, Tuple
 
 from requests import Response
 
-from TM1py.Exceptions import TM1pyRestException
+from TM1py.Exceptions import TM1pyRestException, TM1pyException
 from TM1py.Objects import ElementAttribute, Element
 from TM1py.Services.ObjectService import ObjectService
 from TM1py.Services.RestService import RestService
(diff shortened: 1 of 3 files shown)


TM1py/Services/ElementService.py
```diff
@@ -659,14 +664,22 @@ def _get_mdx_set_cardinality(self, mdx: str) -> int:
...
 
         first_member = Member.of(dimension_name, hierarchy_name, first_element_name)
         second_member = Member.of(dimension_name, hierarchy_name, second_element_name)
-        mdx = MdxHierarchySet.members([first_member]).tm1_drill_down_member(all=True, recursive=recursive)\
-            .intersect(MdxHierarchySet.members([second_member])).to_mdx()
 
+        if not hasattr(MDXDrillMethod, mdx_method.upper()):
+            raise TM1pyException('Invalid MDX Drill Method Specified, Options: TM1DrillDownMember or Descendants')
+        elif MDXDrillMethod.DESCENDANTS.name == mdx_method.upper():
+            mdx = MdxHierarchySet.members([first_member]).tm1_drill_down_member(all=True, recursive=recursive) \
```
MariusWirtz, line 677: @rclapp I think we should use `descendants` here in this condition and `tm1_drill_down_member` in the elif below.

MariusWirtz:
Thanks, @rclapp, and @Kevin-Dekker!

I will add a comment to the `element_is_ancestor` function about when to use which MDX.

PR #773, opened 2022-07-27
Introduce new method in `element_is_ancestor` function: `TI`

MariusWirtz:


MariusWirtz:
@rclapp or @Kevin-Dekker if you have time, please review the PR and approve

rclapp:
Few questions
1) Do we want to check if the Hierarchy/Dim/Element exists? That seems like extra overhead. TI allows invalid elements, hierarchies, and dims. Returning False if objects don't exist. Do these checks match other methods in TM1py?

2) Are there risks to using TI debugging like this? While it may be faster It may have unintended locking consequences and could carry risks if the script drops its connection and the server is left debugging a TI.

MariusWirtz:
Thanks for the feedback!

> * Do we want to check if the Hierarchy/Dim/Element exists? That seems like extra overhead. TI allows invalid elements, hierarchies, and dims. Returning False if objects don't exist. Do these checks match other methods in TM1py?

in the proposed code I did as few additional checks as possible, while ensuring the 3 approaches have the same behaviour.
- In the `TM1DrillDownMember` case no additional checks are done. 
- In the `TI` case one additional check is done only if the TI function returns False.
- In the `Descendants` case, one additional check is done only if the Element exists check is False.

If we don't do these checks the functions would behave differently in the scenarios of Element doesn't exist or Dimension doesn't exist etc.

> * Are there risks to using TI debugging like this? While it may be faster It may have unintended locking consequences and could carry risks if the script drops its connection and the server is left debugging a TI.

Yes. I would say there are risks associated with using TI debugging like this. I will do more heavy testing before merging it. While potentially risky, the approach can be quite rewarding in scenarios with large dimensions.
With the dimension you posted in #758 the `ti` approach is 10x faster than the `TM1DrillDownMember` approach.
Now that you say it like this I tend to think the arguments against using debugging in production are quite strong.

MariusWirtz:
OK. Instead of debugging we can call the `ELISANC` function in an unbound TI and terminate with Either ProcessQuit or normally depending on whether the `ELISANC` result is 1 or 0.

I will push another commit. Then we get the performance of TI without the drawbacks of using debugging.

rclapp:
Looks better now

MariusWirtz:
> Looks better now

Thanks. Using it as the default method now if the user has admin permissions.

rclapp:
ARC Console also supports this, but in a slightly different way.

POST  ExecuteProcessWithReturn?$expand=ErrorLogFile

Payload: {"Process":
{"Name":"TIConsole",
"PrologProcedure":"function = ELISANC('Big Dim', 'Div by 33', '3');
itemreject(expand('%function%'));"
}}

GET  ErrorLogFiles('TM1ProcessError_20220727153020_36212908_.log')/Content

Error: Prolog procedure line (1):      0.000

I would bet the log file way would slow it down a lot but I figured I would share

Issue #764, opened 2022-07-18
Add Multiple Break Points


ianboltz:
**Describe what did you try to do with TM1py**
I am using add break point on a process to debug, but I would like to add multiple in a single call

**Describe what's not working the way you expect**
Didn't get the expected result? Describe:
1. Add break point is singular, I want to be able to pass a range, or list of points to add in a single call

adscheevel:
That should be feasible, in the meantime you can nest the add breakpoint call in a FOR loop that iterates through a list of breakpoint objects.

MariusWirtz:
Fixed with #791 and #800

PR #791, opened 2022-08-19
Breakpoint refactor & new functions

adscheevel:
I updated the ProcessDebugBreakpoint object and debug_add_breakpoint function to be able to add multiple breakpoints as requested in issue:https://github.com/cubewise-code/tm1py/issues/764.

I also added some new functions for getting current breakpoint details like procedure and current line number as well as new step_in function to complement the step_over, step_out, and continue functions

I shamefully did **not** update any tests around the breakpoint.


MariusWirtz:
Thanks for the changes @adscheevel

I will merge and make a few small enhancements. I think I would create a separate function to add multiple breakpoints `debug_add_breakpoints` and I will try to add some test cases.

PR #792, opened 2022-08-19
Element functions

adscheevel:
I've added new functions to element service for string and numeric elements. I've also added a `get_parents_all_elements` function which is similar to `get_parents` but instead of taking an element argument, will return parents for all elements in dim.

I've taken a stab at the tests for these new functions. I wrapped all of the string element testing in one function because I add a string function to test and then immediately delete. This seemed better than adding it when the class is created, seemed like it could mess up some other tests.

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -112,6 +112,21 @@ class ElementService(ObjectService):
         response = self._rest.GET(url, **kwargs)
         return [e["Name"] for e in response.json()['value']]
 
+    def get_numeric_elements(self, dimension_name: str, hierarchy_name: str, **kwargs) -> List[Element]:
+        url = format_url(
+            "/api/v1/Dimensions('{}')/Hierarchies('{}')/Elements?$expand=*&$filter=Type eq 1",
+            dimension_name,
+            hierarchy_name)
+        response = self._rest.GET(url, **kwargs)
+        return [Element.from_dict(element) for element in response.json()["value"]] 
+
+    def get_numeric_element_names(self, dimension_name: str, hierarchy_name: str, **kwargs) -> List[str]:
+        url = format_url("/api/v1/Dimensions('{}')/Hierarchies('{}')/Elements?$select=Name&$filter=Type eq 1",
+                         dimension_name,
+                         hierarchy_name)
+        response = self._rest.GET(url, **kwargs)
+        return [e["Name"] for e in response.json()['value']]
+
     def get_element_names(self, dimension_name: str, hierarchy_name: str, **kwargs) -> List[str]:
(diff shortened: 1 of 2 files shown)


TM1py/Services/ElementService.py
```diff
@@ -652,6 +713,15 @@ def get_parents(self, dimension_name: str, hierarchy_name: str, element_name: st
 
         return [record["Name"] for record in response.json()["value"]]
 
+    def get_parents_all_elements(self, dimension_name: str, hierarchy_name: str, **kwargs) -> List[str]:
```
MariusWirtz, line 716: the type hint for the return value should be Dict[str, str] I think

TM1py/Services/ElementService.py
```diff
@@ -135,6 +180,22 @@ def get_number_of_leaf_elements(self, dimension_name: str, hierarchy_name: str,
...
+            hierarchy_name)
+        response = self._rest.GET(url, **kwargs)
+        return int(response.text)
+
+    def get_number_of_string_elements(self, dimension_name: str, hierarchy_name: str, **kwargs) -> int:
+        url = format_url(
+            "/api/v1/Dimensions('{}')/Hierarchies('{}')/Elements/$count?$filter=Type eq 2",
+            dimension_name,
+            hierarchy_name)
+        response = self._rest.GET(url, **kwargs)
```
MariusWirtz, line 196: Did you verify that the response contains only the number and no actual elements in the JSON?

TM1py/Services/ElementService.py
MariusWirtz: Nice function. Definitely useful. 

Let's call it `get_parents_of_all_elements`. Then it's more in line with existing functions like `get_number_of_elements`, `get_attribute_of_elements`

TM1py/Services/ElementService.py
adscheevel: Yep, verified only number returned in response. This is copy of the `get_number_of_leaf_elements` function with the filter value being modified.

TM1py/Services/ElementService.py
adscheevel: Thanks for the pointer on the return value, fixed with https://github.com/cubewise-code/tm1py/pull/792/commits/6d66c211c833443f9d3f3500e986716e31b6242a

Good call, function renamed with https://github.com/cubewise-code/tm1py/pull/792/commits/6d66c211c833443f9d3f3500e986716e31b6242a and test case with https://github.com/cubewise-code/tm1py/pull/792/commits/0fe40b946a19ffadb11b91a29afcce275966f269

MariusWirtz:
Nice work @adscheevel Thanks!

PR #796, opened 2022-08-31
Rename stuff in GitProject GitService

MariusWirtz:
work in progress


MariusWirtz:
@scrumthing @nicolasbisurgi 
What helper functions do you think we need for the `GitProject` class?

When I last saw the TM1Project used in a real environment, I think we were explicitly including all attribute dimensions in the project file with `!`.
So possibly we would want to add a function like `include_all_attribute_dimensions` that queries all attribute dimensions from TM1 and adds a line for each one to the project file?

nicolasbisurgi:
@MariusWirtz 
thanks for starting the conversation! I think you are correct, most of what people would use are the exceptions section. I did create the following functions already:
- Add/remove tasks
- Add/remove pre/post push/pull tasks
- Add/remove ignore items
And I'm working on the add remove deployments. I think that the latter will allow developers to make use of the settings functions (as we know tm1s.cfg can usually vary per environment).
I think the pre/post push/pull tasks is also a good addition as it will allow you to automate the pre-post deployment steps that you want to perform. 
Example:
- TasksA: exports (slice) cube data into flat files
- TaskB: saves data, backs up the entire instance
- TaskC: imports flat files created by TaskA.
- IgnoreSection: includes these files created by TaskA

1. TaskA is added as a pre-push run.
2. TaskB is added as a pre-pull run.
3. TaskC is added as post-pull run.

If we include deployments too (as I think we should) it will allow to customize these executions per environment (as deploying in a DEV environment is not the same than in PROD one).

Thoughts?

MariusWirtz:
Hi @nicolasbisurgi,

let's bring our efforts together. 

I will merge this PR. Please open a new PR into `master` for your additions.

TM1py/Objects/GitProject.py
```diff
@@ -1,42 +1,116 @@
...
+            objects: Optional[Dict] = None,
+            ignore: Optional[List] = None,
+            files: Optional[List] = None,
+            deployment: Optional[Dict] = None,
+            pre_push: Optional[List] = None,
+            post_push: Optional[List] = None,
+            pre_pull: Optional[List] = None,
+            post_pull: Optional[List] = None,
+            dependencies: Optional[List] = None,
+            preconditions: Optional[str] = ''):
```
nicolasbisurgi, line 112: removed dependencies and preconditions as they are not a property of the Tm1project but to either objects and tasks

TM1py/Objects/GitProject.py
```diff
@@ -60,13 +134,110 @@ def __init__(self,
...
         self._dependencies = dependencies
         self._preconditions = preconditions
 
+    def add_task(self, project_task: TM1ProjectTask):
+        if self._tasks is None:
+            self._tasks = dict()
+
+        if project_task.task_name in self._tasks:
+            raise ValueError(f"Task with name '{project_task.task_name}' already exists in TM1 project. "
+                             f"Task name must be unique")
```
nicolasbisurgi, line 150: incldued a removal option

TM1py/Objects/GitProject.py
```diff
@@ -60,13 +134,110 @@ def __init__(self,
...
+        Example of the `ignore` property in the tm1project:
+            Exclude all the new Cubes and Views in the source, except Cube_A;
+            include control Process }Drill_Drill_A;
+            and exclude all the new Dimensions which has a name starting with 'Dim'
+
+            "Ignore":
+            [
+              "Cubes/Views",
+              "!Cubes('Cube_A')",
+              "!Processes('}Drill_Drill_A')",
```
nicolasbisurgi, line 223: incldued a removal option

TM1py/Objects/GitProject.py
```diff
@@ -60,13 +134,110 @@ def __init__(self,
...
+              "Dimensions('Dim*')"
+            ]
+        """
+
```
nicolasbisurgi, line 227: ability to add a specific deployment to tm1project

TM1py/Objects/GitProject.py
```diff
@@ -133,72 +327,72 @@ def settings(self) -> dict:
...
 
     @property
-    def tasks(self) -> dict:
+    def tasks(self) -> Dict:
         return self._tasks
 
     @tasks.setter
-    def tasks(self, value: dict):
+    def tasks(self, value: List):
         self._tasks = value
```
nicolasbisurgi, line 339: added the TM1ProjectDeployment as a subclass of the tm1project.

TM1py/Objects/GitProject.py
```diff
@@ -1,42 +1,116 @@
...
+        The property specifies an array of URIs of tasks or objects,
+        which will be executed or loaded, respectively, before executing the current task.
+        E.g.: ["Cubes('Cube_A')", "Dimensions('Dimension_C')"]
+
+        A Task MAY have a Precondition property.
+        The server only executes a Task when either the precondition is not specified, or it is evaluated to TRUE.
+
+        The server only executes a Task one time during a deployment.
+        """
+        if precondition:
```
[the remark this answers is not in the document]
MariusWirtz: Thanks. I agree. We would need helper functions to remove from "tasks" and "ignore" as well.

For the preconditions, I'm not sure. I guess we could allow simple string values if they were passed.

How do you imagine people use the TM1ProjectDeployment class. Can you post a code sample?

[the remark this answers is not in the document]
MariusWirtz: I found the deployment in the IBM docs. Yeah this is cool. It makes sense to implement it as a subclass of TM1Project


> Deployment
> Deployment is an OPTIONAL property. Each of its property defines a named deployment and its specific properties.
> 
> All the tm1project properties can be redefined for a deployment, except Version. Those properties overrides the tm1project properties for the specific deployment. Current deployment is set by action GitInit.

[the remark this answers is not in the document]
MariusWirtz: I agree.

I will change the TM1ProjectTask to allow for preconditions.

Are you saying that preconditions can also be defined on a project level? I couldn't find anything about this in the docs.

nicolasbisurgi:
Hi Marius,
	Sorry for the confusion. After I wrote the comment I realized that it wouldn’t make sense for those to be in the project on there own, that’s why I removed them from the project in the files I shared with you.
We can include them as just strings in the Task class and leave it alone for now.

MariusWirtz:
> Sorry for the confusion. After I wrote the comment I realized that it wouldn’t make sense for those to be in the project on there own, that’s why I removed them from the project in the files I shared with you.
We can include them as just strings in the Task class and leave it alone for now.

I understand. Thanks!

nicolasbisurgi:
Yep, this is actually the only reason why I would mess with tm1project. I think this will be particularly useful for re-defining both ‘Settings' (Performance Monitor, TIDebugger, etc.) and ‘PrePull’,’PostPull’ (savedataall, kickoff users, backup the instance, etc)  per environment.

TM1py/Objects/GitProject.py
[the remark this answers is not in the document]
nicolasbisurgi: > I found the deployment in the IBM docs. Yeah this is cool. It makes sense to implement it as a subclass of TM1Project
> 
> > Deployment
> > Deployment is an OPTIONAL property. Each of its property defines a named deployment and its specific properties.
> > All the tm1project properties can be redefined for a deployment, except Version. Those properties overrides the tm1project properties for the specific deployment. Current deployment is set by action GitInit.

Yep, this is actually the only reason why I would mess with tm1project. I think this will be particularly useful for re-defining both ‘Settings' (Performance Monitor, TIDebugger, etc.) and ‘PrePull’,’PostPull’ (savedataall, kickoff users, backup the instance, etc)  per environment.

PR #801, opened 2022-09-19
Add alternative separators to get_value function

tobiaskapser:
Hi together,

I recently had some problems while trying to read single values from a cube with the get_value function.
This was due to some elements having commas in their names.
While this should normally not be case if elements are named properly, it could happen especially if some aliases are added from other datasources. The get_value function in the CellService is not able to handle that at the moment.

It would be nice to change the default delimiter setting if needed. I propose a change in this pull request but it would have to be added also to some other functions in the CellService (e.g _compose_odata_tuple_from_string).

Best regards
Tobias


MariusWirtz:
Thank you for the contribution @tobiaskapser.

I agree it would be nice to keep the `get_value` function consistent with other functions that use the concept of an `element_string`.
Feel free to open a separate PR with that change. I will merge this one.

PR #805, opened 2022-09-21
Add alternative separators for element strings of the remaining functions in CellService

tobiaskapser:
Hey,

In addition to #801 I added the alternative separators for the remaining functions to keep the interface consistent.

The rule and feeder checking functions have the possibility to pass an Iterator instead of a string. That could also make sense for the ```get_value``` function and would make string splitting obsolete. Maybe we could pass some kind of data structure that allows for the definition of multiple hierarchies too.

Best regards
Tobias

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -244,17 +244,20 @@ class CellService(ObjectService):
     def _compose_odata_tuple_from_string(self, cube_name: str,
                                          element_string: str,
                                          dimensions: Iterable[str] = None,
+                                         element_separator: str = ",", 
+                                         hierarchy_separator: str = "&&", 
+                                         hierarchy_element_separator: str = "::", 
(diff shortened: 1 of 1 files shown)


MariusWirtz:
Thanks for the PR @tobiaskapser. I will merge it as it is now.

I agree it would be nice if the function could consume an `Iterable` of strings.
In that case, if a string is passed it would treat it as the special element string.
If any other type of iterable is passed the function could treat it accordingly.

As you already pointed out, a simple iterable couldn't be used to describe cells on multiple alternate hierarchies.
I wonder if perhaps we could accept an iterable of `mdxpy.Member` 

``` python
[
  Member.of('d1', 'd1', 'e1'),
  Member.of('d2', 'h1', 'c4'),
  Member.of('d2', 'h2', 'c8'),
]

```

PR #807, opened 2022-09-23
Enable TCP keepalive option for executing long-run requests

macsir:
Hi, @MariusWirtz 
I have tested four methods and they look good to me. 
The usage of this parameter is similar to the async one. If both are specified, then async_requests_mode is used by default.
`with TM1Service(address, port, ssl, user, password, namespace, async_requests_mode=True, tcp_keepalive=True) as tm1:`
As you can see, the TCP keepalive method captures the response a bit faster (<1s) than the async mode and since it is operating at TCP layer level, it would be more efficient (no polling and no more connections) than the async method at HTTP layer level, I suppose. 
image
I don't have other cloud environments to test. If someone can confirm it in other environments, it would be good. Thanks.

### DIFF
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -126,6 +126,14 @@ class RestService:
                'Accept': 'application/json;odata.metadata=none,text/plain',
                'TM1-SessionContext': 'TM1py'}
 
+    # You can reset the following TCP socket options based on your own use cases when tcp_keepalive is eanbled
+    # TCP_KEEPIDLE: Time in seconds until the first keepalive is sent
+    # TCP_KEEPINTVL: How often should the keepalive packet be sent
+    # TCP_KEEPCNT: The max number of keepalive packets to send
+    TCP_SOCKET_OPTIONS = {'TCP_KEEPIDLE': 30,
+                          'TCP_KEEPINTVL': 15,
+                          'TCP_KEEPCNT': 60}
+
     def __init__(self, **kwargs):
         """ Create an instance of RESTService
         :param address: String - address of the TM1 instance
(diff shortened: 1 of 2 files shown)


TM1py/Services/RestService.py
```diff
@@ -217,21 +227,38 @@ def __init__(self, **kwargs):
...
+        if self._tcp_keepalive:
+            # Set the following socket options
+            # SO_KEEPALIVE: set 1 to enable TCP keepalive
+            # TCP_KEEPIDLE: Time in seconds until the first keepalive is sent
+            # TCP_KEEPINTVL: How often should the keepalive packet be sent
+            # TCP_KEEPCNT: The max number of keepalive packets to send
+            socket_options = urllib3.connection.HTTPConnection.default_socket_options + [(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1),
+                                                                                         (socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, 30),
+                                                                                         (socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, 15),
+                                                                                         (socket.IPPROTO_TCP, socket.TCP_KEEPCNT, 60)]
```
MariusWirtz, line 248: Nice work @macsir. Thank you.

One suggestion on the code. Perhaps we could store the default values for `TCP_KEEPIDLE` etc. in a class variable similar to the `HEADERS`.
So that in the rare case that someone wants to change it, they wouldn't need to change the code in the RESTService. 

Kinda like here:
https://github.com/cubewise-code/tm1py/discussions/797#discussioncomment-3533883

macsir: Hi, @MariusWirtz That's a great idea and will implement it soon.

macsir:
Hi, @MariusWirtz  Could you review them again? User now can setup their own values if needed. I have run three test cases with different sets of values and they are all passed. TI finished as expected and no timeout in Lambda functions. Thanks.
`  RestService.TCP_SOCKET_OPTIONS["TCP_KEEPIDLE"] = 45
    RestService.TCP_SOCKET_OPTIONS["TCP_KEEPINTVL"] = 20
    RestService.TCP_SOCKET_OPTIONS["TCP_KEEPCNT"] = 70`

MariusWirtz:
Nice work @macsir 
This looks good. Thanks for the contribution

PR #810, opened 2022-10-07
Additional features - CubeSaveData and DeleteAllPersistentFeeders

Mr-SabyasachiBose:
This branch has two new updates in terms of additional features. 

i) A new function **cube_save_data** is added in CubeService.py to serialize a cube using TM1Py. This requires cube name as a parameter and returns the response code.
ii) A new function **delete_persistent_feeders** is added in ServerService.py to delete all persistent feeders using TM1Py. This does not require any parameter and returns the response code.

Both these functions executes a temporary ti process to execute CubeSaveData and DeleteAllPersistentFeeders functions respectively.


MariusWirtz:
Thanks for this pull request @Mr-SabyasachiBose

Issue #789, opened 2022-08-19
Execute_mdx - Sandbox Dimension


sbethea-cw:
I have run into an issue when transferring data between two servers where tm1s.cfg parameter EnableSandboxDimension is different between the two servers.  If EnableSandboxDimension=T on server1 and =F on server2, then the dimensionality of the cellset is not the same.
When I use execute_mdx on server1, the cellset looks like:
{('Base', 'ele1', 'ele2', 'ele3', 'Value'): 12345}
but server2 has EnableSandboxDimension=F and expects the cellset to look like:
{('ele1', 'ele2', 'ele3', 'Value'): 12345}
TM1 throws an error saying 'Base' is not a valid element, like this:
Error: Prolog procedure line (3): Invalid key: Dimension Name: "Version", Element Name (Key): "Base"

Can a feature be added to execute_mdx to exclude the sandbox dimension if it is turned on?

MariusWirtz:
I agree. We should introduce an optional argument to the `execute_mdx` function. Something like `skip_sandbox_dimension`

sbethea-cw:
For reference, when using `execute_mdx` I can put all dimensions on rows/columns and use `skip_context=True`.  This will leave the implicit sandbox dimension on context and exclude from the resulting cellset.  This seems to work as a workaround.

MariusWirtz:
Hi @sbethea-cw,

can you confirm that fix #812 works for you?

```
pip uninstall tm1py
pip install https://github.com/cubewise-code/tm1py/archive/refs/heads/master.zip

```

sbethea-cw:
Hi @MariusWirtz,

I tested ```execute_mdx``` with new argument ```skip_sandbox_dimension=True``` and it worked!  Thanks!

PR #812, opened 2022-10-11
Add argument skip_sandbox_dimension to execute_mdx 

cubewise-gng:
Add argument skip_sandbox_dimension to execute_mdx function for #789

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -278,7 +278,7 @@ def build_content_from_cellset_dict(
     :return:
     """
     cube_dimensions = [dim['Name'] for dim in raw_cellset_as_dict['Cube']['Dimensions']]
-    if skip_sandbox_dimension:
+    if skip_sandbox_dimension and cube_dimensions[0].lower() == "sandboxes":
         cube_dimensions = cube_dimensions[1:]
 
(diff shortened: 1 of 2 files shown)


TM1py/Utils/Utils.py
```diff
@@ -302,6 +304,10 @@ def build_content_from_cellset_dict(
                 coordinate = extract_unique_names_from_members(axis['Tuples'][tuple_ordinal]['Members'])
                 coordinates.extend(coordinate)
 
+                if skip_sandbox_dimension:
+                    coordinates = [element for element in coordinates if
+                                   "[sandboxes].[sandboxes].[" not in element.lower()]
+
         coordinates = sort_coordinates(cube_dimensions, coordinates, element_unique_names=element_unique_names)
```
MariusWirtz, line 311: Thanks for the feature @cubewise-gng!
This will be helpful.

I am sure it works, but a bit fearful to include another string operation so deep in the loop. This line gets executed for every cell in a result set so potentially 100k times for an MDX execution.

I wonder if there might be a "cheaper" way to drop the sandbox from the coordinates.
If we drop the first entry in the `cube_dimensions` list would that produce the desired outcome in the `sort_coordinates` function?
I would prefer to use some kind of shortcut for that operation

PR #814, opened 2022-10-12
Add option to pass element selection as iterator in get_value function

tobiaskapser:
Hi together,

@MariusWirtz 
We discussed the option to pass element coordinates as Iterators to the ```get_value()```-function. That would remove the need to specify special separators for the element string.
You also had the idea to use the MDXpy ```Member``` object. I tried that and it worked out really nice! :)

It also simplifies the building of the MDX statement, because it provides the ```unique_name``` directly.
This could be used to create the ```Member``` directly from the splitted strings too. After that the MDX query can be build independently of the input type always from the ```Member``` object.
So I did some little restructuring to use the ```Member``` consistently throughout the function.

The ```get_value``` function is now callable without hierarchies like this:

```python
get_value("CubeName", [(Dimension1, Element1), (Dimension2, Element2), (Dimension3, Element3)])
```

or with hierarchies like this:
```python
get_value("CubeName", [(Dimension1, Hierarchy1, Element1), (Dimension1, Hierarchy2, Element2), (Dimension2, Element3)])
```

or directly with the ```mdxpy.Member``` objects

Unfortunately this has the disadvantage that the user has to always provide the dimension name. If not given it is not distinguishable if the next element belongs to the next dimension or to the same dimension but another hierarchy.
Does anyone see an option to determine that?

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -13,7 +13,7 @@ from io import StringIO
 from typing import List, Union, Dict, Iterable, Tuple, Optional
 
 import ijson
-from mdxpy import MdxHierarchySet, MdxBuilder
+from mdxpy import MdxHierarchySet, MdxBuilder, Member
 from requests import Response
 
 from TM1py.Exceptions.Exceptions import TM1pyException, TM1pyWritePartialFailureException, TM1pyWriteFailureException, \
@@ -172,7 +172,7 @@ class CellService(ObjectService):
         """
         super().__init__(tm1_rest)
 
-    def get_value(self, cube_name: str, element_string: str, dimensions: List[str] = None, sandbox_name: str = None,
+    def get_value(self, cube_name: str, element_string: Union[str, Iterable], dimensions: List[str] = None, sandbox_name: str = None,
                   element_separator: str = ",", hierarchy_separator: str = "&&", hierarchy_element_separator: str = "::", 
                   **kwargs) -> Union[str, float]:
         """ Element_String describes the Dimension-Hierarchy-Element arrangement
@@ -183,6 +183,10 @@ class CellService(ObjectService):
             - The , separates the element-selections
             - If more than one hierarchy is selected per dimension && splits the elementselections
             - If no Hierarchy is specified. Default Hierarchy will be addressed
+        or
+        Iterable of type mdxpy.Member or similar
+            - Dimension names must be provided in this case! Example: [(Dimension1, Element1), (Dimension2, Element2), (Dimension3, Element3)]
+            - Hierarchys can be included. Example: [(Dimension1, Hierarchy1, Element1), (Dimension1, Hierarchy2, Element2), (Dimension2, Element3)]
         :param dimensions: List of dimension names in correct order
         :param sandbox_name: str
         :param element_separator: Alternative separator for the element selections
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Hi @tobiaskapser,

thanks! I like it.
It's nice that you added tests with the change.
I think the function is more useful now IMO. 

It's kinda disturbing that the argument is called `element_string` when in reality it accepts a range of types besides strings. I guess we couldn't change this though without breaking backward compatability.

I will merge it as is.

MariusWirtz:
Or maybe we could change it. Like we could introduce a new argument as position 2 and call it `elements`, `members` or `intersection`.

Then we could extract `element_string` from kwargs in case someone is passing the "old" keyword argument to the function.
That would be backward compatible I guess.

Not sure if it's worth it though only to have descriptive argument names...

MariusWirtz:
> Unfortunately this has the disadvantage that the user has to always provide the dimension name. If not given it is not distinguishable if the next element belongs to the next dimension or to the same dimension but another hierarchy. Does anyone see an option to determine that?

I could not think of a way to resolve this

tobiaskapser:
> Or maybe we could change it. Like we could introduce a new argument as position 2 and call it `elements`, `members` or `intersection`.
> 
> Then we could extract `element_string` from kwargs in case someone is passing the "old" keyword argument to the function. That would be backward compatible I guess.
> 
> Not sure if it's worth it though only to have descriptive argument names...

That is a good idea! ```element_string``` really is misleading. I was also thinking about this problem while implementing it but did not came up with a solution.
I would implement this before we merge this. It is not a big change and would improve clarity.

tobiaskapser:
Tested it. There is one problem with this approach:

Suppose the user passed ```element_string``` as a keyword argument, not as a positional argument. Like this:
```python
get_value(cube_name, element_string="Element1,Element2,Element3")
```
This function call would be okay in the current version.


 If we now change the second argument to ```elements``` (still positional) and extract the ```element_string``` from the kwargs the user gets an ```TypeError``` because the positional argument ```elements``` is not passed by the user.
That would break backwards compatibility for all function calls that pass the positional arguments as a keyword argument.

MariusWirtz:
> If we now change the second argument to elements (still positional) and extract the element_string from the kwargs the user gets an TypeError because the positional argument elements is not passed by the user.
That would break backwards compatibility for all function calls that pass the positional arguments as a keyword argument.

Would it work if we set the default value for elements to None then?

TM1py/Services/CellService.py
```diff
@@ -191,50 +195,47 @@ def get_value(self, cube_name: str, element_string: str, dimensions: List[str] =
...
-                    hierarchy_name, element_name = element_selection.split(hierarchy_element_separator)
-                else:
-                    hierarchy_name = dimension_name
-                    element_name = element_selection
+        mdx_strings_list = []
 
-                mdx_rows_list.append("{[" + dimension_name + "].[" + hierarchy_name + "].[" + element_name + "]}")
+        # Keep backwards compatibility with the earlier used "element_string" parameter
+        if elements is None and "element_string" in kwargs:
+            elements = kwargs.get("element_string")
```
MariusWirtz, line 202: Just one small change request: Please use `kwargs.pop` here so that the element isn't delegated further down.

tobiaskapser:
That seems to work! I added the commits to my fork.

TM1py/Services/CellService.py
tobiaskapser: Thanks for the input, Marius!
I changed it to ```kwargs.pop```.

Did some tiny changes on the comments while I had the chance 😄

MariusWirtz: Nice work. Thanks!

PR #823, opened 2022-11-02
Update CellService.py

meyersrl:
updated the skip header code for execute_mdx_elements_value_dict to include a default return of None. If using skip_rule_derived_cells you would potentially encounter an error if there were no records returned by the mdx. Essentially this error handles that problem by returning None.


MariusWirtz:
Good catch! Thank you

PR #831, opened 2022-11-16
adding lamda expression to filter tuple

ldelberg:
It's possible to filter the transactionlog entries with a lambda expression: Tuple/any(t: t eq 'filter value')
The function get_transaction_log_entries accepts kwargs so we could  add a section to the log_filter list in order to generate a lambda expression for the query:
` get_transaction_log_entries(reverse=True, user='Admin', cube='plan_BudgetPlan', tuple={'105': 'eq', '405': 'eq', '41101': 'ge'})`

tuple is a dict with 'filter value' as key and a operator as value. 

This could be extended to a lambda all expression by defining a lambda_method parameter which defaults to "any" but can changed to "all".
What do you think?

### DIFF
--- a/TM1py/Services/ServerService.py
+++ b/TM1py/Services/ServerService.py
@@ -203,16 +203,16 @@ class ServerService(ObjectService):
         return timestamp_utc
 
     @require_admin
-    def get_transaction_log_entries(self, reverse: bool = True, user: str = None, cube: str = None,
+    def get_transaction_log_entries(self, reverse: bool = True, user: str = None, cube: str = None, elements: Dict = None,
                                     since: datetime = None, until: datetime = None, top: int = None, **kwargs) -> Dict:
         """
         :param reverse: Boolean
         :param user: UserName
         :param cube: CubeName
+        :param elements: of type dict. Filtervalue as key and comparison operator as value tuple={'Filtervalue1':'eq','Filtervalue2': 'ge'}
         :param since: of type datetime. If it doesn't have tz information, UTC is assumed.
         :param until: of type datetime. If it doesn't have tz information, UTC is assumed.
         :param top: int
-        :param tuple: of type dict. Filtervalue as key and comparison operator as value tuple={'Filtervalue1':'eq','Filtervalue2': 'ge'}
         :return:
         """
         reverse = 'desc' if reverse else 'asc'
@@ -225,9 +225,9 @@ class ServerService(ObjectService):
                 log_filters.append(format_url("User eq '{}'", user))
             if cube:
                 log_filters.append(format_url("Cube eq '{}'", cube))
-            if kwargs.get('tuple', None):
+            if elements:
                 log_filters.append(format_url(
-                    "Tuple/any(t: {})".format(" or ".join([f"t {v} '{k}'" for k, v in kwargs.get('tuple').items()]))))
+                    "Tuple/any(t: {})".format(" or ".join([f"t {v} '{k}'" for k, v in elements.items()]))))
             if since:
                 # If since doesn't have tz information, UTC is assumed
                 if not since.tzinfo:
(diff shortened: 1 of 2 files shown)


zsoltmoravcsik:
That is a great idea. +1

TM1py/Services/ServerService.py
```diff
@@ -194,27 +212,34 @@ def get_transaction_log_entries(self, reverse: bool = True, user: str = None, cu
         :param since: of type datetime. If it doesn't have tz information, UTC is assumed.
         :param until: of type datetime. If it doesn't have tz information, UTC is assumed.
         :param top: int
+        :param tuple: of type dict. Filtervalue as key and comparison operator as value tuple={'Filtervalue1':'eq','Filtervalue2': 'ge'}
```
MariusWirtz, line 215: I would suggest calling this parameter `elements` instead of `tuple`. That would be more in line with the existing arguments.
I think there is no harm in including it in the function signature if we set the default value to `None`.
That should make it easier to find and use for people

I can't really think of a good use case for `gt` or `le` for element names (maybe years?). I imagine `eq` and `ne` would be helpful. 
The user might want to choose whether `any` or `all` is used in the filtering. So perhaps we could introduce another argument like `elements_filter_type` or `elements_any_or_all`

Maybe for both values (`gt`/`eq`/... and `any`/`all`), we could offer an enum but also accept a str kinda like we did for element types.
https://github.com/cubewise-code/tm1py/blob/2d45a621e4de9794c239dc2958a6fcad94ed464a/TM1py/Objects/Element.py#L34

MariusWirtz:
Nice work @ldelberg! Thanks

I added a suggestion regarding naming and parametrization. 

If you have time you can also add a test case here: https://github.com/cubewise-code/tm1py/blob/2d45a621e4de9794c239dc2958a6fcad94ed464a/Tests/ServerService_test.py#L151

Otherwise, I would write a test as part of a review next week or so.

MariusWirtz:
@adscheevel adding you to this conversation since you wrote the previous version of this function. 
I'm curious about your thoughts on the feature.

TM1py/Services/ServerService.py
ldelberg: Thanks @MariusWirtz, I'll add `elements: dict = None`  and `elements_any_or_all: str = 'any'` as parameter and update the log_filter section.
For the comparison operators I've in mind to filter ranges of i.e. accounts with `ge`/`gt`...
so I would stay with the dict of `'filter value':'operator'`

ldelberg: I think we don't need the `Tuple/all` expression. This expression is only true if all elements in the collection meet the condition. In this case we have a list with elements and therefore `all` can't be true:
```
Tuple['e1','e2','e3','e4','e5']
Tuple/all(t: t eq 'e1') # always false
```
In the future we'll have a better way to filter. Hubert announced tm1 will support a odata v4.01 feature to request a member of an ordered collection.
With this feature we could use:
`(Tuple/2 eq 'PLAN') # dimension 2 is actual, plan, fc....`

For now I would go with `any` only and update the function when tm1 supports this feature

ldelberg:
@MariusWirtz I've added a test for this part of the function. Please check if it meets your expectations

TM1py/Services/ServerService.py
ldelberg: I know what you're thinking about :) 
Unfortunatly /all works in an other way you would expect.
Let's say we have a Tuple with `[e1,e2]` - now your expectation is to select this Tuple by using this expression:
`Tuple/all(e: e eq 'e1' or e eq 'e2')`
But this expression will only return these Tuples:
`[e1,e1] or [e2,e2]`
because /all means every element in the list have to meet the condition.
With the announced feature for a future tm1 version could you select the Tuple with this expression:
`(Tuple/1 eq 'e1') and (Tuple/2 eq 'e2') `
With this feature we're able to refactor the function in tm1py ;-)

MariusWirtz: Understood. `all` is not helpful in thise case.

PR #834, opened 2022-11-17
Order args in `get_transaction_log_entries`

MariusWirtz:
@ldelberg please approve.

For backward compatibility, I moved the elements as the last argument in the function.

I wonder how we should support position-based filtering in the function once the REST API allows for it.
I am not sure if we could manage it through the same argument. Maybe we introduce two arguments:
- `element_tuple_filter`
- `element_position_filter`

```
element_tuple_filter={'2020': 'gt', 'Actual': 'eq', 'Revenue': 'eq'}
```

```
element_position_filter={
0: {'2020': 'gt'},
2: {'Actual': 'eq', 'Budget': 'eq'},
7: {'Revenue': 'eq'}
}
```
What do you think?

### DIFF
--- a/TM1py/Services/ServerService.py
+++ b/TM1py/Services/ServerService.py
@@ -8,8 +8,8 @@ from typing import Dict, Optional
 
 import pytz
 from requests import Response
-from TM1py.Objects.Process import Process
 
+from TM1py.Objects.Process import Process
 from TM1py.Services.ObjectService import ObjectService
 from TM1py.Services.RestService import RestService
 from TM1py.Utils import format_url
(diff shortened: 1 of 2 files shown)


TM1py/Services/ServerService.py
```diff
@@ -205,30 +205,36 @@ def utc_localize_time(timestamp):
...
         :return:
         """
+        if element_position_filter:
+            raise NotImplementedError("Feature expected in upcoming releases of TM1, TM1py")
+
         reverse = 'desc' if reverse else 'asc'
         url = '/api/v1/TransactionLogEntries?$orderby=TimeStamp {} '.format(reverse)
+
         # filter on user, cube and time
         if user or cube or since or until:
```
ldelberg, line 229: I think `or element_tuple_filter or element_position_filter` should be added here

MariusWirtz: good catch. Thank you

rclapp:
In a future version we might think about making TM1Py more ODATA like and less TI like. It would bring it closer to MDXPy and how it handles function chaining.
You can see a sample here: https://pyodata.readthedocs.io/en/latest/usage/querying.html#get-entities-matching-a-filter:~:text=Get%20entities%20matching%20a%20filter%20in%20more%20Pythonic%20way%C2%B6

We could create methods that would extend TM1 End points. So it might look like this

t_log_entries = tm1.server.get_transaction_log_entries().filter(…)

From: ldelberg ***@***.***>
Date: Thursday, November 17, 2022 at 10:03 AM
Cc: Subscribed ***@***.***>
Subject: Re: [cubewise-code/tm1py] Order args in `get_transaction_log_entries` (PR #834)


Hi @MariusWirtz<https://github.com/MariusWirtz> , introducing these two parameters is a good approach. I like the element_position_filter for future use.
Implementing the log_filter adjustments is a lean thing when you can derive the filter statements by parameter.

MariusWirtz:
@ldelberg 
Thanks will merge as is now.

@rclapp 
Yeah that would be so powerful. Sounds like a big architectural change IMO. Possibly something TM1py 2.0.

rclapp:
Agree, something for 2.0.

PR #837, opened 2022-11-18
adds `element_position_filter` to `get_transaction_log_entries`

ldelberg:
@MariusWirtz, 
I think I found a way to make use this new parameter. This enhancement of the function allows to filter by Elements in the tuple for multiple sections.
Let's say the message log tuple is `[d1e1,d2e1,d3e4,d4e2]` then you could select this Tuple with the parameter 
```
element_postion_filter={
1: {'d1e1': 'eq'},
2: {'d2e1': 'eq'},
3: {'d3e4': 'eq'},
4: {'d4e2': 'eq'}
}
```
What do you think?

### DIFF
--- a/TM1py/Services/ServerService.py
+++ b/TM1py/Services/ServerService.py
@@ -4,7 +4,7 @@ import functools
 import json
 from collections.abc import Iterable
 from datetime import datetime
-from typing import Dict, Optional
+from typing import Dict, List, Optional
 
 import pytz
 from requests import Response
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Hi @ldelberg,
thanks! I like it. But I can't confirm that it works as one would assume. See below

image

ldelberg:
@MariusWirtz that' right. At the moment it' would only be possible to filter a full qualified set:
```element_position_filter={1:{'e1':'eq'},2:{'e2':'eq'}}```
leads to this results:
```['e1','e2'] or ['e2','e1]```
You're right. This ist not the expecting result when I use a positional parameter. This will be possible when Hubert has build this new function. 
For my actual use case it would work as I have unique elements in my dimensions but it's not the best workaround.

MariusWirtz:
I see. I suggest we leave this branch on hold until Hubert has the new feature released

MariusWirtz:
closed for the moment

PR #844, opened 2023-01-06
Sandbox updates

adscheevel:
Added load/unload sandbox functions. Also added Active, Loaded, Queued properties to sandbox object. I did not change the sandbox body, so all existing tests should still pass. I've taken a stab at adding tests for the new functions and properties.


MariusWirtz:
Thank you @adscheevel

I will review and comment this week!

MariusWirtz:
Thank you. 

I will merge and a small commit to it.

Issue #855, opened 2023-01-27
Add boolean `re_authenticate` argument to `TM1Service`


MariusWirtz:
In cases where a connection is idle for a while _(e.g. TM1py waiting for a long rule change to compile)_, it can happen that the TM1py connection is closed on the server side due to the session timeout defined in TM1. In that case, TM1py raises a 401 error on the next function call.

In that case, it would be convenient if TM1py would attempt to re-authenticate (once) with the TM1 Server instead of raising the 401 error.

Potential names for the new argument:
- `re_authenticate`
- `re_authenticate_on_session_timeout`
- `re_authenticate_on_401`

Any additional thoughts or ideas on this feature request?

macsir:
Would it be convenient to allow user to specify max attempts?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Marius Wirtz ***@***.***>
Sent: Friday, January 27, 2023 10:21:56 PM
To: cubewise-code/tm1py ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [cubewise-code/tm1py] Add boolean `re_authenticate` argument to `TM1Service` (Issue #855)


In cases where a connection is idle for a while (e.g. TM1py waiting for a long rule change to compile), it can happen that the TM1py connection is closed on the server side due to the session timeout defined in TM1. In that case, TM1py raises a 401 error on the next function call.

In that case, it would be convenient if TM1py would attempt to re-authenticate (once) with the TM1 Server instead of raising the 401 error.

Potential names for the new argument:

  *   re_authenticate
  *   re_authenticate_on_session_timeout
  *   re_authenticate_on_401

Any additional thoughts or ideas on this feature request?

MariusWirtz:
@macsir Would you put a max_attempts in a way that
- the re-authentication must work on the first attempt (otherwise 401 is raised)?
- or would you say the script can only re-authenticate once and not again 2hours later on another session timeout?

Acteon-Jon:
Would a delay option be useful? re_authenticate_on_401, re_authenticate_delay

macsir:
> @macsir Would you put a max_attempts in a way that
> 
> * the re-authentication must work on the first attempt (otherwise 401 is raised)?
> * or would you say the script can only re-authenticate once and not again 2hours later on another session timeout?

[...]

PR #856, opened 2023-01-31
Reconnect on session timeout

MariusWirtz:
Fixes https://github.com/cubewise-code/tm1py/issues/855

### DIFF
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import functools
 import re
+import socket
 import time
 import warnings
 from base64 import b64encode, b64decode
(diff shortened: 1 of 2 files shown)


macsir:
Hi, @MariusWirtz , I assume this new re_auth method is reusing the same session id?

MariusWirtz:
> I assume this new re_auth method is reusing the same session id?

I imagine you would only call the `re_auth` function when your session is already invalidated.
The below script has a new session_id after the re_auth.

``` python
from TM1py import TM1Service

tm1 = TM1Service(base_url="https://localhost:12354", user="admin", password="apple")
print(tm1.connection.session_id)

# imagine this logout happens naturally after a long wait time through HttpSessionTimeout on TM1 side
tm1.logout()

tm1.re_connect()

print(tm1.connection.session_id)

```

> out:

fHA5dgLnIGAhRh-B3Nktdpl1IqA
BLWCG9y58JK2_XsVBkteLZp1IqA

PR #881, opened 2023-03-27
Add use_blob() to write() function to increase performance

### LINKED ISSUE #848: Impement `use_blob` argument in `write` and `write_dataframe` function for better write speed
As discussed in https://github.com/cubewise-code/tm1py/issues/774 increasing the number of statements in a TI hasn't increased the write performance for the `use_ti` mode.

The time it takes to create a TI process in TM1 has a linear relationship to the number of lines that are included in the process.
So increasing the numbers doesn't help.

To accomplish parity between TI write speed and TM1py write speed an alternative solution is more promising: `use_blob`.

- uploads data as a csv to the TM1 server (using `ApplicationService`). 
- inject a small bedrock-like process into the TM1 server to load the csv to the cube.
- call the bedrock-like process
- Delete the process

One needs to consider that this approach should work seamlessly with the already existing functions: `write_async` and `write_dataframe_async`.

cwffonseca:
Closes #848 

Added `use_blob` parameter to function `write()`, and added underlying function `write_through_blob()` to increase performance of writing to cells. 
This function uploads a CSV file, and then runs a TI process which uses this CSV as data source to load data to a cube.

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -1124,9 +1124,6 @@ class CellService(ObjectService):
         enable_sandbox = self.generate_enable_sandbox_ti(sandbox_name)
         dataload_process.prolog_procedure = '\r' + enable_sandbox
 
-        # Create or update TI process
-        process_service.update_or_create(process=dataload_process)
-
         # Call the TI process with result
         success, status, log_file = process_service.execute_process_with_return(process=dataload_process)
         if not success:
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -172,8 +172,10 @@ class CellService(ObjectService):
         """
         super().__init__(tm1_rest)
 
-    def get_value(self, cube_name: str, elements : Union[str, Iterable] = None, dimensions: List[str] = None, sandbox_name: str = None,
-                  element_separator: str = ",", hierarchy_separator: str = "&&", hierarchy_element_separator: str = "::",
+    def get_value(self, cube_name: str, elements: Union[str, Iterable] = None, dimensions: List[str] = None,
+                  sandbox_name: str = None,
+                  element_separator: str = ",", hierarchy_separator: str = "&&",
+                  hierarchy_element_separator: str = "::",
                   **kwargs) -> Union[str, float]:
         """ Returns cube value from specified coordinates
 
@@ -216,7 +218,7 @@ class CellService(ObjectService):
                     else:
                         hierarchy_name = dimension_name
                         element_name = element_selection
-                    
+
                     element_definition = Member.of(dimension_name, hierarchy_name, element_name)
                     mdx_strings_list.append("{" + element_definition.unique_name + "}")
 
@@ -246,9 +248,9 @@ class CellService(ObjectService):
     def _compose_odata_tuple_from_string(self, cube_name: str,
                                          element_string: str,
                                          dimensions: Iterable[str] = None,
-                                         element_separator: str = ",", 
-                                         hierarchy_separator: str = "&&", 
-                                         hierarchy_element_separator: str = "::", 
+                                         element_separator: str = ",",
+                                         hierarchy_separator: str = "&&",
+                                         hierarchy_element_separator: str = "::",
                                          **kwargs) -> OrderedDict:
         if not dimensions:
             dimensions = self.get_dimension_names_for_writing(cube_name=cube_name)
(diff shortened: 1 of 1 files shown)


MariusWirtz:
Wow, this is great! Thank you @cwffonseca 

I will immediately get to work to review!

TM1py/Services/CellService.py
```diff
@@ -991,6 +1008,138 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+        if not application_service.exists(path='', application_type='FOLDER', name=tm1py_folder.name):
+            application_service.create(tm1py_folder)
+        # Update or create the CSV file inside the TM1py folder
+        application_service.update_or_create_document_from_file(path=tm1py_folder.name, name=export_filename,
+                                                                path_to_file=export_path_to_file)
+        # Retrieve the file ID to be used as data source in the TI process. Default folder is .\Externals
+        blob_filename = application_service.get_document(path=tm1py_folder.name, name=export_filename).file_id
+
+        # Create a TI processto load CSV  data into the cube
+        dataload_process_name = 'cube.' + cube_name + '.load.fromfile.' + random_key + 'test'
```
rclapp, line 1061: Let's remove test from the end of process name, but maybe add the username.

TM1py/Services/CellService.py
```diff
@@ -991,6 +1008,138 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+        # Generate a random key to name uniquely the data extract file
+        random_key = random.randint(1000, 9999).__str__()
+        # Transform the cellset into a dictionary and export to CSV
+        df = build_pandas_dataframe_from_cellset(cellset_as_dict, multiindex=False)
+        # Define the path and file name
+        export_path = '.\\'
+        export_filename = 'data-source-extract-' + random_key + '.csv'
+        export_path_to_file = export_path + export_filename
+        # export cell dataframe to csv without indexes adn with all values with doublequotes
+        df.to_csv(export_path_to_file, index=False, sep=',', quoting=csv.QUOTE_ALL)
```
rclapp, line 1047: Is there a way to write directly from memory to an application folder skipping the local on disk copy?

TM1py/Services/CellService.py
```diff
@@ -991,6 +1008,138 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+
+        # Create variables, all String
+        variable_prefix = 'v'
+        variable_cube_measure = variable_prefix + cube_measure
+        variable_list = []
+        for n, dimension in enumerate(cube_dimensions):
+            variable_list.append(variable_prefix + dimension.replace(' ', ''))
+            dataload_process.add_variable(name=variable_list[n], variable_type='String')
+
+        df = build_pandas_dataframe_from_cellset(cellset_as_dict)
```
rclapp, line 1084: Is this a duplicate of line 1041?

TM1py/Services/CellService.py
```diff
@@ -991,6 +1008,138 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+            application_service.create(tm1py_folder)
+        # Update or create the CSV file inside the TM1py folder
+        application_service.update_or_create_document_from_file(path=tm1py_folder.name, name=export_filename,
+                                                                path_to_file=export_path_to_file)
+        # Retrieve the file ID to be used as data source in the TI process. Default folder is .\Externals
+        blob_filename = application_service.get_document(path=tm1py_folder.name, name=export_filename).file_id
+
+        # Create a TI processto load CSV  data into the cube
+        dataload_process_name = 'cube.' + cube_name + '.load.fromfile.' + random_key + 'test'
+        dataload_process = Process(name=dataload_process_name,
```
rclapp, line 1062: Let's move the process creation to its own method. Perhaps a utility like "create_csv_load_process"

TM1py/Services/CellService.py
```diff
@@ -991,6 +1008,138 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+            else:
+                raise TM1pyWriteFailureException(status, log_file)
+
+        # Delete CSV file on TM1 server
+        application_service.delete(path='TM1py', application_type='DOCUMENT', application_name=export_filename)
+        # Optinally delete the folder TM1py
+        # application_service.delete(path='', application_type='FOLDER', application_name='TM1py')
+
+        # Delete CSV file local
+        os.remove(export_path_to_file)
```
rclapp, line 1141: We should also add a parameter to clean up the applications folder

MariusWirtz:
Thanks for the review @rclapp.
I'm on it :)

TM1py/Services/CellService.py
MariusWirtz: resolved in #882

TM1py/Services/CellService.py
MariusWirtz: Yes. I used the StringIO class in #882 to "mock" a file.

TM1py/Services/CellService.py
MariusWirtz: Let's discuss this in #882
Yeah. I think we could make this optional. If the function is called in parallel though, it would be a nuisance if a separate thread, process, or application removed the tm1py folder in though...

MariusWirtz:
closed in favor of #882

Issue #774, opened 2022-07-27
Drop `MAX_STATEMENTS` in `write` `use_ti` logic in versions > 11.8.15 


MariusWirtz:
I was just informed that TM1 release 11.8.15 will not have a limitation on the number of statements that a TI process can have.

We can increase the MAX_STATEMENTS logic in the `write_through_unbound_process` function. This should lead to better performance as the CellPut statements won't need to be broken down into n sequentially executed processes.

MariusWirtz:
Here is a sample reproducing the issue

``` python
from TM1py import TM1Service, Dimension, Hierarchy, Cube, Process

CUBE_NAME = "TM1py Cube"
DIMENSION_NAMES = "TM1py Demension 1", "TM1py Dimension 2"
PROCESS_NAME = "TM1py Process"

CELL_UPDATES = 250_000


def prep(tm1):
    hierarchy = Hierarchy(DIMENSION_NAMES[0], DIMENSION_NAMES[0])
    hierarchy.add_element("Total", "Consolidated")
    for element_number in range(CELL_UPDATES):
        hierarchy.add_component(parent_name="Total", component_name=f"Element {str(element_number)}", weight=1)
    dimension = Dimension(DIMENSION_NAMES[0], hierarchies=[hierarchy])
    tm1.dimensions.update_or_create(dimension)

    hierarchy = Hierarchy(DIMENSION_NAMES[1], DIMENSION_NAMES[1])
    hierarchy.add_element("Element 1", "Numeric")
    dimension = Dimension(DIMENSION_NAMES[1], hierarchies=[hierarchy])
    tm1.dimensions.update_or_create(dimension)

    cube = Cube(CUBE_NAME, dimensions=DIMENSION_NAMES)
    tm1.cubes.update_or_create(cube)


def create_and_run(tm1):
    process = Process(
        name="TM1py Process",
        prolog_procedure="\r".join([
            f"CellPutN(1, '{CUBE_NAME}', 'Element {e}', 'Element 1');"
            for e
            in range(CELL_UPDATES)]))

    tm1.processes.update_or_create(process)

    success, status, error_log_file = tm1.processes.execute_with_return(process.name)

    if not success:
        raise RuntimeError("Failed to execute process")


def run_unbound(tm1):
    process = Process(
        name="",
        prolog_procedure="\r".join([
            f"CellPutN(1, '{CUBE_NAME}', 'Element {e}', 'Element 1');"
            for e
[...]

PR #882, opened 2023-03-27
Feature/write with blob

MariusWirtz:
The successor of #881 

Changes:
- use private application
- don't create an actual CSV file on the disk
- name application and process with hash based on tm1 session + local thread
- break process creation into a separate function


MariusWirtz:
Things to consider before the merge
- Gather stats on the performance of. raw TI vs `use_ti` vs `use_blob`. `use_blob` and raw TI should presumably be en par
- Provided that `use_blob` is faster than `use_ti` we should change the `write_async` function to make use of `use_blob` as well
- challenge the uniqueness of process names and application names in multithreading environment

rclapp (approved):
Awesome! Thanks!

rclapp:
With such a big improvement can you push this to PyPi sometime soon?

TM1py/Services/CellService.py
```diff
@@ -991,6 +1011,156 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+            list(elements) + [value.replace('\r', '').replace('\n', '') if isinstance(value, str) else value]
+            for elements, value
+            in cellset_as_dict.items())
+
+        # Define the path and file name
+        filename = f'{unique_hash}.csv'
+
+        # Upload CSV to TM1 server using ApplicationService:
+        # Create a folder TM1py
+        tm1py_folder = FolderApplication(path='', name='TM1py')
```
rkvinoth, line 1056: Should this folder `name` be dynamic?

TM1py/Services/CellService.py
```diff
@@ -991,6 +1011,156 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+                private=True).file_id
+
+            # Create a TI process to load CSV  data into the cube
+            process_name = f"}}tm1py.{unique_hash}"
+            process = self._create_blob_write_process(cube_name, process_name, blob_filename,
+                                                      cube_service.get_dimension_names(cube_name), increment,
+                                                      skip_non_updateable, sandbox_name)
+
+            # Call the TI process with result
+            # process_service.create(process)
```
rkvinoth, line 1081: Delete this comment?

TM1py/Services/CellService.py
```diff
@@ -991,6 +1011,156 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+            success, status, log_file = process_service.execute_process_with_return(process=process)
+            if not success:
+                if status in ['HasMinorErrors']:
+                    raise TM1pyWritePartialFailureException([status], [log_file], 1)
+                else:
+                    raise TM1pyWriteFailureException([status], [log_file])
+
+        # delete application in TM1
+        finally:
+            application_service.delete(
```
rkvinoth, line 1091: Could this be added as a default argument? Could come in handy for debugging.

ajain86:
I am intrigued by seeing such a big difference in use_ti vs use_blob. In my experience, the creation of the previous unbound TI with the CellPutN statements, did not add a significant processing time. Any insight on what might be creating this improvement.

TM1py/Services/CellService.py
```diff
@@ -991,6 +1011,156 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+            datasource_data_source_name_for_server='.\\}Externals\\' + blob_filename,
+            datasource_data_source_name_for_client='.\\}Externals\\' + blob_filename,
+            datasource_ascii_delimiter_char=',',
+            datasource_ascii_decimal_separator='.',
+            datasource_ascii_thousand_separator='',
+            datasource_ascii_quote_character='"')
+        cube_measure = cube_dimensions[-1]
+
+        # Create variables as all String
+        variable_prefix = 'v'
```
rkvinoth, line 1112: why not just hard code instead of creating a variable?

MariusWirtz:
> With such a big improvement can you push this to PyPi sometime soon?

Yes. We should do this short term.

TM1py/Services/CellService.py
MariusWirtz: > I am intrigued by seeing such a big difference in use_ti vs use_blob. In my experience, the creation of the previous unbound TI with the CellPutN statements, did not add a significant processing time. Any insight on what might be creating this improvement.

Here is my hypothesis:
The TM1 server does some kind of "scanning" when a process is created. That operation is expensive in terms of performance and the cost scales linear with the number of lines in the process.
https://github.com/cubewise-code/tm1py/issues/774#issuecomment-1258088290

In the old approach, we would create a process with 1 million lines. Now we always create a process with 15 lines or so. Even if we write 1 million elements.

MariusWirtz:
> I am intrigued by seeing such a big difference in use_ti vs use_blob. In my experience, the creation of the previous unbound TI with the CellPutN statements, did not add a significant processing time. Any insight on what might be creating this improvement.

Here is my hypothesis:
The TM1 server does some kind of "scanning" when a process is created. That operation is expensive in terms of performance and the cost scales linear with the number of lines in the process.
https://github.com/cubewise-code/tm1py/issues/774#issuecomment-1258088290

In the old approach, we would create a process with 1 million lines. Now we always create a process with 15 lines or so. Even if we write 1 million elements.

TM1py/Services/CellService.py
MariusWirtz: You mean whether or not we delete the blob?

rkvinoth: yeah right

MariusWirtz:
Any more feedback is welcome on this one!
Since we release it to PyPI soon we should get the namings and arguments and design decisions right short term.

Once it's out on PyPI we can't really change it much anymore without breaking compatibility

TM1py/Services/CellService.py
```diff
@@ -991,6 +1012,163 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+            list(elements) + [value.replace('\r', '').replace('\n', '') if isinstance(value, str) else value]
+            for elements, value
+            in cellset_as_dict.items())
+
+        # Define the path and file name
+        filename = f'{unique_hash}.csv'
+
+        # Upload CSV to TM1 server using ApplicationService:
+        # Create a folder TM1py
+        tm1py_folder = FolderApplication(path='', name=folder_name)
```
gbryant-dev, line 1060: I'm wondering if uploading to the 'Blobs' folder instead of 'Applications' would be a cleaner approach here? 

The main reason being is that in the `_create_blob_write_process` function, the }Externals folder is being directly referenced. It feels like applications residing in the }Externals folder is an implementation detail of TM1 that TM1py shouldn't care about. If the document was uploaded through the 'Blobs' folder, the id of the blob can be passed directly to the datasource parameters e.g.

`datasource_data_source_name_for_server=blob_filename,
datasource_data_source_name_for_client=blob_filename`

MariusWirtz: I think you have a point there @gbryant-dev.

We should change it to use blobs and ignore the concepts of TM1 applications.

rkvinoth: @MariusWirtz Did you get a chance to test this in a Linux env?

rclapp:
That would just use the contents endpoint, right? Similar to how arc let's you upload a source file.

Sent from my mobile device

On Mar 29, 2023 6:18 PM, Marius Wirtz ***@***.***> wrote:

In TM1py/Services/CellService.py<https://github.com/cubewise-code/tm1py/pull/882#discussion_r1152541007>:

> +        csv_content = StringIO()
+        csv_writer = csv.writer(
+            csv_content,
+            delimiter=",",
+            quoting=csv.QUOTE_ALL)
+        csv_writer.writerows(
+            list(elements) + [value.replace('\r', '').replace('\n', '') if isinstance(value, str) else value]
+            for elements, value
+            in cellset_as_dict.items())
+
+        # Define the path and file name
+        filename = f'{unique_hash}.csv'
+
+        # Upload CSV to TM1 server using ApplicationService:
+        # Create a folder TM1py
+        tm1py_folder = FolderApplication(path='', name=folder_name)


I think you have a point there @gbryant-dev<https://github.com/gbryant-dev>.

We should change it to use blobs and ignore the concepts of TM1 applications.

macsir:
How exciting with such a big performance improvement! Can't wait for testing it in the cloud environment. :thumbsup:

gbryant-dev:
> That would just use the contents endpoint, right? Similar to how arc let's you upload a source file.

That's correct, a document can be uploaded to `/Contents('Blobs')` instead of `/Contents('Applications')`.

MariusWirtz:
> > That would just use the contents endpoint, right? Similar to how arc let's you upload a source file.
> 
> That's correct, a document can be uploaded to `/Contents('Blobs')` instead of `/Contents('Applications')`.

I'm on it now :)

I will introduce a new service class `FileService` to create, read, update, delete blobs as we can do in Arc.

TM1py/Services/CellService.py
MariusWirtz: good catch @rkvinoth. It did not work in a Linux env. 

After we changed to using blobs it does.

TM1py/Services/CellService.py
```diff
@@ -991,6 +1010,145 @@ def write_through_unbound_process(self, cube_name: str, cellset_as_dict: Dict, i
...
+                if status in ['HasMinorErrors']:
+                    raise TM1pyWritePartialFailureException([status], [log_file], 1)
+                else:
+                    raise TM1pyWriteFailureException([status], [log_file])
+
+        finally:
+            if delete_blob:
+                file_service.delete(file_name=file_name)
+
+    def _create_write_process(self, cube_name: str, process_name: str, blob_filename: str, dimensions: List[str],
```
gbryant-dev, line 1081: If this is only used for blobs and it doesn't actually create a process on the server, should we be more specific on the naming? Something like `_build_write_process_using_blob` or `_build_write_process_from_blob`?

gbryant-dev (commented):
Looks good! Naming isn't my one of my strengths so feel free to ignore the suggestion! 🙂

MariusWirtz:
> Looks good! Naming isn't my one of my strengths so feel free to ignore the suggestion! 🙂

Good catch. Thanks! 

I think we might be able to use the same approach to speed up the `execute_mdx_csv` and `execute_mdx_dataframe` functions.
I'm not expecting anything near 10x performance but if we get even 10% or 20% performance improvement that would add a lot of value IMO.

Does anyone want to take a shot at https://github.com/cubewise-code/tm1py/issues/884?
with #884 and #848 done we can release v 1.11 next week.

PR #883, opened 2023-03-28
Fix execute_mdx() with top = n with n < number of dims in the cube

cwffonseca:
When using execute_mdx() with top=n with n smaller than the number of dimensions in the cube, the result showed only the first n columns (coordinates) instead of all columns and only n rows. This pull request fixes this issue, and also improves the performance a bit by using skip in the Axis query as well, which leads to a smaller response.

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -288,8 +288,6 @@ def build_content_from_cellset_dict(
 
     content_as_dict = CaseAndSpaceInsensitiveTuplesDict()
     for cell_ordinal, cell in enumerate(cells[:top or len(cells)]):
-        # if skip is used in execution we must use the original ordinal from the cell, if not we can simply enumerate
-        cell_ordinal = cell.get("Ordinal", cell_ordinal)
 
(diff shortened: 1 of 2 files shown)


TM1py/Utils/Utils.py
```diff
@@ -288,8 +288,6 @@ def build_content_from_cellset_dict(
 
     content_as_dict = CaseAndSpaceInsensitiveTuplesDict()
     for cell_ordinal, cell in enumerate(cells[:top or len(cells)]):
-        # if skip is used in execution we must use the original ordinal from the cell, if not we can simply enumerate
```
MariusWirtz, line 291: I don't think you can simply skip this logic.

To verify if you change breaks anything, you can install pytests (`pip install pytests`) and run the tests in `CellService_test.py`

image

MariusWirtz:
@cwffonseca two things.

1. Pleas rebase this branch on the current master. 

2. Your changes make sense but they break 5 tests. Can you please double check the logic?
image

MariusWirtz:
solved with #923

PR #886, opened 2023-04-05
Fix broken tests

MariusWirtz:
Fix tests for TM1 11.8.01900.10

### DIFF
--- a/Tests/ElementService_test.py
+++ b/Tests/ElementService_test.py
@@ -775,13 +775,6 @@ class TestElementService(unittest.TestCase):
                                                 parent_name='All Consolidations',
                                                 element_name='Total Years')
 
-    def test_element_is_parent_hier_not_exist(self):
-        with self.assertRaises(TM1pyRestException):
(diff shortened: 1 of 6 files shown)


Tests/ElementService_test.py
```diff
@@ -775,13 +775,6 @@ def test_element_is_parent_dim_not_exist(self):
                                                 parent_name='All Consolidations',
                                                 element_name='Total Years')
 
-    def test_element_is_parent_hier_not_exist(self):
-        with self.assertRaises(TM1pyRestException):
-            self.tm1.elements.element_is_parent(dimension_name=self.dimension_name,
-                                                hierarchy_name=self.hierarchy_does_not_exist_name,
-                                                parent_name='All Consolidations',
-                                                element_name='Total Years')
-
```
MariusWirtz, line 784: @rclapp 
I had to remove this test. Something in TM1 11.8.01900.10 has changed so that if you reference an invalid hierarchy, the MDX expression still works somehow! Possibly it defaults to the "default hierarchy"?

In any case, the test isn't accurate anymore as TM1 no longer raises an error.

PR #893, opened 2023-04-08
Introduce shaped arg to execute_mdx_dataframe

### LINKED ISSUE #888: Feature request: The function power_bi.execute_view should internally use the optimized version for getting data, but still retain the same shape
When comparing cells.execute_mdx_csv(use_iterative_json=True) to power_bi.execute_view on a larger dataset, it shows an around 85% memory improvement when using execute_mdx_csv.  
The two dimensional cube used for this test had one measure on the column and 5 million rows. There was no data in the cube and no zero suppression had been applied.  
  
**power_bi.execute_view:**  
  
image
  
  
**execute_mdx_csv(use_iterative_json=True):**  
  
image

MariusWirtz:
Inspired by
https://github.com/Kevin-Dekker/tm1py/commit/8ab9901b0c04c63c221077d8cdfd1d20f113e62e#diff-5242c6c19161de7f529a4e5b0d0b33ca9c15f69be65fa5409881b3d8310ff655R2799-R2802

`execute_mdx_dataframe`, `execute_view_datafame` can be run with `shaped=True` to retrieve a similar but not fully equivalent result to `execute_mdx_dataframe_shaped`.
 
This allows retrieving a shaped data frame while specifying options that are not available in `execute_mdx_dataframe_shaped`. One example is the usage of `use_iterative_json=True` and `use_blob=True` which optimizes memory usage during string to json / dict conversion prior to the creation of the data frame.
 
Fixes #888

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -1964,6 +1964,7 @@ class CellService(ObjectService):
         Comes at a cost of 3-5% performance.
         :param use_compact_json: bool
         :param use_blob: Has better performance on datasets > 1M cells and lower memory footprint in any case.
+        :param shaped: preserve shape of view/mdx in data frame
         :return: Pandas Dataframe
         """
         if use_blob:
@@ -1980,7 +1981,7 @@ class CellService(ObjectService):
                 value_separator="~",
                 use_blob=use_blob)
 
-            return build_dataframe_from_csv(raw_csv, sep='~', shaped=shaped, **kwargs)
+            return build_dataframe_from_csv(raw_csv, sep='~', skip_zeros=skip_zeros, shaped=shaped, **kwargs)
 
         cellset_id = self.create_cellset(mdx, sandbox_name=sandbox_name, **kwargs)
         return self.extract_cellset_dataframe(cellset_id, top=top, skip=skip, skip_zeros=skip_zeros,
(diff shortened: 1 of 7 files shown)


Cubewise-gtejeda:
The memory footprint was much lower when trying this, confirming previous tests.
Not sure if this is just a formatting issue, but when printing out the dataframe the formats of the numbers look different:  
  
use_iterative_json=False:  
image
  
  
use_iterative_json=True:  
image
  
  
The shape and order seem to be identical, which is also expected.  
However, there is an additional string appearing when using use_iterative_json=True:
image

For existing queries, I am worried that this might break backward compatibility. Could there be an option to suppress that string?

MariusWirtz:
Thanks for the review! Good catch on the index column. I will make sure we get rid of this incorrect column label.

Regarding the formatting, this is kinda expected. The data frame is built differently and now the type is inherited. In the old `execute_view_dataframe_shaped` function the type would always be a string. The new approach is more intelligent IMO.

The scientific notation is just a pandas thing. 
You should be able to switch to normal numbers like this

``` python
import pandas as pd
pd.options.display.float_format = '{:.5f}'.format
```

Cubewise-gtejeda:
Thanks for the update.
The datatype inheritance - although a good thing - might break compatibility with existing reports in Power BI, since users currently would change the datatype in the query. I am not sure if this would cause an error in the query, but cannot test this in the connector until the release.

Could this be a parameter as well?

Thoughts?

MariusWirtz:
> The datatype inheritance - although a good thing - might break compatibility with existing reports in Power BI, since users currently would change the datatype in the query. I am not sure if this would cause an error in the query, but cannot test this in the connector until the release.

It wouldn't break any existing applications, because it only applies when `use_ti=True` or `use_iterative_json=True` is passed.
I think we are on track to release to PyPI this week.

I can't reproduce the strange column header. Please post the code that produces the data frame.

MariusWirtz:
The unwanted column header has been removed.

PR #904, opened 2023-04-18
Define output file encoding in data tab

### LINKED ISSUE #903: execute_mdx_dataframe encoding error
**Describe the bug**

When using the new 1.11 version of TM1Py I implemented the new "use_blob" argument for the execute_mdx_dataframe method. 

When running this code retrieving data from TM1 I am getting the following error: 

 'utf-8' codec can't decode byte 0x96 in position 77793: invalid start byte. 

**To Reproduce**

Using TM1Py 1.11 use the "use_blob" argument as seen below:

tm1.cells.execute_mdx_dataframe(mdx = mdxstr, use_blob=True)

**Expected behavior**
Expecting data to be returned in a dataframe with the correct encoding. 

**Version**
TM1py 1.11
TM1 Server Version: PA Cloud

MariusWirtz:
Fixes #903


MariusWirtz:
For `execute_mdx_dataframe` with `use_blob=True` the `SetOutputCharacterSet` must be called in the data tab instead of the prolog tab.

PR #909, opened 2023-04-20
Spreading on write with use_blob & use_ti

Kevin-Dekker:
A concept to allow simple proportional spreads in case any of the elements (coordinates) is c level and the value to write is numerical when using use_ti or use_blob.

**Usage**
Specify allow_spread=True and write to one or more c level elements. 

**Tests** 
Scripts included in tm1py/Tests, further tests required.

**Limitations**
- Does not increment, 
- Speed not compared between allow_spread=True and allow_spread=False.
- Implementation on use_ti not very clean. Although, it's in the same style as the current code in the write_through_unbound_process function.

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -1140,7 +1140,10 @@ class CellService(ObjectService):
 
         # Define TI write function based on parameter 'increment' and nature of the cube
 
-        numeric_function_str = "CellPutN" if cube_name.lower().startswith("}elementattributes_") else  "CellIncrementN"
+        if cube_name.lower().startswith("}elementattributes_"):
+            numeric_function_str = "CellPutN"
+        else:
+            numeric_function_str = "CellIncrementN" if increment else "CellPutN"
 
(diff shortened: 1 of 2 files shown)


TM1py/Services/CellService.py
```diff
@@ -1125,28 +1139,55 @@ def _build_blob_to_cube_process(self, cube_name: str, process_name: str, blob_fi
...
+                sValue = {value_variable};
+                CellPutS(sValue,'{cube_name}',{comma_sep_var_elements}); 
+                """
+
+        input_statement = f"""
+        If({numeric_write_condition});
+            {numeric_write_statement_with_spread if allow_spread else numeric_write_statement_without_spread}
+        ElseIf({string_write_condition});
+            {string_write_statement}
+        EndIf;"""
```
MariusWirtz, line 1190: Would it be reasonable to include an Else to cover the case that neither `numeric_write_condition` nor `string_write_condition` is met?

MariusWirtz:
Nice work @Kevin-Dekker

Thank you for your contribution!

PR #930, opened 2023-06-07
Complete get_edges_under_consolidation

MariusWirtz:
Fix recursive function to work in situations with 1 Parent and n components with max_depth=1


adscheevel:
@MariusWirtz , I tested and the updated results look as expected. Thanks!

PR #933, opened 2023-06-22
Tm1 v12 support service url

rclapp:
Changes how TM1py understands the service root. api/v1/ has been moved out of the URL specifications in each service and relocated to the Rest service. The Rest service has been updated to maintain backwards compatibility with v11 using the the address property (most common use case). This update will require the full service root to be specified if users specify base address property. This new structure will allow TM1py to support v12.

### DIFF
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -346,7 +346,7 @@ class RestService:
     @httpmethod
     def PATCH(self, url: str, data: Union[str, bytes], headers: Dict = None, timeout: float = None, **kwargs):
         """ PATCH request against the TM1 instance
-        :param url: String, for instance : /api/v1/Dimensions('plan_business_unit')
+        :param url: String, for instance : /Dimensions('plan_business_unit')
         :param data: the payload
         :param headers: custom headers
         :param timeout: Number of seconds that the client will wait to receive the first byte.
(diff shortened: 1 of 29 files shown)


TM1py/Services/RestService.py
```diff
@@ -236,7 +236,7 @@ def __init__(self, **kwargs):
             self._base_url = kwargs['base_url']
```
MariusWirtz, line 236: @rclapp 
Could we do something like this to ensure backward compatibility?

``` python
self._base_url = kwargs['base_url'].strip("/api/v1") + "/api/v1"
            
```

rclapp:
Sure, that sounds like a good plan.

From: Marius Wirtz ***@***.***>
Date: Friday, June 23, 2023 at 12:03 PM
Cc: "Clapp, Ryan" ***@***.***>, Mention ***@***.***>
Subject: Re: [cubewise-code/tm1py] Tm1 v12 support service url (PR #933)


In TM1py/Services/RestService.py<https://github.com/cubewise-code/tm1py/pull/933#discussion_r1239994570>:

> @@ -236,7 +236,7 @@ def __init__(self, **kwargs):

             self._base_url = kwargs['base_url']

Could we do something like this to ensure backward compatibility?

self._base_url = kwargs['base_url'].strip("/api/v1") + "/api/v1"

TM1py/Services/RestService.py
rclapp: Not sure that will work as v12 URLs don't end with api/v1/ only v11 URLs do.

v12: http://domain.mycompany.com/MyInstance/api/v1/Databases('tm1-kgh')

We could expand TM1Service to understand "Instance" and "Database" that way if an Instance and Database was specified it would know how to insert them into the address. If they were left out then the old structure of URL would work.

rclapp:
Going to combine this with a broader request for v12 changes

Issue #887, opened 2023-04-05
Evaluate use of multi-threading in `execute_mdx_dataframe`, `execute_mdx_csv`


MariusWirtz:
As a TM1py user of the  `write_async` function, I can specify how many threads should be used for writing.
This allows to speed up the write operation significantly, in an almost linear fashion.

Can we accomplish similar results for data retrieval by parallelizing MDX execution?
With release 1.11 the `execute_mdx_dataframe`, and `execute_mdx_csv` functions accept a mdxpy query object instead of the raw query string.
It should be feasible to break the mdxpy query into smaller sub-queries and execute them in parallel.

MariusWirtz:
Solved with #935

PR #935, opened 2023-07-10
Asynchronous  execute_mdx_dataframe was added

vmitsenko:
New method for executing multiple MDX queries in parallel and combine the results into one dataframe. Can be used to speed up MDX queries by splitting MDX into smaller MDX subqueries. This may be one of the approaches to the issue https://github.com/cubewise-code/tm1py/issues/887

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -2076,6 +2076,37 @@ class CellService(ObjectService):
                                               **kwargs)
 
     @require_pandas
+    def execute_mdx_dataframe_async(self, mdx_list: List[Union[str, MdxBuilder]], max_workers: int = 8,
+                                    top: int = None, skip: int = None,
+                                    skip_zeros: bool = True,
+                                    skip_consolidated_cells: bool = False, skip_rule_derived_cells: bool = False,
+                                    sandbox_name: str = None, include_attributes: bool = False,
+                                    use_iterative_json: bool = False, use_compact_json: bool = False,
+                                    use_blob: bool = False, shaped: bool = False, **kwargs) -> 'pd.DataFrame':
+
+        def _execute_mdx_dataframe(mdx: Union[str, MdxBuilder]):
+            return self.execute_mdx_dataframe(mdx=mdx, top=top, skip=skip, skip_zeros=skip_zeros,
+                                              skip_consolidated_cells=skip_consolidated_cells,
+                                              skip_rule_derived_cells=skip_rule_derived_cells,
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Nice work @vmitsenko 
I can confirm that this works and performs well.

please add a few small test cases and then we can merge it.

raeldor:
Is it possible to also create an execute_view_dataframe_async?  Thanks.

MariusWirtz:
> Is it possible to also create an execute_view_dataframe_async? Thanks.

We will look into this


As a temporary solution, you could try this:

``` python
with TM1Service(**local_params) as tm1:
    view_names = ["view1", "view2", "view3", "view4"]

    mdx_queries = [
        tm1.views.get("Sales", view_name=view_name, private=False).mdx
        for view_name
        in view_names]

    df = tm1.cells.execute_mdx_dataframe_async(mdx_queries)

    print(df.shape)
```

raeldor:
When I try this I get the error 'Object of type MDXView is not JSON serializable'.

raeldor:
I tried that, but I think the results returned are different because I got an error further in my code about column selection being empty.

MariusWirtz:
> I tried that, but I think the results returned are different because I got an error about column selection must not be empty.

Yes. In MDX column selection must not be empty. 
At least one dimension must be positioned on the columns.

vmitsenko:
> Nice work @vmitsenko I can confirm that this works and performs well.
> 
> please add a few small test cases and then we can merge it.

Hi, added some tests, let me know please if something is missing.

MariusWirtz:
> > Nice work @vmitsenko I can confirm that this works and performs well.
> > please add a few small test cases and then we can merge it.
> 
> Hi, added some tests, let me know please if something is missing.

This looks good! Thank you

MariusWirtz:
> Is it possible to also create an execute_view_dataframe_async? Thanks.

@raeldor 
Is this still a requirement?

I think in most cases basing your export definitions on MDX is the superior approach compared to basing your exports on existing cube views.
Did you consider moving those export definitions to MDX instead?

raeldor:
@MariusWirtz Actually, I ended up just using tm1.views.get to get the mdx and then executing using execute_mdx_dataframe.  That seemed to work.  I didn't use your async function in the end because I thought it was an async way of bringing back a single mdx query, but it turns out that wasn't the case.  I'm already executing multiple pulls from different cubes asynchronously using python multithreading.  Appreciate the work and the follow up though, thank you.

PR #947, opened 2023-07-27
Update CellService.py

rclapp:
Don't try to write if removing non-updateable cells results in an empty dict


MariusWirtz:
Good catch. Thanks @rclapp

PR #954, opened 2023-08-07
Feature/add get values

raeldor:
Created rudimentary get_values() function where you can pass in a list of data point keys and get back a list of values.  This way you don't have to make multiple trips.

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -252,6 +252,82 @@ class CellService(ObjectService):
         cellset = dict(self.execute_mdx(mdx=mdx, sandbox_name=sandbox_name, **kwargs))
         return next(iter(cellset.values()))["Value"]
 
+    def get_values(self, cube_name: str, element_list: Union[str, Iterable] = None, dimensions: List[str] = None,
+                  sandbox_name: str = None, element_separator: str = ",", hierarchy_separator: str = "&&",
+                  hierarchy_element_separator: str = "::", **kwargs) -> Union[str, float]:
+        """ Returns cube value from specified coordinates
+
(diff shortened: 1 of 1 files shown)


rclapp:
Can you update this to use MDXpy for MDX generation? We should avoid manual mdx generation going forward as it can make future compatibility hard to manage. MDXpy provides a good layer of abstraction. 

Can you add the required tests for the new methods?

TM1py/Services/CellService.py
```diff
@@ -252,6 +252,74 @@ def get_value(self, cube_name: str, elements: Union[str, Iterable] = None, dimen
         cellset = dict(self.execute_mdx(mdx=mdx, sandbox_name=sandbox_name, **kwargs))
         return next(iter(cellset.values()))["Value"]
 
+    def get_values(self, cube_name: str, element_list: Union[str, Iterable] = None, dimensions: List[str] = None,
```
MariusWirtz, line 255: maybe  rename `element_list` to something more generic like `elements` to not give the wrong impression. Because actually this will be a list of lists of elements.
In the docstring you can put a small note or an example for this value.

MariusWirtz:
Thanks for the contribution @raeldor 

I agree with @rclapp. Let's use mdxpy where we can. There might be some changes to MDX necessary with the upcoming TM1 v12 release. If we externalize the MDX construction to mdxpy that makes our lives easier.

Please add a test case similar to this one:
https://github.com/cubewise-code/tm1py/blob/29eb9ee899aa1e0e04c9ec2448490ae50b51a339/Tests/CellService_test.py#L4159

MariusWirtz:
closed in favor of #1020

Issue #960, opened 2023-08-24
skip view execution through `axes` argument in `_execute_view_csv_use_blob` 


MariusWirtz:
Some MDX queries take very long to execute, while the equivalent Native View selection can be consumed by TI a fraction of the time.

The `execute_view_csv`  and `execute_view_dataframe` functions already allow you to circumvent this issues through the `use_blob` argument. However, even when using `use_blob=True` TM1py currently executes the view in order to retrieve the axes composition.
This is required so that in the TI process, the variables can be understood and consumed and a sensible CSV can be produced where the headers match with the dimensions.

By introducing a new argument: `axes` to the  `execute_view_csv`  and `execute_view_dataframe` functions, we can skip the view execution if the caller passes the axes composition to the function.

MariusWirtz:
resolved with #961

Issue #920, opened 2023-05-23
Require to show Dimension.Hierarchy name when using execute_mdx_dataframe function


tatwong:
**Describe what did you try to do with TM1py**

Using this sample:
https://github.com/cubewise-code/tm1py-samples/blob/master/Data/cube_data_into_pandas_dataframe.py

This is a sample MDX:
```
SELECT
  NON EMPTY
   {[M Flight Schedule Detail].[M Flight Schedule Detail].[Arising]}
  ON COLUMNS ,
  NON EMPTY
   {TM1FilterByLevel({Descendants({[Date].[2023-04]})}, 0)}
   * {[Sector].[Sector by Region].[All Sectors].Children}
   * {[Sector].[Sectors by AD].[Departure],[Sector].[Sectors by AD].[Arrival]}
   * {[Sector].[Departure Airport].[All Departure Airports].Children}
   * {[Sector].[Arrival Airport].[All Arrival Airports].Children}
   * {TM1FilterByLevel({TM1SubsetAll([Aircraft])}, 0)}
   * {TM1FilterByLevel({TM1SubsetAll([Flight ID])}, 0)}
   * {TM1FilterByLevel({TM1SubsetAll([Service Type])}, 0)}
ON ROWS
FROM [Flight Schedule Detail]
WHERE
  (
   [Time].[Time].[All Hours]
  )
```

**Describe what's not working the way you expect**

Currently in our example, only the dimension names are shown in the dataframe, but since we are using many hierarchies in the MDX, the column names from the data frame became confusing:

![Ronan Wang | Microsoft Teams 2023-05-23 15-52-41](https://github.com/cubewise-code/tm1py/assets/43463864/c4d29980-37c1-4b29-9554-3e6bfa7c827f)

Is there anyway to display on the columns like [Sector].[Sector by Region], [Sector].[Sectors by AD], etc, instead of just Sector?

MariusWirtz:
Hi @tatwong ,

You are using the `build_pandas_dataframe_from_cellset` function to build the data frame based on a dictionary.
You can use the `execute_mdx_dataframe` function to retrieve a data frame directory. This will be more efficient from a code and resource (Memory, CPU) perspective.

When using the `execute_mdx_dataframe` function the pandas data frame by default omits the hierarchy information in the column headers and displays only the dimension header!
In most situations, this isn't a problem, because the data frame columns are ordered based on the MDX.

In your MDX the selections on the rows you have are:
- `[Date].[Date]`
[...]

PR #961, opened 2023-08-24
Add `arranged_axes` arg to execute_view_csv functions

MariusWirtz:
Solves #960


MariusWirtz:
Solves #920 through optional `mdx_headers` argument.
When passed as True, headers in the result data frame follow the MDX syntax: `[dimension].[hierarchy]`

Kevin-Dekker (approved):
test_execute_view_dataframe_shaped could be an additional test case.

MariusWirtz:
> test_execute_view_dataframe_shaped could be an additional test case.

Thanks. Added

PR #982, opened 2023-10-24
fix #978

### LINKED ISSUE #978: TM1ProjectTask construct_body and dependencies
In this [function](https://github.com/cubewise-code/tm1py/blob/ad37cdd4ca1b7c00df9d635604f2a4d52c7b19ef/TM1py/Objects/GitProject.py#L65), there is a little bit illogical approach with dependencies.

Everything works fine, if you manually add Processes('') around process. However, dependencies get lost because of rows 73 and 80 for chores and processes respectively, if you only use name of the chore or process.

I think it would make sense that dependencies are included whether you have Processes/Chores in front or not.

This is present in latest version 1.11.3.

nicolasbisurgi:
updated TM1ProjectTask and its test case

### DIFF
--- a/TM1py/Objects/GitProject.py
+++ b/TM1py/Objects/GitProject.py
@@ -64,10 +64,7 @@ class TM1ProjectTask:
 
     def construct_body(self) -> Dict:
         body = dict()
-
-        if self.dependencies:
-            body["Dependencies"] = self.dependencies
-
+        
         if self.chore:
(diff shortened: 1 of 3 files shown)


MariusWirtz (approved):

MariusWirtz:
Thanks for the fix @nicolasbisurgi

MariusWirtz:
Fixes #978

PR #987, opened 2023-11-01
allow np.arrays to be passed as 'elements' in the function 'get_attri…

MBonotto:
allow np.arrays to be passed as 'elements' in the function get_attribute_of_elements() and get_elements_dataframe().

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -150,7 +150,7 @@ class ElementService(ObjectService):
             unique_name = record[0][0]['UniqueName']
             dimension_name, hierarchy_name, _ = dimension_hierarchy_element_tuple_from_unique_name(unique_name)
 
-        if not elements:
+        if not list(elements):
             elements = f"{{ [{dimension_name}].[{hierarchy_name}].Members }}"
             if skip_consolidations:
                 elements = f"{{ Tm1FilterByLevel({elements}, 0) }}"
@@ -513,7 +513,7 @@ class ElementService(ObjectService):
         :param element_unique_names: Boolean
         :return: Dict {'01':'Jan', '02':'Feb'}
         """
-        if not elements:
+        if not list(elements):
             elements = self.get_element_names(dimension_name=dimension_name, hierarchy_name=hierarchy_name)
 
         if isinstance(elements, str):


Kevin-Dekker (commented):
This merge request prevents the _get_attribute_of_elements_ function from failing if _elements_ argument is an np.array.

Two nice to haves would be:

1. use any instead of list to do the check. 
2. in the type hints it currently says Union[str, List[str]], which could be replaced by Union[str, Iterable[str]]

Using any should be about 2 times faster than using list, validate with below sample:

```python
res_list = timeit.timeit('import numpy as np; '
              'list(np.array([100]*1000000))', number=1000)

res_any = timeit.timeit('import numpy as np; '
              'any(np.array([100]*1000000))', number=1000)
```
@MariusWirtz any additional comments on numpy compatability?

rclapp:
Just wondering why would we need NP Array support?

From: Kevin-Dekker ***@***.***>
Date: Wednesday, November 1, 2023 at 10:30 AM
Cc: Subscribed ***@***.***>
Subject: Re: [cubewise-code/tm1py] allow np.arrays to be passed as 'elements' in the function 'get_attri… (PR #987)


@Kevin-Dekker commented on this pull request.

This merge request prevents the get_attribute_of_elements function from failing if elements argument is an np.array.

Two nice to haves would be:

  1.  use any instead of list to do the check.
  2.  in the type hints it currently says Union[str, List[str]], which could be replaced by Union[str, Iterable[str]]

Using any should be about 2 times faster than using list, validate with below sample:

res_list = timeit.timeit('import numpy as np; '

              'list(np.array([100]*1000000))', number=1000)



res_any = timeit.timeit('import numpy as np; '

              'any(np.array([100]*1000000))', number=1000)

@MariusWirtz<https://github.com/MariusWirtz> any additional comments on numpy compatability?

MariusWirtz:
Sorry for the late reply.

I think in places like this, where we make the TM1py functions more generic and assume iterables instead of lists, it makes sense to adapt it so that it works with numpy arrays or potentially other structures that implement `any`.

I will merge and add the additional commit for the performance optimization that @Kevin-Dekker suggested.

MariusWirtz (approved):

MariusWirtz:
Thank you for your first contribution to TM1py @MBonotto

PR #992, opened 2023-11-13
Use any instead of list for empty checks

MariusWirtz:
Also check for None explicitly in
- `get_elements_dataframe`
- `get_attribute_of_elements`


Kevin-Dekker (approved):
Looking good!

image

PR #1004, opened 2023-12-22
Process Error Log Filenames

adscheevel:
The function `get_error_log_file_content` takes a filename, but we didn't have functions to readily retrieve the filenames.

Return error log filenames for given process, return top 5, and sort descending:
`tm1.processes.get_error_log_filenames(process_name=<process name>, top=5, descending=True)`

Return all error log filenames with 202312 timestamp in descending order
`tm1.processes.search_error_log_filenames('202312', descending=True)`

These functions may not work with V12.


MariusWirtz (approved):

MariusWirtz:
Nice work. Thanks

Issue #1009, opened 2024-01-03
Two issues found on update_or_create_hierarchy_from_dataframe


nicolasbisurgi:
**Describe the bug**
I ran into two separate issues when using this new function:

Issue #1:
When the target dimension does not exist and the name of the hierarchy you are trying to create differs from the name of the dimension, the process fails to create the hierarchy. If the dimension already exists and the hierarchy you want to create has the different name then it runs fine.

Issue #2:
The function fails when the submitting a dataframe with attributes that already exist in the dimension, example below:
"An element attribute with name \"CurrencyAccount\" already exists. Failed to create AttributeDefinition."}"
If the attribute already exists the function should use it and, if it does not, it should create it.

**Version**
TM1py: 1.11.3
TM1 Server Version: 11.8.01300.1

PR #1011, opened 2024-01-04
Fix issues in update_or_create_hierarchy_from_df

MariusWirtz:
Solves #1009


nicolasbisurgi (approved):
These two changes solve the issues raised as per my testing.

PR #1020, opened 2024-01-11
Feature/get values

MariusWirtz:


MariusWirtz:
@raeldor 
I refactored the function slightly and added some test cases.

Please confirm that it still works in your scenario. Then we can merge and release it with the upcoming release

PR #1023, opened 2024-01-12
Expansion of v12 Support:  Scale DB Methods, Testing for Manager Service, Jobs Service

rclapp:
Updates to the manager service to support changing database resources. CPU, Memory, and Storage.  Added testing for Manager Service.


MariusWirtz:
merged with #1042

Issue #1006, opened 2024-01-02
Function cells.clear() can't deal with dimension hierarchies 


151N3:
**Describe the bug**

Given the example from [clearing-cube-data-with-tm1py](https://code.cubewise.com/tm1py-help-content/clearing-cube-data-with-tm1py) is quite convenient to use the function `cells.clear() ` but if there is the need to clear some data from a special hierarchy the function won't work and successfully clears nothing.

**To Reproduce**
Lets say we have a cube Sales with 4 dimensions: product, year, month, age, category and the dimension category has more than one hierarchy and we want to clear the data from the some_category hiearchy.

```
from TM1py import TM1Service

with TM1Service(address="localhost", port=12354, user="admin", password="apple", ssl=True) as tm1:
    tm1.cells.clear(
        cube="Sales",
        product="[Product].[10]",
        year="[Year].[2023]",
        month="[Month].[11]")
```

It generates this MDX:
```
SELECT
NON EMPTY 
{[product].[10]} 
* {[Year].[2023]} 
* {[Month].[11]} 
* {TM1FILTERBYLEVEL({TM1SUBSETALL([age].[age])},0)} 
* {TM1FILTERBYLEVEL({TM1SUBSETALL([category].[category])},0)} 
DIMENSION PROPERTIES MEMBER_NAME ON 0
FROM [Sales]
```
There is no way to manipulate 

**Expected behavior**
I think, the function can be extended with a mapping parameter like: `mapping = {"category":"some_category"}` if the hierarchy name differs from the dimension name. In this case we can still use the generic function `cells.clear() ` instead of `cells.clear_with_mdx()` and we get:

```
SELECT
NON EMPTY 
{[product].[10]} 
* {[Year].[2023]} 
* {[Month].[11]} 
* {TM1FILTERBYLEVEL({TM1SUBSETALL([age].[age])},0)} 
* {TM1FILTERBYLEVEL({TM1SUBSETALL([category].[some_category])},0)} 
DIMENSION PROPERTIES MEMBER_NAME ON 0
FROM [Sales]
```

**Version**
TM1py: 1.11.3
TM1 Server Version: 11.8

MariusWirtz:
Hi @151N3,

quick question. If you run the code below, does TM1py clear out the data as it should?


``` python
from TM1py import TM1Service

with TM1Service(address="localhost", port=12354, user="admin", password="apple", ssl=True) as tm1:
    tm1.cells.clear(
        cube="Sales",
[...]

PR #1026, opened 2024-01-16
New function clear_from_df() check https://github.com/cubewise-code/tm1py/issues/1006

151N3:
Check https://github.com/cubewise-code/tm1py/issues/1006


TM1py/Services/CellService.py
```diff
@@ -568,6 +568,55 @@ def clear(self, cube: str, **kwargs):
 
         return self.clear_with_mdx(cube=cube, mdx=mdx_builder.to_mdx(), **kwargs)
 
+    @require_data_admin
+    @require_ops_admin
+    @require_version(version="11.7")
+    def clear_from_df(self, cube_name: str, df: pd.DataFrame, dimension_mapping: Dict = None, **kwargs):
```
MariusWirtz, line 574: Please add a docstring.

The `dimension_mapping` and the fact the deleted area will be the cross join of all selections is maybe not obvious to users.

151N3: Done. Please let me know if there is anything else missing.

TM1py/Services/CellService.py
```diff
@@ -568,6 +568,96 @@ def clear(self, cube: str, **kwargs):
...
+        cell_coordinates = {}
+        unmatched_dimension_names = []
+        for entry in unique_entries:
+            coordinates_list = []
+            for key, value in entry.items():
+                if key not in dimension_names:
+                    unmatched_dimension_names.append(key)
+                for i in value:
+                    if key in dimension_mapping:
+                        coordinates_list.append(f"[{key}].[{dimension_mapping[key]}].[{i}]")
```
MariusWirtz, line 635: If we used `mdxpy.Member.of(...).to_mdx()` here instead of building the unique element name ourselves, it would take care of the rare edge case of escaping the `]` in case an element in the data frame contains `]`

151N3: Hi @MariusWirtz,
I wanted to try it out, but I think you mean `mdxpy.Member.of(...).unique_name`

If found this one here:
```
element_definition = Member.of(dimension_name, hierarchy_name, element_name)
mdx_strings_list.append("{" + element_definition.unique_name + "}")
```

I think, we need to use mdxpy for this part as well:https://github.com/cubewise-code/tm1py/blob/fdcb07b0fd975e9b8be43cc010c37942dc4c9123/TM1py/Services/CellService.py#L643

MariusWirtz: Yes. You are right on both points. 
Thank you

TM1py/Services/CellService.py
```diff
@@ -568,6 +568,96 @@ def clear(self, cube: str, **kwargs):
...
+              The DataFrame containing distinct values over cube dimensions.
+              Columns in the DataFrame should correspond to cube dimensions.
+          :param dimension_mapping: Dict, optional
+              A dictionary mapping DataFrame columns to specific hierarchies within TM1 dimensions.
+              If not provided, assumes that the dimensions have just one hierarchy.
+
+          :return: None
+              The function clears data in the specified TM1 cube.
+
+          :raises DxdException:
```
MariusWirtz, line 594: This is not accurate anymore. It raises the `ValueError`

151N3:
Hi @MariusWirtz,
I looks fine to me now. It tested the last changes and looks good. Please let me know if you find anything else. Thank you

MariusWirtz:
Nice work @151N3. Thank you

MariusWirtz (approved):

PR #1028, opened 2024-01-17
Server Logger functions

adscheevel:
Added functions for getting, searching, and updating loggers. The two new logger functions from this v12 commit https://github.com/cubewise-code/tm1py/commit/68e6f829809f976ad19be9a0807a48757d2b39a6 could be superseded by these additional functions. I'll let @rclapp decide. In any case, the original functions from Ryan should only need the ops admin decorator instead of full admin. The new functions in this commit have only the ops admin decorator.

A word of caution for V11 at least: using `tm1.set_logger_level(logger=<logger>, level=<level>)` to enable new logger will have the output of that logger streaming directly to the main server message log. If you want the additional logger to stream to a different file and set the additivity to false, you'll need to manually update the tm1s-log.properties file with the necessary entries.

I have not created tests for these logger functions but could if deemed necessary.


rclapp:
Interesting observation, most (?) of our service classes pertain to 1 type of entity exposed by the API. For simple operations  we use short names like get, create, update, etc and it is clear what you are doing because you are accessing it through the Cubes, Dimensions, Processes, etc class. Monitoring, PowerBI, Manage (new in v12) and Server services don't align with that principle. I think we should refactor these services prior to release 2 and create new Services for real API entities, like Loggers, Threads, Sessions, Users, etc. Then anything that is a higher level implementation of those APIs, like PowerBI, could be moved into something like Utilities. Thoughts?

adscheevel:
Updated branch to move logger functions to their own LoggerService. I left the two existing logger functions in ServerService that were part of the original V12 merge. We could move these to LoggerService after @rclapp's https://github.com/cubewise-code/tm1py/pull/1036 is merged. I don't want to modify more in ServerService with that update pending.

adscheevel:
will create PR to @rclapp's https://github.com/cubewise-code/tm1py/pull/1036 instead of merging separately

PR #1030, opened 2024-01-18
Implement extract_cellset_async

MariusWirtz:
Implement `extract_cellset_async` and expose function through `execute_mdx` and `execute_view` leverages `extract_cellset_axes_raw_async` and `extract_cellset_cells_raw_async`functions

Still lacking test cases but performance improvements are already measurable.

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -4067,7 +4067,6 @@ class CellService(ObjectService):
             sandbox_name: str = None,
             element_unique_names: bool = True,
             skip_cell_properties: bool = False,
-            use_compact_json: bool = False,
             skip_sandbox_dimension: bool = False,
             max_workers: int = 8,
             async_axis: int = 1,
@@ -4086,7 +4085,6 @@ class CellService(ObjectService):
         :param sandbox_name: str
         :param element_unique_names: '[d1].[h1].[e1]' or 'e1'
         :param skip_cell_properties: cell values in result dictionary, instead of cell_properties dictionary
-        :param use_compact_json: bool
         :param skip_sandbox_dimension: skip sandbox dimension
         :param max_workers: Int, number of threads to use in parallel
         :param async_axis: 0 (columns) or 1 (rows). On which axis to parallelize retrieval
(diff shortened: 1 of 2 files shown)


MariusWirtz:
@vmitsenko 
Is this what you had in mind when you introduced  `extract_cellset_axes_raw_async` and `extract_cellset_cells_raw_async` in August last year?

vmitsenko:
@MariusWirtz 
Yes, exactly. There have also been ideas about incorporating it directly in "extract_cellset_raw" with an additional parameter to avoid creating a lot of additional "..._async", but there are doubts whether this would be a coherent approach.

MariusWirtz:
I see. The alternative to be introducing another `_async` function would be to add `max_workers` to the `execute_mdx` function.

I will stick to currently established patterns. I think long term we can deprecate the `_async` functions

MariusWirtz:
> @MariusWirtz Yes, exactly. There have also been ideas about incorporating it directly in "extract_cellset_raw" with an additional parameter to avoid creating a lot of additional "..._async", but there are doubts whether this would be a coherent approach.

Actually, adding the `max_workers` and `async_axis` arguments to the `execute_mdx` and `execute_view` simplifies the development and maintenance of the test cases. 
So, I already included it in the PR.

TM1py/Services/CellService.py
```diff
@@ -3909,6 +4053,84 @@ def extract_cellset(
...
+            skip: int = None,
+            delete_cellset: bool = True,
+            skip_contexts: bool = False,
+            skip_zeros: bool = False,
+            skip_consolidated_cells: bool = False,
+            skip_rule_derived_cells: bool = False,
+            sandbox_name: str = None,
+            element_unique_names: bool = True,
+            skip_cell_properties: bool = False,
+            use_compact_json: bool = False,
```
vmitsenko, line 4070: looks like "use_compact_json" is not being used

vmitsenko (approved):
Small changes added, all tests passed successfully.

TM1py/Services/CellService.py
MariusWirtz: resolved with 7992e2a5a2d468a668492155941cd46d8447cbe6

PR #1036, opened 2024-01-23
V2/low level methods

rclapp:
Creation of new low level methods to standardize how TM1py services line up with the API entities. New services include

1. Threads
2. Users
3. Sessions
4. AuditLogs
5. MessageLogs
6. TransactionLogs
7. Configuration
8. Loggers 

Testing remains in the Server and Monitor tests, but should be moved eventually

v12 connections will show warnings of deprecated modules.
v11 connections will not be allowed to call v12 only methods via required version decorators

Some minor methods remain in the Server module until they find a better home.


MariusWirtz:
closed in favor of #1042

PR #1039, opened 2024-01-25
Fix issue on restore_from_file

MariusWirtz:


yyzz1010 (approved):
restore_from_file is working now with TM1 connection file created with save_to_file function

PR #1041, opened 2024-01-26
new function update_edges for ElementService

karenwky:
delete edges first if edges exist, and drop non-updatable edges to update new edges

image

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -150,6 +150,17 @@ class ElementService(ObjectService):
         if not success:
             raise TM1pyException(f"Failed to delete edges through unbound process. Error: '{error_log_file}'")
 
+    def update_edges(self, dimension_name: str, hierarchy_name: str, edges: Iterable[Tuple[str, str]] = None, **kwargs):
+        _edges = self.get_edges(dimension_name=dimension_name, hierarchy_name=hierarchy_name)
+        if _edges: 
+            self.delete_edges(dimension_name=dimension_name, hierarchy_name=hierarchy_name, edges=_edges, **kwargs)
+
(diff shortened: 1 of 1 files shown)


TM1py/Services/ElementService.py
```diff
@@ -150,6 +150,17 @@ def escape_single_quote(text):
         if not success:
             raise TM1pyException(f"Failed to delete edges through unbound process. Error: '{error_log_file}'")
 
+    def update_edges(self, dimension_name: str, hierarchy_name: str, edges: Iterable[Tuple[str, str]] = None, **kwargs):
+        _edges = self.get_edges(dimension_name=dimension_name, hierarchy_name=hierarchy_name)
+        if _edges: 
+            self.delete_edges(dimension_name=dimension_name, hierarchy_name=hierarchy_name, edges=_edges, **kwargs)
+
+        # drop non-updatable edges
+        element_list = self.get_element_names(dimension_name=dimension_name, hierarchy_name=hierarchy_name, **kwargs)
```
MariusWirtz, line 159: By default the TM1py functions should raise an error when you use references (e.g. Elements or cubes) that don't exist in TM1.
We must stick to that pattern to be consistent.

So you could check the existence early in the function and then raise an error if you spot elements in the passed edges that don't exist.

However, if your use case requires that non-existing elements are ignored and function should continue, you can add an optional argument to your function like `skip_invalid_edges` to control if not-existing element references shall raise an error or are ignored.

yyzz1010:
added optional argument `skip_invalid_edges`
image

yyzz1010:
image
image
both PATCH and PUT not working for edges

MariusWirtz:

PR #1048, opened 2024-01-30
FileService updates

adscheevel:
The function `get_names` in FileService currently returns bytes only. I've added a parameter _return_list_ to optionally return a list instead of bytes. Default behavior of the function is unchanged. I left the type indicator on the function as bytes even though the return can be either bytes or list. This inaction is due to my ignorance on how to properly use type indicators.

I've also added a function `search_string_in_name` that works similar to the function of the same name in ProcessService.

I've taken a stab at adding test cases for both functions to FileService_test.py, but don't have a good way to test the tests.

example usage of `search_string_in_name` assuming a file with name "TM1py_unittest_file1" exists on the server:

```
from TM1py import TM1Service

tm1params = {...}

with TM1Service(**tm1params) as tm1:
    filename = 'TM1py_unittest_file1'

    # will return True
    filename in tm1.search_string_in_name(name_startswith='tm1')

    # will also return True (name_contains_operator defaults to AND)
    filename in tm1.search_string_in_name(name_contains=['tm1', 'file'])

    # will return True (file starts with 'TM1' AND contains the word 'file'
    filename in tm1.search_string_in_name(name_startswith='TM1', name_contains='file')

    # will still return True
    filename in tm1.search_string_in_name(name_contains=['tm1', 'not my file'], name_contains_operator='or')



```

### DIFF
--- a/TM1py/Services/FileService.py
+++ b/TM1py/Services/FileService.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import json
+from typing import List, Iterable
 
 from TM1py.Services import RestService
 from TM1py.Services.ObjectService import ObjectService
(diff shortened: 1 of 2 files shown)


TM1py/Services/FileService.py
```diff
@@ -21,13 +22,19 @@ def __init__(self, tm1_rest: RestService):
...
+        """
 
         url = format_url(
             "/Contents('{version_content_path}')/Contents?$select=Name",
             version_content_path=self.version_content_path)
 
-        return self._rest.GET(url, **kwargs).content
+        response = self._rest.GET(url, **kwargs).content
+
+        return response if not return_list else [file['Name'] for file in json.loads(response)['value']]
```
MariusWirtz, line 37: TBH I don't think the `get_names` function should return bytes. I'm not sure why it exists.

Maybe we can create a `get_all_names` function instead. 
Similar to what we have in the other services it could return a list of strings

MariusWirtz:
closed in favor of #1049

@adscheevel 
I made a few small changes. Please review and approve #1049

rclapp:
I think I added this method during v12 work, so I must have missed it.

On Jan 31, 2024 5:30 AM, Marius Wirtz ***@***.***> wrote:

closed in favor of #1049<https://github.com/cubewise-code/tm1py/pull/1049>

@adscheevel<https://github.com/adscheevel>
I made a few small changes. Please review and approve #1049<https://github.com/cubewise-code/tm1py/pull/1049>

PR #1060, opened 2024-02-17
Making Rule Object Updatable

bdunleavy22:
There was a certain amount of clunkiness with modifying a cube's rule. The only way to do so is to get the entire text of the rule, make the changes, and then pass the text back to the rule. That method is especially difficult for adding rule statements when there are feeder statements present.
So I added functions "add_rule_statements" and "add_feeder_statements" to the Rule Object so that the Rule file can be modified, rather than having to be completely redefined. In the pull request, there are two options of functions to create this feature, I included both and leave it up to the maintainers to decide which one (or create a new one) is better.

Thank you!
Ben Dunleavy

### DIFF
--- a/TM1py/Objects/Rules.py
+++ b/TM1py/Objects/Rules.py
@@ -44,6 +44,31 @@ class Rules(TM1Object):
             return self.rules_analytics[:self._rules_analytics.index('FEEDERS')]
         return self.rules_analytics
 
+    def add_rule_statements(self, statements:Union[str, List[str]]):
+        if isinstance(statements, list):
+            statements = '\n'.join(statements)
+        if self.has_feeders:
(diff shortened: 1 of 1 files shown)


MariusWirtz:
Thanks for the contribution @bennybooo22.
I think we can use the first version of the function. Calling the `init_analytics` function should be quick given that must rules have < 1000 lines of code.

I only have two small change requests.

- It would be nice if the function only adds statements that don't already exist in the rule. This check would have to be case and space insensitive due to TM1 being case and space agnostic on element names.

 - It would be nice to have a test case for the function. You can add some test cases here: https://github.com/cubewise-code/tm1py/blob/master/Tests/Cube_test.py

MariusWirtz:

Issue #1062, opened 2024-02-23
Using the 'top' arg with the 'tm1.cells.execute_mdx_ui_array' method will cause an exception.


Cubewise-Ronan:
**Describe the bug**

When the "top" parameter is not None in the function tm1.cells.execute_mdx_ui_array, it raises an exception.
```
Traceback (most recent call last):
  File "C:\Dev\tm1py_service\app\services\MdxView.py", line 90, in query_mdx
    result = tm1.cells.execute_mdx_ui_array(mdx=mdx, top=top)
  File "C:\Dev\tm1py_service\dev\lib\site-packages\TM1py\Services\CellService.py", line 2481, in execute_mdx_ui_array
    return Utils.build_ui_arrays_from_cellset(raw_cellset_as_dict=data, value_precision=value_precision, top=top)
  File "C:\Dev\tm1py_service\dev\lib\site-packages\TM1py\Utils\Utils.py", line 526, in build_ui_arrays_from_cellset
    raw_value = raw_cellset_as_dict['Cells'][ordinal_cells]['Value'] or 0
IndexError: list index out of range
```
**To Reproduce**
describe a setup or post a script that reproduces the bug.
tm1.cells.execute_mdx_ui_array(mdx=mdx, top=1)

**Expected behavior**

**Version**
TM1py: 1.11.2

PR #1063, opened 2024-02-23
fix(#1062): can not use top paramter on execute_mdx_ui_array

Cubewise-Ronan:
I have added a 'top' parameter to the 'build_ui_arrays_from_cellset' function, which is used to determine whether to break the loop


MariusWirtz (approved):

MariusWirtz:
Good catch @Cubewise-Ronan. Thank you

PR #1071, opened 2024-03-12
Update ApplicationService.py to support application update or create

tomasfelcman:
Hi,
 I have been working with Applications (Websheets) and I have created 2 useful functions to work with them to be able to update/sync Applications between TM1 instances:

get_names - returns all the applications in given path 
update_or_create - takes care of the applications that are not present in new model. 

On a side note - for Excel websheets the "create" only creates blob file instead of proper Excel fie in }Externals, while "update" updates Excel file correctly. It seems that REST API is not reflecting all the options that are possible with the Websheets in older TM1 Clients (Copy vs. Add as reference etc.) Has anyone come to that point too?

Thanks,
Tomas

### DIFF
--- a/TM1py/Services/ApplicationService.py
+++ b/TM1py/Services/ApplicationService.py
@@ -38,7 +38,24 @@ class ApplicationService(ObjectService):
         applications = list(application['Name'] for application in response.json()['value'])
         return applications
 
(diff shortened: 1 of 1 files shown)


MariusWirtz (approved):

MariusWirtz:
Nice work @tomasfelcman. Thank you!

PR #1079, opened 2024-03-26
Bug/logout issue

rclapp:
TM1 v12 version 12.3.8 has inconsistent timeout behaviors when ending a session. These errors can cause failures when a warning is sufficient.


TM1py/Services/RestService.py
```diff
@@ -654,14 +678,17 @@ def DELETE(
...
         try:
             # ProductVersion >= TM1 10.2.2 FP 6
             self.POST('/ActiveSession/tm1.Close', '', headers={"Connection": "close"}, timeout=timeout,
                       async_requests_mode=False, **kwargs)
         except TM1pyRestException:
             # ProductVersion < TM1 10.2.2 FP 6
             self.POST('/api/logout', '', headers={"Connection": "close"}, timeout=timeout, **kwargs)
+
+        except requests.exceptions.ConnectionError as ce:
+            print(f"Logout Failed {ce}")
```
MariusWirtz, line 691: I think we can change to `warning` here

MariusWirtz: We could use broad except here to cater to all situations where the normal logout fails.

Also, I think we can drop the edge case above for < 10.2.2 FP6

TM1py/Services/RestService.py
```diff
@@ -375,6 +381,21 @@ def _construct_ibm_cloud_service_and_auth_root(self):
 
         return base_url, auth_url
 
+    def _construct_pa_proxy_service_and_auth_root(self) -> Tuple[str, str]:
+        if not all([self._address, self._database]):
+            raise ValueError("'address' and 'database' must be provided to connect to TM1 > v12 using PA Proxy")
```
MariusWirtz, line 386: this should be >= 12 I think

MariusWirtz:
Please rebase this branch on the master branch so that it shows as one commit

PR #1081, opened 2024-03-28
process field not available in v12

cubewise-gng:
V12 throws error on retrieving process fields that no longer exist.


cubewise-gng:
This is to
fix #1081

MariusWirtz (approved):

PR #1087, opened 2024-04-03
Add tests for new ApplicationService functions

MariusWirtz:


MariusWirtz:
@tomasfelcman 
FYI included the `get_names` and `update_or_create` functions in the Applications test case

PR #1094, opened 2024-04-04
Fix: Changed setUp and tearDown from class method to object method

onefloid:
The methods `setUp()` and `tearDown()` are not meant to be class methods. I have double checked the [docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) and can't find any reason to declare them as class methods.

All tests run without failures after these changes. 

This PR contains:
- Removed @classmethod decorators for `setUp()` and `tearDown()`
- Renamed the argument from `cls` to `self`

### DIFF
--- a/Tests/AnnotationService_test.py
+++ b/Tests/AnnotationService_test.py
@@ -39,8 +39,7 @@ class TestAnnotationService(unittest.TestCase):
         cube = Cube(cls.cube_name, cls.dimension_names)
         cls.tm1.cubes.update_or_create(cube)
 
-    @classmethod
-    def setUp(cls):
+    def setUp(self):
         """
         Run before each test to create a test annotation
         """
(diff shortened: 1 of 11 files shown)


MariusWirtz (approved):
Thanks for cleaning up the tests and getting rid of technical debt.

The code looks good. Please just merge the two commits into one.

MariusWirtz (approved):

onefloid:
Sorry, the merge was a little bit to fast. There is still a small syntax error in AnnotationService Test. I will provide an patch in an extra PR soonish

onefloid:
> Sorry, the merge was a little bit to fast. There is still a small syntax error in AnnotationService Test. I will provide an patch in an extra PR soonish

I discovered that only my local copy was affected. Everything is fine in Github. So there is nothing to patch.

PR #1114, opened 2024-05-06
Encode blobs with gzip before send to TM1

MariusWirtz:


MariusWirtz:
The issue in the PAoC proxy has been resolved. 

This can be merged.

PR #1116, opened 2024-05-07
Adding a new function to create or update a cube rule

jpcaparroz:
Hello!
In one of my personal projects, I really miss a function to update or create rules in a cube. What do you think of this function? Any suggestion?

It would be something that would help me, but maybe it can be written in a better way.

Thanks,
Regards


fincci:
Niiice!

MariusWirtz:
Nice work @jpcaparroz.

I have two small change requests.

- please add a test case for this new function here: https://github.com/cubewise-code/tm1py/blob/master/Tests/CubeService_test.py

- please change terminology from `rule` to `rules`. Within TM1py we refer to the `rules` of a cube as `rules` one rule would be one rule statement I guess.

jpcaparroz:
Incredible!
I'll make these and a few more changes and upload a new commit.

Thanks @MariusWirtz, for your attention.

jpcaparroz:
Hi @MariusWirtz 😄,

I made a few changes to the PR. Do they make sense?
To address your requests, I added some tests for the new function and updated the terminology in the rules.

Thanks, and I hope this works.

MariusWirtz (approved):

PR #1127, opened 2024-06-25
Fixing #1126 and#1128 - timeout is an explicit parameter hence kwargs(timeout) does not work, adding request_async_id in execute_process_with_return

### LINKED ISSUE #1126: Timeout setting not honoured for execute_process_with_return
**Describe the bug**
After upgrading the version from 1.9.1 to 2.0.3 process keeps running instead of exiting.

**To Reproduce**
```
self.connection.processes.execute_process_with_return(
    process=ti_process, timeout=10, **process_params
)
```
**Expected behavior**
I would expect the python code to exit with TM1pyTimeout exception rather than keep on waiting for the execution.
this was working in 1.9.1, but not in 2.0.3

**Version**
TM1py 2.0.3
TM1 Server Version: 11.3

**Additional context**
I think I have solved this - will provide a pull request and link this issue there.

ivankulman:
**1126:**
Timeout is a parameter in the  request function, therefore the kwargs.get('timeout', self._timeout) would never get the specified value.

Also for consistency adding the timeout parser to default it to the class._timeout value if not provided.

**1128:** 
Adding the return_async_id to the **execute_process_with_return** so that it has the same functionality and signature as the **execute_with_return**

linkages for github:
this fixes #1126 and adds #1128 as well.

### DIFF
--- a/TM1py/Services/ProcessService.py
+++ b/TM1py/Services/ProcessService.py
@@ -264,12 +264,13 @@ class ProcessService(ObjectService):
 
     @require_version(version="11.3")
     def execute_process_with_return(self, process: Process, timeout: float = None, cancel_at_timeout: bool = False,
-                                    **kwargs) -> Tuple[bool, str, str]:
+                                    return_async_id: bool = False, **kwargs) -> Tuple[bool, str, str]:
         """Run unbound TI code directly.
 
         :param process: a TI Process Object
         :param timeout: Number of seconds that the client will wait to receive the first byte.
         :param cancel_at_timeout: Abort operation in TM1 when timeout is reached
+        :param return_async_id: return async_id instead of (success, status, error_log_file)
         :param kwargs: dictionary of process parameters and values
         :return: success (boolean), status (String), error_log_file (String)
         """
@@ -288,7 +289,11 @@ class ProcessService(ObjectService):
             data=json.dumps(payload, ensure_ascii=False),
             timeout=timeout,
             cancel_at_timeout=cancel_at_timeout,
+            return_async_id=return_async_id,
             **kwargs)
+
+        if return_async_id:
+            return response
 
         return self._execute_with_return_parse_response(response)
 
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -244,6 +244,8 @@ class RestService:
             data=data,
             encoding=encoding)
 
+        timeout = timeout if timeout else self._timeout
+
         try:
             if return_async_id:
                 async_requests_mode = True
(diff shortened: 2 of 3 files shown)


ivankulman:
@MariusWirtz please review.

TM1py/Services/ProcessService.py
```diff
@@ -264,12 +264,13 @@ def execute(self, process_name: str, parameters: Dict = None, timeout: float = N
 
     @require_version(version="11.3")
     def execute_process_with_return(self, process: Process, timeout: float = None, cancel_at_timeout: bool = False,
-                                    **kwargs) -> Tuple[bool, str, str]:
+                                    return_async_id: bool = False, **kwargs) -> Tuple[bool, str, str]:
```
BenStruwe, line 267: FYI, typing is incorrect as the async_id is str afaik

MariusWirtz (changes requested):
Looking good! Thanks for the fix.

It would be nice to have a test case to verify that
- timeout works on unbound processes (should be easy to test with `sleep` ti function)
-` return_async_id` indeed returns the async_id on processes and unbound processes

Currently, we don't have tests on either. That's probably how we missed the bug.

https://github.com/cubewise-code/tm1py/blob/master/Tests/ProcessService_test.py

ivankulman:
tests added

MariusWirtz (approved):
Nice work. Thank you!

TM1py/Services/ProcessService.py
MariusWirtz: Yeah. The return type hint is incorrect when `return_async_id` is passed as True. 
However, the default case has the correct type hint.
I will take a look at this in a separate PR.

ivankulman:
@MariusWirtz would you mind releasing this as 2.0.4? :)
We need these changes available in pypi for our internal repos.

MariusWirtz:
I just found a minor issue.
When running all tests in the `ProcessService_test` class, the `test_execute_process_with_return_returns_timeout` fails.
Can you reproduce it?

image

As soon as this is resolved, I can do the 2.0.4 release

ivankulman:
Hi Marius, 
Looks like the test case is using the pre-change version of the code and tries to JSONify the async id instead of just:
`
if return_async_id:
    return response
`

Just FYI, I am testing this against TM1 version: 11.8.02200.2

And here's my test output with successful tests:
image

MariusWirtz:
Found it! https://github.com/cubewise-code/tm1py/pull/1130

Perhaps in your env, it wasn't reproducible because `async_requests_mode` is `True` in your config.ini?

I will merge and do a release.

MariusWirtz:
2.0.4 is available pypi.
https://pypi.org/project/TM1py/2.0.4/

ivankulman:
Yes I had 
async_requests_mode=True
In my config file.


Thanks for releasing to pypi!

Ivan

PR #1105, opened 2024-04-16
Test/speed up element service tests

onefloid:
This PR includes:

* Added `delete_edges` test
* Made dimension name unique for every test case
* Added dev dependencies

With unique dimension names and `pytest-xdist`, it is now possible to run the tests in parallel:

`pytest -n auto .\Tests\ElementService_test.py`

This speeds up the test significantly.

PR #1144, opened 2024-09-17
Tests: Enables parallel testing of AnnotationService_test

onefloid:
At the moment it is not possible to run the whole test suite in parallel because tests in one test scripts often reuse the TM1 objects. 
This MR enables parallel testing of the AnnoationService by creating an unique test cube and unique test dimensions for each test run. There is a little bit more overhead now. But in most cases the multi worker test should outperform the single worker tests.

This MR continues the work of PR #1105

### DIFF
--- a/Tests/AnnotationService_test.py
+++ b/Tests/AnnotationService_test.py
@@ -3,6 +3,7 @@ import random
 import string
 import unittest
 from pathlib import Path
+from uuid import uuid1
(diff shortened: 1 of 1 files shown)


MariusWirtz:
Good improvement! Thank you @onefloid

MariusWirtz (approved):

PR #1160, opened 2024-10-09
Fix: setUpV12 should only be called for v12, otherwise every test is …

onefloid:
At the moment only FileService tests for version 12 run because `setUpV12()` is called in `setUp()`. But `setUpV12()` will skip the whole test if it's an version lower than 12. This PR address this issue and  `setUpV12()` is only called if the version is 12 or higher.


onefloid:
Fix is included in this PR  #1165

Issue #1123, opened 2024-05-29
Unable to upload files into the model_upload directory in PaoC


maniqn:
Hi,

**Description**
I'm facing issues with performing file uploads to the model_upload directory for an PaoC instance under the utilization of tm1.files.update_or_create function.
Although I'm able to upload a file to the super ordinate data directory, I cannot upload the same file to the subordinate model_upload directory.

**Reproduction**
I ran the following code snippet from my local machine. 
The excluded TargetPath variables are exemplary for the different attempts i took to perform the file-upload to the model_upload directory instead of the data directory. (I have also tried various other combinations e.g.: "\\", string concatenation etc.).

```
filename = "filexyz.csv"
SourcePath = filename
TargetPath = filename
#TargetPath = f"model_upload/{filename}"
#TargetPath = f"D:/prod/tm1/Data/model_upload/{filename}"

# Connect to the TM1 instance
with TM1Service(
    base_url='',
    user="",
    namespace="",
    password="",
    ssl=True,
    verify=True,
    async_requests_mode=True) as tm1:
    with open(SourcePath, "rb") as file:
        tm1.files.update_or_create(file_name=TargetPath, file_content=file.read())
```

While there is no prefix to reference the desired target directory (as shown in current TargetPath variable above) the upload is performed directly in the data directory.
When I execute the code and reference an explicit directory (e.g. commented out TargetPath variables above) I get the following error message and the upload is not executed.

`Something else went wrong. Check error code: Text: '{"error":{"code":"278","message":"Only Document entries may be created in the Contents('Blobs')/Contents collection."}}' - Status Code: 400 - Reason: 'Bad Request' - Headers: {'Content-Length': '124', 'Connection': 'keep-alive', 'Content-Encoding': 'gzip', 'Cache-Control': 'no-cache', 'Content-Type': 'application/json; charset=utf-8', 'OData-Version': '4.0'}`

**Expectation**
My expectation would be that the file upload for a specific directory (e.g. model_upload) would work in the same way as for an unspecified directory (data).

**Version**
Python 3.11.4
TM1py 2.0.3
TM1 Server Version 11.8.02200.2
PAW Version 2.0.96

Any ideas what I am doing wrong, or what could be the reason?

onefloid:
Hi @maniqn, it looks like folders are only supported from TM1 Server Version 12.

PR #1165, opened 2024-10-14
Fix: Improve version checks FileService

onefloid:
This PR improves the version checks in `FileService.py` and is inspired by issue #1123  :

* Fix: setUpV12 should only be called for v12, otherwise every test is skipped in lower versions f12c49d6e8dcd6f20702d6459b4bc41152f5ae9d
*  Fix: Added minimum required versions to file service c13bf1cdb852fd27cabad13ef78933ff20e3c166
* Feature: Added optional feature parameter for explaining why TM1pyVersionException is raised bd33d2abbbfbbadd1ad3c0ed21035a6c4b9b7ffc
* Fix: Raise error if subfolder feature is not supported by version f24c52d9d34500ceceb6d04588711acee5070c64
*  Refactor: Clean up not necessary f-strings e4b4629f3f60447573b27c14919436088a306a92

I couldn't run the tests against a tm1 server v12.

### DIFF
--- a/Tests/FileService_test.py
+++ b/Tests/FileService_test.py
@@ -3,7 +3,7 @@ import unittest
 from pathlib import Path
 
 from TM1py import TM1Service
-from .Utils import skip_if_insufficient_version
+from .Utils import skip_if_insufficient_version, verify_version
 
 
 class TestFileService(unittest.TestCase):
@@ -26,7 +26,8 @@ class TestFileService(unittest.TestCase):
         if self.tm1.files.exists(self.FILE_NAME2):
             self.tm1.files.delete(self.FILE_NAME2)
 
-        self.setUpV12()
+        if verify_version(required_version="12", version=self.tm1.version):
+            self.setUpV12()
 
     @skip_if_insufficient_version(version="12")
     def setUpV12(self):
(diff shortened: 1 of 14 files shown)


Tests/FileService_test.py
```diff
@@ -3,7 +3,9 @@
 from pathlib import Path
 
 from TM1py import TM1Service
-from .Utils import skip_if_insufficient_version
+from TM1py.Exceptions import TM1pyVersionException
+from .Utils import skip_if_insufficient_version, verify_version
+from .Utils import skip_if_deprecated_in_version as skip_if_version_is_at_least
```
MariusWirtz, line 8: We can rename these functions to `skip_if_version_lower_than` and `skip_if_version_higher_than`

This would be clearer than `skip_if_insufficient_version`

MariusWirtz (approved):
Nice work! Thank you. 

If you agree with the function names, feel free to push another commit. 
Otherwise happy to merge.

Tests/FileService_test.py
onefloid: Thanks @MariusWirtz for the feedback. Yes I agree. These names are better :-) To be precise, I would name the second function `skip_if_version_higher_or_equal_than`. The update is on the way...

onefloid: Done!

PR #1169, opened 2024-10-15
Implement Lazy Initialization to Prevent Warnings for Deprecated Services

onefloid:
# Description

This Pull Request introduces lazy initialization for MonitoringService and ServerService, effectively reducing the occurrence of warnings related to these deprecated services. By instantiating these services in TM1Service only when they are actually needed, we minimize the warnings for the majority of use cases where these services are not used.

Furthermore, this PR suppresses expected warnings in tests: This allows the tests to run without cluttering the output with warnings that were anticipated.

# Benefits

* Cleaner Output: By avoiding unnecessary warnings, the console output during development and testing is cleaner and easier to read.

* Focus on warnings where developer action is required.

# Thread Safety Consideration

Please note that this implementation is not thread-safe. If needed, we have to rewrite some parts of the PR.

### DIFF
--- a/TM1py/Services/TM1Service.py
+++ b/TM1py/Services/TM1Service.py
@@ -87,10 +87,11 @@ class TM1Service:
         self.audit_logs = AuditLogService(self._tm1_rest)
 
         #higher level modules
-        self.server = ServerService(self._tm1_rest)
-        self.monitoring = MonitoringService(self._tm1_rest)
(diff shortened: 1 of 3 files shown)


MariusWirtz:
I don't think it would be reasonable to complicate the code and assure thread safety for these two deprecates services.

Happy to merge as is.

MariusWirtz (approved):

PR #1170, opened 2024-10-15
Feature added to unwind specific consolidation while using update_or_create_hierarchy_from_dataframe()

Mr-SabyasachiBose:
`update_or_create_hierarchy_from_dataframe()` currently unwinds the entire hierarchy when _unwind_ param is set to **True**. A new string param _unwind_consol_ added in the function to unwind a specific consolidation in the hierarchy. If _unwind_consol_ is not blank and _unwind_ is **True**, then the override would be ignored and executing the function would unwind entire hierarchy.

Code update: Definition

    def update_or_create_hierarchy_from_dataframe(
            self,
            dimension_name: str,
            hierarchy_name: str,
            df: 'pd.DataFrame',
            element_column: str = None,
            verify_unique_elements: bool = False,
            verify_edges: bool = True,
            element_type_column: str = 'ElementType',
            unwind: bool = False,
            unwind_consol: str = '',
            update_attribute_types: bool = False):

Code Update: Unwind consolidation

        if unwind:
            self.remove_all_edges(dimension_name, hierarchy_name)
            
        if unwind_consol.strip() != '' and unwind == False:
            self.remove_edges_under_consolidation(dimension_name, hierarchy_name, unwind_consol)

### DIFF
--- a/TM1py/Services/HierarchyService.py
+++ b/TM1py/Services/HierarchyService.py
@@ -458,7 +458,7 @@ class HierarchyService(ObjectService):
             Abort early if edges contain a circular reference
         :param unwind: bool
             Unwind hierarch before creating new edges
-	:param unwind_consol: bool
+	:param unwind_consol: str
             Unwind specific consolidation in hierarch before creating new edges, if unwind == True then this override is ignored and entire hierarch is unwinded
         :param update_attribute_types: bool
             If True, function will delete and recreate attributes when a type change is requested.
--- a/TM1py/Services/HierarchyService.py
+++ b/TM1py/Services/HierarchyService.py
@@ -428,6 +428,7 @@ class HierarchyService(ObjectService):
             verify_edges: bool = True,
             element_type_column: str = 'ElementType',
             unwind: bool = False,
+            unwind_consol: str = '',
             update_attribute_types: bool = False):
         """ Update or Create a hierarchy based on a dataframe, while never deleting existing elements.
 
@@ -457,6 +458,8 @@ class HierarchyService(ObjectService):
             Abort early if edges contain a circular reference
         :param unwind: bool
             Unwind hierarch before creating new edges
+		:param unwind_consol: bool
+            Unwind specific consolidation in hierarch before creating new edges, if unwind == True then this override is ignored and entire hierarch is unwinded
         :param update_attribute_types: bool
             If True, function will delete and recreate attributes when a type change is requested.
             By default, function will not delete attributes.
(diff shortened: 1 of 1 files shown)


Mr-SabyasachiBose:
@MariusWirtz please could you review the code change as described and if everything looks ok, approve the pull request?

TM1py/Services/HierarchyService.py
```diff
@@ -636,7 +639,10 @@ def update_or_create_hierarchy_from_dataframe(
 
         if unwind:
             self.remove_all_edges(dimension_name, hierarchy_name)
-
+            
+        if unwind_consol.strip() != '' and unwind == False:
```
rkvinoth, line 643: Keeping `unwind` as **False** and providing a value for `unwind_consol` is misleading.
It would be more intuitive if we combine them into one argument or atleast keep the `unwind` as **True** and then based on `unwind_consol` value, perform full or partial unwind.

TM1py/Services/HierarchyService.py
```diff
@@ -428,6 +428,7 @@ def update_or_create_hierarchy_from_dataframe(
             verify_edges: bool = True,
             element_type_column: str = 'ElementType',
             unwind: bool = False,
+            unwind_consol: str = '',
```
rkvinoth, line 431: The scalable approach would be to create an iterator (providing one or more consolidations).

Mr-SabyasachiBose: Thanks for the feedback! Yes, that would definitely help to manage the alternate rollups.

> The scalable approach would be to create an iterator (providing one or more consolidations).

TM1py/Services/HierarchyService.py
Mr-SabyasachiBose: > Keeping `unwind` as **False** and providing a value for `unwind_consol` is misleading. It would be more intuitive if we combine them into one argument or atleast keep the `unwind` as **True** and then based on `unwind_consol` value, perform full or partial unwind.

I have actually wanted to call _unwind_ as _unwind_all_ but this means any existing script using this function has to be modified. Agree that it is a bit confusing, but not sure if we can do anything with _unwind_ param at the moment. I might go for the second option you have suggested, i.e. Unwind = True and then Full or Partial unwind operation. Thanks for the tip!

Mr-SabyasachiBose: @rkvinoth you are suggesting this and then inside this, an iterator to unwind multiple rollups

        if unwind:
            if unwind_consol.strip() != '':
                self.remove_edges_under_consolidation(dimension_name, hierarchy_name,unwind_consol)
            else:
                self.remove_all_edges(dimension_name, hierarchy_name)

Mr-SabyasachiBose:
Opening a separate pull request resolving a bug and updating recommended features,

PR #1172, opened 2024-10-16
Feature/simplify execute set mdx migration

onefloid:
# Description

To simplify the migration from the deprecated `DimensionService::execute_mdx` method to `ElementService::execute_set_mdx`, this PR introduces a new method, `get_element_names_by_mdx`, which simulates the old behavior. This new function utilizes execute_set_mdx under the hood.

Last but not least, the *SecurityService_test.py* has been migrated to use the new function.

# Example

For example, we have the following line:

```
elements = self.tm1.dimensions.execute_mdx(dim, mdx)
```

With the changes from this PR, the new call looks like this:

```
elements = self.tm1.elements.get_element_names_by_mdx(mdx)
```

Without these changes, you must write the following to achieve the old behavior:

```
elements = [
    element[0]["Name"]
    for element in self.tm1.elements.execute_set_mdx(
        mdx, member_properties=["Name"], parent_properties=None, element_properties=None
    )
]
```

### DIFF
--- a/TM1py/Services/DimensionService.py
+++ b/TM1py/Services/DimensionService.py
@@ -168,7 +168,7 @@ class DimensionService(ObjectService):
         :return: List of Element names
         """
 
-        warnings.warn("execute_mdx() will be deprecated; use ElementService execute_set_mdx.", DeprecationWarning,
+        warnings.warn("execute_mdx() will be deprecated; use ElementService execute_set_mdx_element_names().", DeprecationWarning,
                       stacklevel=2)
 
         mdx_skeleton = "SELECT " \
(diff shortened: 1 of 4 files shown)


MariusWirtz:
Thank you @onefloid 
I must have written this code countless times in recent years. It's nice to finally have a function.

No strong opinion on the name but I wonder if we could find a name that is in line with how we did it in the `CellService`.

In the `CellService` we have a convention that all MDX functions start with `execute_mdx` e.g., `execute_mdx_dataframe`, `execute_mdx_cellcount`
 
Perhaps for the `ElementService` we could stick to the same idea. So I guess we could call it `execute_set_mdx_element_names`
Then if we add another one later, we could we could call it something like `execute_set_mdx_element_count` or something.

What do you think?

onefloid:
Thanks @MariusWirtz  for the feedback. I have renamed the function as you suggested :-) Now the names are consistent.

MariusWirtz (approved):

PR #1175, opened 2024-10-17
 Feature/unwind specific consolidations while using update_or_create_hierarchy_from_dataframe() and Bug Fix #1174

### LINKED ISSUE #1174: remove_edges_under_consolidation cannot unwind a consol if it is in different case and has different number of spaces
**Describe the bug**
calling remove_edges_under_consolidation() doesn't work if the actual element name in the dimension is "_Parent1_" but the parameter supplied is "_parent1_" or "_Parent     1_", etc. although TM1 considers them as same names for an element.

**To Reproduce**
add a consolidation in a dimension/hierarchy, say **Parent1** and add a child under it say **Child1**
execute the function `remove_edges_under_consolidation( dim_name,  hierarchy_name, "parent 1" )`

**Expected behavior**
**Parent1** should not have **Child1** under it anymore

**Version**
TM1py [2.0]
TM1 Server Version: [e.g. 11.8]

Mr-SabyasachiBose:
## Feature addition ## 

`update_or_create_hierarchy_from_dataframe() `currently unwinds the entire hierarchy when _unwind_ param is set to **True**. A new param _unwind_consol_  of type **List** is added in the function to unwind a list of specific consolidations in the hierarchy. if unwind is **True** and list of elements is provided to _unwind_consol_, it would unwind the valid consolidations from the list. If _unwind_consol_ is empty and unwind is True, then executing the function would unwind entire hierarchy. if _unwind_ is **False**, no unwind operation would execute.

## Code update ##

**Definition**
```

def update_or_create_hierarchy_from_dataframe(
        self,
        dimension_name: str,
        hierarchy_name: str,
        df: 'pd.DataFrame',
        element_column: str = None,
        verify_unique_elements: bool = False,
        verify_edges: bool = True,
        element_type_column: str = 'ElementType',
        unwind: bool = False,
        unwind_consol: list = [],
        update_attribute_types: bool = False):
```

**Unwind consolidation**

        if unwind:
            if len(unwind_consol) == 0:
                self.remove_all_edges(dimension_name, hierarchy_name)
            else:
                for elem in unwind_consol:
                    if self.elements.exists(dimension_name, hierarchy_name, elem):
                        self.remove_edges_under_consolidation(dimension_name, hierarchy_name, elem)    

## Sample Execution ##

```
tm1_serv.hierarchies.update_or_create_hierarchy_from_dataframe(
        dimension_name="Sample Dim",
        hierarchy_name="Sample Dim",
        element_column="ElementID",
        df=df,
        unwind=True,
        unwind_consol=['Child2','parent   2',1,'PARENT1','Parent4']
)
```
 

 

## Bug Fix ##
#1174 

remove_edges_under_consolidation() wasn't unwinding an element if the name is in different in case and having different number of spaces (case and space sensitive).

Added/modified coding to allow the function to work in case and space insensitive nature.

```

    def remove_edges_under_consolidation(self, dimension_name: str, hierarchy_name: str,
                                         consolidation_element: str, **kwargs) -> List[Response]:
        """
        :param dimension_name: Name of the dimension
        :param hierarchy_name: Name of the hierarchy
        :param consolidation_element: Name of the Consolidated element
        :return: response
        """
        hierarchy = self.get(dimension_name, hierarchy_name)
        from TM1py.Services import ElementService
        element_service = ElementService(self._rest)
        elements_under_consolidations = element_service.get_members_under_consolidation(dimension_name, hierarchy_name,
                                                                                        consolidation_element)                                                            
        elements_under_consolidations.append(consolidation_element)
        elements_under_consolidations = [member.lower().replace(' ','') for member in elements_under_consolidations]
        remove_edges = []
        for (parent, component) in hierarchy.edges:    
            if parent.lower().replace(' ','') in elements_under_consolidations and component.lower().replace(' ','') in elements_under_consolidations:
                remove_edges.append((parent, component))
        hierarchy.remove_edges(remove_edges)
        return self.update(hierarchy, **kwargs)
```

### DIFF
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -43,6 +43,7 @@ class AuthenticationMode(Enum):
     SERVICE_TO_SERVICE = 6
     PA_PROXY = 7
     BASIC_API_KEY = 8
+    ACCESS_TOKEN = 9
 
     @property
     def use_v12_auth(self):
@@ -193,7 +194,8 @@ class RestService:
             if self._auth_mode in [
                 AuthenticationMode.IBM_CLOUD_API_KEY,
                 AuthenticationMode.SERVICE_TO_SERVICE,
-                AuthenticationMode.BASIC_API_KEY
+                AuthenticationMode.BASIC_API_KEY,
+                AuthenticationMode.ACCESS_TOKEN
             ]:
                 return True
             else:
@@ -454,7 +456,7 @@ class RestService:
         elif self._auth_mode is AuthenticationMode.SERVICE_TO_SERVICE:
             return self._construct_s2s_service_and_auth_root()
 
-        if self._auth_mode is AuthenticationMode.BASIC_API_KEY:
+        if self._auth_mode in [AuthenticationMode.BASIC_API_KEY, AuthenticationMode.ACCESS_TOKEN]:
             return self._construct_all_version_service_and_auth_root_from_base_url()
 
     def _manage_http_adapter(self):
@@ -699,6 +701,9 @@ class RestService:
         elif self._auth_mode == AuthenticationMode.IBM_CLOUD_API_KEY:
             access_token = self._generate_ibm_iam_cloud_access_token()
             self.add_http_header('Authorization', "Bearer " + access_token)
+
+        elif self._auth_mode == AuthenticationMode.ACCESS_TOKEN:
+            self.add_http_header('Authorization', "Bearer " + self._kwargs.get('access_token'))
 
         # v11 authorization (Basic, CAM) through Headers
         else:
@@ -1063,6 +1068,8 @@ class RestService:
             ]):
                 if self._kwargs.get('user', None) == 'apikey' and 'planninganalytics.saas.ibm.com' in self._base_url:
                     return AuthenticationMode.BASIC_API_KEY
+                elif self._kwargs.get('access_token'):
+                    return AuthenticationMode.ACCESS_TOKEN
                 return AuthenticationMode.BASIC
 
             if self._kwargs.get('gateway', None):


TM1py/Services/HierarchyService.py
```diff
@@ -356,9 +356,10 @@ def remove_edges_under_consolidation(self, dimension_name: str, hierarchy_name:
         elements_under_consolidations = element_service.get_members_under_consolidation(dimension_name, hierarchy_name,
                                                                                         consolidation_element)
         elements_under_consolidations.append(consolidation_element)
+        elements_under_consolidations = [member.lower().replace(' ','') for member in elements_under_consolidations]
```
onefloid, line 359: Have look at *Utils.py*. It offers a lot of helper functions. So  you can `use lower_and_drop_spaces()` for that. Using the helper function improves readability and reduces duplicated code. 
[Link](
https://github.com/cubewise-code/tm1py/blob/2aacccef23737bd9849c6c17f85fa5b10902f4ff/TM1py/Utils/Utils.py#L916)

It would be better to have a separate commit for the bugfix, which only includes the changes for the bugfix. The benefits are:
* It’s easier to understand the commit history
* If there is case where the feature commits must be reverted, the fix remains. 
* The maintainer of the code can cherry pick the bugfix to deliver the bugfix earlier or when your feature request would be declined. 

Maybe it might also be worth to have a own branch/pull request to separate the bugfix from the new feature and allows independent merging.

onefloid (changes requested):
Hi @Mr-SabyasachiBose,

Thank you for sharing your code! I’ve left a review regarding the bug fix.

I also have a general tip for your commits. Your descriptions are good, but the commit titles could be more meaningful. Since most developers only glance at the title, it’s helpful to include a brief explanation of why you’re making these changes. For example, instead of:

> Update HierarchyService.py

It would be clearer to use something like:

> Fix: Compare case and space insensitively

You can see examples of commit messages from other developers in this project [here](https://github.com/cubewise-code/tm1py/commits/master/).

TM1py/Services/HierarchyService.py
```diff
@@ -428,6 +429,7 @@ def update_or_create_hierarchy_from_dataframe(
             verify_edges: bool = True,
             element_type_column: str = 'ElementType',
             unwind: bool = False,
+            unwind_consol: list = [],
```
rkvinoth, line 432: We shouldn't use a mutable default value, use `None` instead and handle the argument accordingly.

TM1py/Services/HierarchyService.py
```diff
@@ -635,8 +641,13 @@ def update_or_create_hierarchy_from_dataframe(
                 use_blob=True)
 
         if unwind:
-            self.remove_all_edges(dimension_name, hierarchy_name)
-
+            if len(unwind_consol) == 0:
```
rkvinoth, line 644: `if not unwind_consol` is the right way to check for an empty list.

TM1py/Services/HierarchyService.py
Mr-SabyasachiBose: Thank you @onefloid, I will explore the function you suggested here for the purpose. Will try isolating the merge request as well.

Mr-SabyasachiBose:
Thanks @MariusWirtz! I am totally buying unwind_all as that was something I thought of earlier but stepped back because of backward compatibility in mind. 

we could add these to the code probably?
```
        # backward compatibility for unwind, the value for unwind would be assinged to unwind_all. expected type is bool
        if "unwind" in kwargs:
            unwind_all = kwargs["unwind"]
```

I have made a few updates as suggested by @onefloid  and @rkvinoth  about using existing utils functions and working with None or list data types. Also, adding additional two lines of code to enforce unwind_consolidation to be list only (if not None)

Issue #1186, opened 2024-10-26
multipart file upload for TM1 v12


MariusWirtz:
Currently, TM1py can't leverage the `use_blob` feature on a large scale with TM1 v12.

TM1 v12 will reject large blob requests. 
To upload large blob files to TM1 v12, TM1py must implement multipart file upload.

PR #1190, opened 2024-11-06
implement mpu file upload

MariusWirtz:
Fixes TM1 v12 (in PA aaS) constraint of 200 MB max file size in file update / create endpoints

Speeds up file upload through parallelization of multi-part upload.

### DIFF
--- a/TM1py/Services/FileService.py
+++ b/TM1py/Services/FileService.py
@@ -1,17 +1,22 @@
 # -*- coding: utf-8 -*-
+import concurrent.futures
 import json
+import time
 import warnings
+from io import BytesIO
 from pathlib import Path
-from typing import List, Iterable, Union
+from typing import List, Iterable, Union, Tuple
 
+from TM1py.Exceptions import TM1pyVersionException
 from TM1py.Services import RestService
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Fixes #1186

Hubert-Heijkers:
Nice one @MariusWirtz!

Quick question though, just because I don't know enough of the details, what happens if one of the requests, specifically the POST of a part, fails? Does the underlying POST repeat that request until it doesn't?

Asking because MPU, other then speeding up things, also helps deal with issues of uploading large files which could still happen whilst uploading a part. Even if the upload of such part succeeds but, due to some network glitch, that is never returned to the initiator of that request. You don't need to clean up extra parts or anything, I'll do that for you in the end, but you might need to retry uploading a part in some of these cases???

PS There is absolutely no constraint on the TM1 v12 side that limits the maximum file size, this is something that's imposed by the PA SaaS environment apparently (just saying;-).

MariusWirtz:
Good catch @Hubert-Heijkers. Thank you

I will make sure to include a retry on the part uploads.

PR #1207, opened 2024-12-04
Fix update_or_create_hierarchy_from_dataframe fails when consolidated…

### LINKED ISSUE #1133: `update_or_create_hierarchy_from_dataframe` fails when Consolidated element already exists with same name but different case
The function should not attempt to create consolidated elements that already exist

cubewise-gng:
Fix #1133 update_or_create_hierarchy_from_dataframe fails when consolidated element exists


MariusWirtz:
Excellent work, Gary! Thank you

MariusWirtz (approved):

PR #1210, opened 2024-12-11
Support paas MTLS certs to RestService

dataslaves:
Update TM1Service and RestService to support users utilizing MTLS connections.


MariusWirtz (approved):

MariusWirtz:
Thanks for the contribution @dataslaves and apologies for the delayed review

PR #1217, opened 2025-01-21
Timezone in Seconds component corrected and format aligned to always include seconds

gtonkin:
Previously there was a bug where the timezone offset would be interpreted as the seconds e.g. UTC+2:00 would result in the chore value having 02 seconds in the schedule which was incorrect.
Updated from_string to always include seconds. Where seconds are not retrieved from TM1 as TM1 uses an ISO standard where 0 seconds are dropped, TM1Py will return 00 as seconds.


MariusWirtz:
Thank you for the contribution @gtonkin!

PR #1219, opened 2025-01-24
Add insert task function to chore

MrKiebitz:


onefloid:
Thanks @MrKiebitz for your contribution.

Altering a property `task._step` which is marked as internal (starts with underscore) from outside the class without a proper setter method is a bit dirty.

But I'm wondering if the `task._step` property is really needed in `ChoreTask` @MariusWirtz  Maybe we could move this to the `Chore` Class and determine the step number (if needed) based on the index of the `tasks` list?

MariusWirtz:
Good catch @onefloid.
The `step` property is not required in the JSON that TM1py sends to TM1 for a `ChoreTask`. 
I agree design-wise, it's better suited for the Chore class. 

However, the TM1py object classes, for the most part, follow the odata entity model and the ChoreTask entity has a `step` property.

@MrKiebitz do you want to implement the proposed change? Or shall we leave it for another PR?

```xml
<EntityType Name="ChoreTask">
    <Key>
        <PropertyRef Name="Step"/>
    </Key>
    <Annotation Term="Core.Description">
        <String>A task to be completed as part of a Chore.</String>
    </Annotation>
    <Annotation Term="Core.Revisions">
        <Collection>
            <Record>
                <PropertyValue Property="Version">
                    <String>10.2.2.1</String>
                </PropertyValue>
                <PropertyValue Property="Kind">
                    <EnumMember>Core.RevisionKind/Added</EnumMember>
                </PropertyValue>
            </Record>
        </Collection>
    </Annotation>
    <Property Name="Step" Type="Edm.Int64" Nullable="false">
        <Annotation Term="Core.Description">
            <String>A step in a task.</String>
        </Annotation>
    </Property>
    <NavigationProperty Name="Process" Type="tm1.Process">
        <Annotation Term="Core.Description">
            <String>The process completed in the task.</String>
        </Annotation>
    </NavigationProperty>
    <NavigationProperty Name="Chore" Type="tm1.Chore" Nullable="false" Partner="Tasks">
        <Annotation Term="Core.Description">
            <String>The Chore that the task belongs to.</String>
        </Annotation>
        <Annotation Term="Core.Revisions">
            <Collection>
                <Record>
                    <PropertyValue Property="Version">
                        <String>11.4.0</String>
                    </PropertyValue>
                    <PropertyValue Property="Kind">
                        <EnumMember>Core.RevisionKind/Deprecated</EnumMember>
                    </PropertyValue>
                </Record>
            </Collection>
        </Annotation>
    </NavigationProperty>
    <Property Name="Parameters" Type="Collection(tm1.ChoreTaskParameter)">
        <Annotation Term="Core.Description">
            <String>A collection of parameters for the task.</String>
        </Annotation>
    </Property>
</EntityType>
```

onefloid:
Good hint @MariusWirtz

Maybe it's beneficial to introduce a generic `update_indexes` method in `Chore` class. Because we need a similar logic for future methods like `remove_task`. 
If cost of execution is crucial, we could work with a flag `chore._index_needs_update` and only update the indexes if the task list is accessed.

MariusWirtz (approved):

PR #1237, opened 2025-04-09
Fix dtype inference and session cookie handling

Cubewise-Ronan:
- Set default dtype to `str` in `pd.read_csv` to avoid misinterpreting values like '2025' as floats.
- Fixed issue with multiple 'TM1SessionId' cookies causing CookieConflictError during session start.
```
File "C:\Dev\share_service\venv\lib\site-packages\TM1py\Services\RestService.py", line 789, in _start_session
    session_id = self._s.cookies.pop('TM1SessionId', None)
  File "C:\Program Files\Python310\lib\_collections_abc.py", line 954, in pop
    value = self[key]
  File "C:\Dev\share_service\venv\lib\site-packages\requests\cookies.py", line 334, in __getitem__
    return self._find_no_duplicates(name)
  File "C:\Dev\share_service\venv\lib\site-packages\requests\cookies.py", line 405, in _find_no_duplicates
    raise CookieConflictError(
requests.cookies.CookieConflictError: There are multiple cookies with name, 'TM1SessionId'
```

### DIFF
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -786,8 +786,22 @@ class RestService:
         finally:
             # If the TM1 REST API is routed through a reverse proxy that alters the expected URL,
             # we explicitly re-set the 'TM1SessionId' cookie to maintain session continuity.
-            session_id = self._s.cookies.pop('TM1SessionId', None)
-            if session_id is not None:
+            session_id = None
+            for cookie in self._s.cookies:
+                if cookie.name == 'TM1SessionId':
+                    session_id = cookie.value
+                    # break  # Use the first match
+
+            # Clear all TM1SessionId cookies to prevent duplicates
+            cookies_to_remove = [
+                (cookie.domain, cookie.path)
+                for cookie in self._s.cookies
+                if cookie.name == 'TM1SessionId'
+            ]
+            for domain, path in cookies_to_remove:
+                self._s.cookies.clear(domain=domain, path=path, name='TM1SessionId')
+
+            if session_id:
                 self._s.cookies.set('TM1SessionId', session_id)
 
             # After we have session cookie, drop the Authorization Header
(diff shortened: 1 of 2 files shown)


TM1py/Utils/Utils.py
```diff
@@ -545,7 +545,7 @@ def build_dataframe_from_csv(raw_csv, sep='~', shaped: bool = False,
 
     # make sure all element names are strings and values column is derived from data
     if 'dtype' not in kwargs:
-        kwargs['dtype'] = {'Value': None, **{col: str for col in range(999)}}
+        kwargs['dtype'] = {'Value': str, **{col: str for col in range(999)}}
```
MariusWirtz, line 548: We don't want the value column to be `str` in all cases, do we?

If we set it to `None` instead, it will be derived depending on the data at hand

TM1py/Services/RestService.py
```diff
@@ -786,8 +786,22 @@ def _start_session(self, user: str, password: str, decode_b64: bool = False, nam
         finally:
             # If the TM1 REST API is routed through a reverse proxy that alters the expected URL,
             # we explicitly re-set the 'TM1SessionId' cookie to maintain session continuity.
-            session_id = self._s.cookies.pop('TM1SessionId', None)
-            if session_id is not None:
+            session_id = None
+            for cookie in self._s.cookies:
+                if cookie.name == 'TM1SessionId':
+                    session_id = cookie.value
+                    # break  # Use the first match
```
MariusWirtz, line 793: Maybe this is an obsolete comment or did you want to break after the first match?

TM1py/Utils/Utils.py
Cubewise-Ronan: Actually, that was the issue I was running into — setting 'Value': None causes pandas to infer the dtype, and in some cases like "2025" (which is an attribute value for an element and should be a string), it gets coerced to 2025.0 if other rows are missing values or inconsistent.

MariusWirtz: I understand. But we can't enforce str as the column type for the value column. Most retrieved cube data will be numeric!

In the edge cases where you load data from attribute cubes it would be easier to pass `dtype=str` to the `execute_mdx_dataframe` function.

PR #1245, opened 2025-06-22
Update Utils.py

wimgielis:
Add a short function to sort lists of names like hierarchy names.
Example use case for a dimnesion called Product. Retrieve the hierarchy names with:
hiers = tm1.hierarchies.get_all_names('Product')

Then use:
print(reorder_with_priority(items=hiers, priority=['Product', 'Leaves'])

Obviously 'Product' will typically be a variable and not hardcoded in the 2 lines of code.

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -1732,6 +1732,23 @@ def utc_localize_time(timestamp):
     return timestamp_utc
 
 
+def reorder_with_priority(items: List[Any], priority: List[Any], unsorted_non_prio_items: bool = True) -> List[Any]:
+    """
+    Reorder a list of items so that items in 'priority' appear first (in given order),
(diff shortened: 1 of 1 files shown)


wimgielis:
Hello,
After that the review and when the merge is done, I will do a second pull request. Since using the function, I added 1 more argument (an exclude_list) which proves useful too. Apologies for the double work.

wimgielis:
Hello Marius,

I can do that. However, what is the specific context?
I need 1 dimension and a few extra hierarchies in the dimension.
Do I need to set up that code too or can I assume, when the test is run,
that some dimensions and hierarchies exist (by name) ?

------

Best regards / Beste groeten,

Wim Gielis
IBM Champion 2024-2025
MS Excel MVP 2011-2014
https://www.wimgielis.com <http://www.wimgielis.be>


Op wo 25 jun 2025 om 17:12 schreef Marius Wirtz ***@***.***>:

PR #1249, opened 2025-07-18
Process changes for JSON Datasource

bdunleavy22:
NOTE:

By adding `
            "DataSource/jsonRootPointer"
            "DataSource/jsonVariableMapping"`
to the URL for getting a process, it may mess up backwards compatability. Needs to be tested in V11


bdunleavy22:
There was a bug in this change. Will create updated request

PR #1252, opened 2025-07-23
Update RestService.py to allow "user" or "username" param

bdunleavy22:
Can use parameter "user" or "username" as discussed in TM1py weekly meeting :)


MariusWirtz:
Thank you @bdunleavy22

MariusWirtz (approved):

PR #1254, opened 2025-07-26
(Enhancement Proposal) alias infromation from MDX View 

Cubewise-JoeCHK:
TM1 Version: 11.8.02700.4

This proposal comes from the project demands. The process requires to output the dataframe from MDXView. At the meantime, the dataframe shows the alias value regarding MDXView setting instead of Principal Name.

In RestAPI response, there's a Meta information to describe View detail setup, including Aliases. but TM1py doesn't cover it yet.  

## Future Plans
MDXView enrichment is foreseeable... in the future, we may construct MDXView object with alias information then send back to TM1 for creating view

## Caution
due to limitation on development resources, this proposal is only tested in v11, not v12.

### DIFF
--- a/TM1py/Objects/MDXView.py
+++ b/TM1py/Objects/MDXView.py
@@ -15,10 +15,10 @@ class MDXView(View):
         IMPORTANT. MDXViews can't be seen through the old TM1 clients (Archict, Perspectives). They do exist though!
     """
 
-    def __init__(self, cube_name: str, view_name: str, MDX: str, Meta: dict = {}):
+    def __init__(self, cube_name: str, view_name: str, MDX: str, meta: dict = {}):
         View.__init__(self, cube_name, view_name)
         self._mdx = MDX
-        self._aliases = Meta.get('Aliases', {})
+        self._aliases = meta.get('Aliases', {})
 
     @property
     def mdx(self):
--- a/TM1py/Services/ViewService.py
+++ b/TM1py/Services/ViewService.py
@@ -67,7 +67,7 @@ class ViewService(ObjectService):
         response = self._rest.GET(url, **kwargs)
         view_as_dict = response.json()
         if "MDX" in view_as_dict:
-            return MDXView(cube_name=cube_name, view_name=view_name, MDX=view_as_dict["MDX"], Meta=view_as_dict.get("Meta", {}))
+            return MDXView(cube_name=cube_name, view_name=view_name, MDX=view_as_dict["MDX"], meta=view_as_dict.get("Meta", {}))
         else:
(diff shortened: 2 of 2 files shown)


TM1py/Objects/MDXView.py
```diff
@@ -15,13 +15,39 @@ class MDXView(View):
         IMPORTANT. MDXViews can't be seen through the old TM1 clients (Archict, Perspectives). They do exist though!
     """
 
-    def __init__(self, cube_name: str, view_name: str, MDX: str):
+    def __init__(self, cube_name: str, view_name: str, MDX: str, Meta: dict = {}):
         View.__init__(self, cube_name, view_name)
         self._mdx = MDX
+        self._aliases = Meta.get('Aliases', {})
```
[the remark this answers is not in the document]
MariusWirtz: @Cubewise-JoeCHK Copilot is right. We must not use mutable default arguments. I think we should apply the suggestion

MariusWirtz:
I was not familiar with the `Meta` property in the response of the `MDXView`!
I can't reproduce it in 11.8.01700.1.

I guess it's an optional or new property then!?
In that case we should make sure all the code works when the property isn't set.

I will add some more comments to the open review that copilot started

TM1py/Objects/MDXView.py
```diff
@@ -15,13 +15,39 @@ class MDXView(View):
         IMPORTANT. MDXViews can't be seen through the old TM1 clients (Archict, Perspectives). They do exist though!
     """
 
-    def __init__(self, cube_name: str, view_name: str, MDX: str):
+    def __init__(self, cube_name: str, view_name: str, MDX: str, Meta: dict = {}):
```
MariusWirtz, line 18: Please name it `meta` instead of `Meta`. 

Unfortunately, we broke with the Python naming conventions many years ago when we named the other argument `MDX` instead of `mdx`, but let's not introduce any more unconventional names

TM1py/Objects/MDXView.py
MariusWirtz, line 21: why not store the `meta` as a property in the class? Perhaps there is more interesting stuff in it?

TM1py/Objects/MDXView.py
```diff
@@ -15,13 +15,39 @@ class MDXView(View):
...
         View.__init__(self, cube_name, view_name)
         self._mdx = MDX
+        self._aliases = Meta.get('Aliases', {})
 
     @property
     def mdx(self):
         return self._mdx
+    
+    @property
+    def aliases(self) -> dict[str, str]:
```
MariusWirtz, line 28: I would call this `alias_by_dimension` and then we need a separate one for `alias_by_hierarchy`

PR #1260, opened 2025-07-30
reorder_with_priority: revised code after Copilot suggestions

wimgielis:
Adding a short function to sort lists. Use case (for me) is sorting of hierarchy names within dimensions, excluding certain hierarchies, such that these hierarchies could be migrated.

### DIFF
--- a/TM1py/Utils/Utils.py
+++ b/TM1py/Utils/Utils.py
@@ -1745,6 +1745,26 @@ def utc_localize_time(timestamp):
     return timestamp_utc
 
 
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Thank you for the contribution @wimgielis 

Perhaps in a separate PR we could expose this functionality as a function on the `Dimension` class.
This would make it more accessible and convenient to use in the context you described.

Kinda like this:

``` python
dim = tm1.dimensions.get()
dim.reorder_hierarchies()
tm1.dimensions.update(dim)
```

PR #1263, opened 2025-08-01
Fix get_elements_dataframe with attribute same name

### LINKED ISSUE #1234: Get_Elements_Dataframe fails when an attribute has the same name as the dimension/hierarchy
Using the Get_Elements_Dataframe function fails when there is an attribute with same name as the dimension.  There is a uniqueness error produced.

ValueError: The column label 'Dimension Name' is not unique.

Because the attribute data is joined back to the element data there has to be a unique key/column (IE the dimension name column that holds the element names)

**To Reproduce**
Create a dimension called MyDim and then create an attribute for elements in that dimension called MyDim.

**Expected behavior**
I have modified the code to add a suffix to any attributes that have the same name as the dimension which solves the issue.  Not sure if this is the best way to go about this or not.

**Version**
TM1py 2.1
TM1 Server Version: 11.0.6.3

cubewise-gng:
Fix #1234 get_elements_dataframe with attribute same name as dimension

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -429,6 +429,18 @@ class ElementService(ObjectService):
 
         cell_service = self._get_cell_service()
 
+        # override column names. hierarchy name with dimension and prefix attributes
+        element_attributes = self.get_element_attributes(dimension_name, hierarchy_name)
+        column_renaming = dict()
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Good catch! Thanks for the contribution Gary!

PR #1267, opened 2025-08-26
changed float parameters to be explicitely defined

nicolasbisurgi:
Added explicit handle on float based parameters in the rest service


MariusWirtz:
@nicolasbisurgi Copilot thinks we should be using integers instead of floats. I think copi is right!

PR #1270, opened 2025-08-27
Update SubsetService.py

bdunleavy22:
Add parameter to get elements dynamically from subset object

### DIFF
--- a/TM1py/Services/SubsetService.py
+++ b/TM1py/Services/SubsetService.py
@@ -13,7 +13,7 @@ from TM1py.Utils import format_url
 
 class SubsetService(ObjectService):
     """ Service to handle Object Updates for TM1 Subsets (dynamic and static)
-    
+
     """
 
     def __init__(self, rest: RestService):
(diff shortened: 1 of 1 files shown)


MariusWirtz:
What do you think about merging the `subset` and `subset_name` arguments into 1 argument like this?


``` python
def get_element_names(
    self,
    dimension_name: str,
    hierarchy_name: str,
    subset: Union[str, "Subset"],
    private: bool = False,
    **kwargs
) -> List[str]:
    """
    Retrieve element names from a static or dynamic subset.

    :param dimension_name: Name of the dimension.
    :param hierarchy_name: Name of the hierarchy.
    :param subset: Subset name (str) or Subset object.
    :param private: Whether the subset is private.
    :param kwargs: Additional arguments.
    :return: List of element names.
    """
    if isinstance(subset, str):
        subset = self.get(subset, dimension_name, hierarchy_name, private=private, **kwargs)
    elif not isinstance(subset, Subet):
        raise ValueError(...)

    if subset.is_static:
        return list(subset_obj.elements)

    element_service = ElementService(self._rest)
    tuples = element_service.execute_set_mdx(
        mdx=subset_obj.expression,
        member_properties=["Name"],
        element_properties=None,
        parent_properties=None,
        **kwargs
    )
    return [entry[0].get("Name", "") for entry in tuples if entry and "Name" in entry[0]]
```

PR #1271, opened 2025-08-27
Update ViewService.py

bdunleavy22:
Added `get_element_names` parameter to `get_native_view` function


bdunleavy22:
I realized afterwards that I also wanted to change the Element properties, so I made it so that you can set the element properties. If you don't want the element names, you can pass an empty list. Although it's a little less clear how to exclude names, it's more in line with other parts of TM1py when it comes to object properties and it's more flexible

PR #1274, opened 2025-09-08
Centered the nice badges

bdunleavy22:
Love the badges, just a small touch up to center them


onefloid (approved):
Thanks for the fine-tuning!

PR #1284, opened 2025-09-25
pass leaves only parameter to recursive function

### LINKED ISSUE #1283: hier function: get_descendants cannot return the right elements if parameter set leaves_only=True
**Describe the bug**
I expected it will return the level0 elements when I set the parameters leaves_only=True, but it still return the consolidated element

**To Reproduce**
from TM1py import TM1Service
import common_util as util
with util.get_tm1_service('FDI') as tm1:
    #print(tm1.server.get_server_name())
    dim = tm1.dimensions.hierarchies.get('Business Unit FPA', 'Business Unit FPA')
    desc = dim.get_descendants('root',recursive=True,leaves_only=True)
    print(type(desc))
    descendants = {}
    for e in desc:
        descendants[e.name]=e.element_type.name
    print(descendants)
    print(len(descendants))

**Expected behavior**
all elements returned expected is level0/leaves elements

**Version**
TM1py 2.1
TM1 Server Version: 11.8.02300.10

**Additional context**

Cubewise-JoeCHK:
close #1283

### DIFF
--- a/TM1py/Objects/Hierarchy.py
+++ b/TM1py/Objects/Hierarchy.py
@@ -175,7 +175,7 @@ class Hierarchy(TM1Object):
                     descendants.add(descendant)
 
             if recursive and descendant.element_type == Element.Types.CONSOLIDATED:
-                descendants = descendants.union(self.get_descendants(descendant.name, True))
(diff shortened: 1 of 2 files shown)


MariusWirtz:
Good catch. Thank you @Cubewise-JoeCHK 

Please add a test case that covers the case described in #1283

PR #1301, opened 2025-10-31
Make networkx dependency optional

### LINKED ISSUE #1300: Make `networkx` an optional dependency
Currently, `networkx` is installed as a required dependency, even though it’s only used for a subset of advanced features. To keep the core package lightweight and minimize dependency bloat, it would be better to make networkx optional.

MariusWirtz:
Switch `verify_edges` default to `False`

Fixes #1300

### DIFF
--- a/TM1py/Services/HierarchyService.py
+++ b/TM1py/Services/HierarchyService.py
@@ -477,7 +477,7 @@ class HierarchyService(ObjectService):
         df: "pd.DataFrame",
         element_column: str = None,
         verify_unique_elements: bool = False,
-        verify_edges: bool = False,
+        verify_edges: bool = True,
         element_type_column: str = "ElementType",
         unwind_all: bool = False,
         unwind_consolidations: Iterable = None,


onefloid:
I hate to ask it, but should we deprecate this first?

This isn’t a real breaking change, but rather a behavioral change with lower risk.

A gradual change could make sense here because the check only affects early error detection, not core functionality. Keeping it temporarily enabled helps avoid surprising users while still allowing the dependency to become optional later. Depending on how heavy the workflows are, it might still be beneficial to fail early to save resources. Also, if TM1 declines building the hierarchy, it should ideally raise the same ValueError; otherwise, the risk of breaking existing code increases due to differences in error handling in the calling code.

vmitsenko:
Hi both, just a suggestion — it might be worth replacing the networkx library with a depth-first search implementation to maintain backward compatibility and simplify dependencies.

onefloid:
@vmitsenko I like this approach. In theory, it could be faster and use less memory than networkx. Have you tested it on a large dataframe?

vmitsenko:
@onefloid Current code is definitely not the most efficient implementation of the depth-first search - it’s simply the easiest and most straightforward to write. However, it appears to run faster and use less memory than when using networkx.

I ran tests using the following dataframe, and these are the results:

```
def generate_df(n_rows=100_000, n_parents=10):
    elements = np.arange(1, n_rows + 1)

    data = {"Element": elements}

    for p in range(1, n_parents + 1):
        parents = elements - p
        parents[parents < 1] = 0
        data[f"Parent_{p}"] = parents

    return pd.DataFrame(data)
```

**depth-first search**
Execution time: 7.2776 seconds
Peak memory usage: 65.047 MB

**networkx**
Execution time: 14.5668 seconds
Peak memory usage: 185.592 MB

Issue #1307, opened 2025-11-07
`write_dataframe_async` with `increment=True` should aggregate before splitting dataframe


MariusWirtz:
When using `write_dataframe_async` with `increment=True`, the aggregation/incrementation should happen before the dataframe is split into smaller chunks for parallel processing.

Currently, the aggregation appears to happen per chunk, meaning that if the original dataframe contains duplicate records (same intersection), these duplicates might be distributed across different chunks. As a result, parallel writes can overwrite each other instead of properly incrementing values.

MariusWirtz:
This was solved with #1310

PR #1310, opened 2025-11-07
Fix write_dataframe_async parallel write overwriting

cubewise-gng:
To address #1307 `write_dataframe_async` to aggregate dataframe before parallel processing.

### DIFF
--- a/Tests/CellService_test.py
+++ b/Tests/CellService_test.py
@@ -1412,7 +1412,7 @@ class TestCellService(unittest.TestCase):
         df = pd.DataFrame(
             {
                 self.dimension_names[0]: ["element 1", "element 1", "element 1", "element 1", "element 1"],
-                self.dimension_names[1]: ["element 2", "element 2", "element 2", "element 2", "element 2"],
+                self.dimension_names[1]: ["element 2", "element 2", "element 2", "element 2", "element 1"],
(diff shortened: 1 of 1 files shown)


cubewise-gng:
In case the dataframe contains a mix of string and numeric values.
I did not want to aggregate strings by calling `aggregate_duplicate_intersections` directly.
Is that the right approach?

MariusWirtz:
> In case the dataframe contains a mix of string and numeric values.
I did not want to aggregate strings by calling aggregate_duplicate_intersections directly.
Is that the right approach?

Yes. Good catch. If increment is `True`, we should increment the numeric values and apply last-one-wins on string values.

Changing the `increment` default value to `True` is technically a breaking change, but I would argue that this counts as a bugfix.
With increment set to `False` (the current behaviour), the function produces nondeterministic results when handling duplicate numeric records.

I will run all tests and merge

TM1py/Utils/Utils.py
MariusWirtz: This should return a dataframe.

 Currently, it returns a dictionary and that breaks existing code and existing test cases.

cubewise-gng: Thanks, made the correction.

MariusWirtz (approved):
Approved. Thanks!

PR #1318, opened 2025-11-28
1317 applicationservice fails to access applications inside private folders

nicolasbisurgi:
### Problem
The ApplicationService failed with 404 errors when accessing applications inside private folders because the TM1 REST API requires PrivateContents('FolderName') for private folder segments, but TM1py was always using Contents('FolderName') for intermediate path segments. 

### Solution: Implemented automatic path resolution with an optimistic probing strategy:
1. Optimistic probing - Try all-public path first (most common case), then all-private path
2. Fallback discovery - If both fail, iteratively probe each segment to find the exact public-to-private transition point
3. Efficient probing - Uses $top=0 query parameter to probe paths without fetching actual data
4. Optional caching - New use_cache parameter (default False) to cache discovered private boundaries for repeated access
5. User warnings - Emits warnings when auto-resolution is used so users are aware of the path correction

### Changes:
- Added _build_path_url() helper to construct URL paths with mixed Contents/PrivateContents segments
- Added _find_private_boundary() to iteratively discover where a path becomes private
- Added _resolve_path() that implements the optimistic probing strategy
- Updated all public methods (get_names, get, exists, delete, rename, create, update, etc.) to use path resolution
- Added use_cache parameter to all affected methods
- Added comprehensive unit tests for the new functionality

### DIFF
--- a/TM1py/Services/ApplicationService.py
+++ b/TM1py/Services/ApplicationService.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 import json
-import warnings
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 from requests import Response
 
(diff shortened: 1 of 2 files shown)


MariusWirtz:
The optimistic approach is very sensible. I like it. The costs for the iterations are acceptable IMO.

However, I noticed that the new code breaks some tests that run fine in the master branch

I think the rename function requires some additional handling, as the URLs with `tm1.Move` extension don't support GET.

image

MariusWirtz (approved):

MariusWirtz:
Yeah. I think the search optimization can be done in a future PR, given the expected performance impact.
 
Thanks everyone, for the feedback on this!

Issue #1362, opened 2026-03-04
MDXView: Support additional body properties (e.g., Meta, IsAutocalculate)


151N3:
### Feature Request

Currently `MDXView.construct_body()` only serializes `@odata.type`, `Name`, and `MDX`. The TM1 REST API accepts additional properties on MDX views, most notably `Meta`, which enables powerful PAW/Planning Analytics Workspace features such as:

- **`Meta.Aliases`** — display members using a specific alias instead of the element name
- **`Meta.ContextSets`** — define subsets or expressions for context (title/WHERE) dimensions
- **`Meta.ExpandAboves`** — control expand-above behaviour per dimension

Without access to these properties from TM1py, these settings can only be configured interactively inside PAW and cannot be managed programmatically or via automation pipelines.

### Proposed Change

Add an optional `properties: dict` parameter to `MDXView.__init__()` and include it in `construct_body()`:

```python
def __init__(self, cube_name: str, view_name: str, MDX: str, properties: Optional[Dict] = None):
    ...
    self._properties = properties or {}

def construct_body(self) -> str:
    mdx_view_as_dict = collections.OrderedDict()
    mdx_view_as_dict["@odata.type"] = "ibm.tm1.api.v1.MDXView"
    mdx_view_as_dict["Name"] = self._name
    mdx_view_as_dict["MDX"] = self._mdx
    mdx_view_as_dict.update(self._properties)
    return json.dumps(mdx_view_as_dict, ensure_ascii=False)
```

Example Usage

```python
view = MDXView(
    cube_name="EIS_CUBE",
    view_name="Dummy_View",
    MDX="SELECT ... FROM [EIS_CUBE] WHERE (...)",
    properties={
        "IsAutocalculate": False,
        "Meta": {
            "Aliases": {"[dim_One].[dim_One]": "Short_DE"},
            "ContextSets": {
                "[dim_Four].[dim_Four]": {"IsPublic": True, "SubsetName": "Default"},
                "[dim_Three].[dim_Three]": {"Expression": "{[dim_Three].[dim_Three].MEMBERS}"}
            },
            "ExpandAboves": {
                "[dim_Four].[dim_Four]": False,
                "[dim_Two].[dim_Two]": False
            }
        }
    }
)
```

onefloid:
Implemented in v2.2.4

PR #1364, opened 2026-03-05
MDXView: Support additional body properties 

151N3:
Check https://github.com/cubewise-code/tm1py/issues/1362

### DIFF
--- a/TM1py/Services/ViewService.py
+++ b/TM1py/Services/ViewService.py
@@ -7,7 +7,7 @@ from requests import Response
 
 from TM1py.Exceptions.Exceptions import TM1pyRestException
 from TM1py.Objects import View
-from TM1py.Objects.MDXView import MDXView, MDX_VIEW_EXCLUDED_KEYS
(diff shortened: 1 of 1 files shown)


TM1py/Objects/MDXView.py
```diff
@@ -70,15 +80,27 @@ def from_json(cls, view_as_json: str, cube_name: Optional[str] = None) -> "MDXVi
...
             MDX=view_as_dict["MDX"],
+            properties={k: v for k, v in view_as_dict.items() if k not in _excluded_keys},
         )
 
     def construct_body(self) -> str:
         mdx_view_as_dict = collections.OrderedDict()
         mdx_view_as_dict["@odata.type"] = "ibm.tm1.api.v1.MDXView"
         mdx_view_as_dict["Name"] = self._name
         mdx_view_as_dict["MDX"] = self._mdx
+        mdx_view_as_dict.update(self._properties)
```
onefloid: @151N3 We need the same logic as in "from_dict," where reserved keys are excluded. Otherwise, you could accidentally overwrite the name, MDX, or cube in the dictionary, which is not the intention.

PR #1366, opened 2026-03-14
Feat: Introduce dynamic property mixin and add dynamic  props to native view

onefloid:
What changed

* Introduce dynamic property mixin
* Add dynamic property support to native view


MariusWirtz (approved):
This looks good. All tests passed on my TM1 environment.

Thanks!!

PR #1396, opened 2026-05-08
Introduce Metric service (for v12)

vmitsenko:
Introduce Metric service for v12


nicolasbisurgi:
closed in favor of #1411

PR #1398, opened 2026-05-20
Fix error message formatting in _handle_proxies and grammatical error

JosephPTM1:
Python doesn't allow `"%s", ` in exceptions, this will raise the exception as a tuple with `"Invalid JSON passed for argument 'proxies': %s"` as a literal string and `proxies` as the other element in the tuple. I have changed it to an f string in this case rather than adapting it to `"%s" % proxies` because TM1py is targeting Python 3.7 and f strings are faster and more readable. I tested this and it does indeed just raise that tuple.

That being said, feel free to change this to "%s" % proxies if the consistency of it is important.

I also spotted a minor grammatical error while reading through the rest of the code with the line `Must be to be of type 'bool' or 'str'"`.

### DIFF
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -257,7 +257,7 @@ class RestService:
             try:
                 return json.loads(proxies)
             except JSONDecodeError:
-                raise ValueError("Invalid JSON passed for argument 'proxies': %s", proxies)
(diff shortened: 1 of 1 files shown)


JosephPTM1:
Apologies for raising 2 PRs in what could have been 1. The first was a fix for something I had just personally experienced, this one was through just reading the rest of the code out of intrigue.

PR #1400, opened 2026-05-21
Created the ability to have an execute_mdx_dataframe_async(...) that will return multiple DFs

stevenorsini:
You can pass in multiple MDXs to the function and get a list of DataFrames back, in order of the MDXs.


MariusWirtz:
Looks good! Thanks @stevenorsini 

Please add a few test cases and use the legacy type annotation (`List`), and then we can merge

PR #1401, opened 2026-05-22
Add element_type/name_pattern/level filters to get_elements, get_element_names, get_elements_dataframe

meyersrl:
## Summary

Adds three optional, AND-composable filter kwargs — `element_type`, `name_pattern`, `level` — to `get_elements`, `get_element_names`, and `get_elements_dataframe`. Translates to OData `$filter` server-side. Refactors `get_elements_by_level` and `get_elements_filtered_by_wildcard` to delegate to the new path (behavior preserved).

## Motivation

Today TM1py can *count* elements by type (`get_number_of_leaf_elements`, `get_number_of_consolidated_elements`, etc.) but can't *list* them without pulling the entire dimension and filtering in Python, or hand-writing MDX. This closes the gap with a small, additive API.

```python
# Before — pull everything, filter in Python:
all_elements = tm1.elements.get_elements(dim, hier)
numeric_leaves = [e for e in all_elements if e.element_type == Element.Types.NUMERIC]

# After:
numeric_leaves = tm1.elements.get_elements(dim, hier, element_type="numeric")

# Or compose all three:
tm1.elements.get_element_names(
    dim, hier,
    element_type=["numeric", "consolidated"],
    name_pattern="Region*",
    level=1,
)
```

## API

`element_type` accepts `Element.Types` enum, str (`'numeric'`/`'string'`/`'consolidated'`, case-insensitive), int (1/2/3), or an iterable of any of those (OR-combined).

`name_pattern` is a glob with `*` wildcard, case- and space-insensitive (matching the semantics of the existing `get_elements_filtered_by_wildcard`). Translates to `startswith`, `endswith`, `contains`, or `eq` depending on `*` placement. `?` raises `ValueError`.

`level` is exact match on the hierarchy level (0 = leaf).

All three default to `None` — purely additive, no breaking changes.

## Design notes worth flagging

**1. Incidental fix in get_elements URL.** The original `get_elements` URL used `?select=Name,Type` (missing the `$` prefix), which TM1's OData endpoint silently ignored — the response always carried the full element payload. This PR corrects it to `?$select=Name,Type`. `Element.from_dict` handles `UniqueName`, `Index`, and `Attributes` via `.get(..., None)` so no existing caller breaks. Same latent bug exists at `get_edges` (line 706) but is **out of scope** for this PR.

**2. element_type overrides skip_consolidations.** In `get_elements_dataframe`, when `element_type` is explicitly set alongside `skip_consolidations=True`, `element_type` is authoritative. The two kwargs are not AND-combined — this is a deliberate design choice documented in the docstring. (Setting `element_type=['numeric','consolidated']` while leaving `skip_consolidations=True` would otherwise be incoherent: you asked for consolidations but the default flag drops them.)

**3. Empty-match dataframe schema.** When the trio filter matches zero elements, `get_elements_dataframe` constructs an empty MDX set via `Tm1FilterByLevel({Members}, 9999)` rather than the bare `{}`. The bare empty set loses the dimension column in the downstream `pd.merge`, breaking the DataFrame schema for callers that index by attribute or level column. The 9999-level filter produces an empty result while preserving the full column schema. A regression test (`test_dataframe_trio_empty_match_preserves_schema`) guards this.

## Test plan

- ~48 pure-function unit tests for the two new private helpers (`_coerce_element_types`, `_build_elements_filter`) in `Tests/ElementService_filtering_helpers_test.py`, covering translation, type coercion, glob-to-OData conversion, single-quote escaping, and every validation error path. No TM1 connection required.
- ~37 integration tests in `Tests/ElementService_test.py::TestElementFiltering` against a fixture dimension covering each kwarg alone, in combination, with case/space variations, quote-escaped names (`O'Brien`), and validation passthrough.
- Regression tests confirming `get_elements_by_level` and `get_elements_filtered_by_wildcard` produce identical output to `master` (snapshots committed under `Tests/fixtures/element_filtering_snapshots/`).
- Regression test confirming `get_elements_dataframe` produces an identical DataFrame to `master` when no trio kwargs are passed.

Total ~85 new tests. Verified against a live TM1 11.8.x server: all green, no regressions in the existing 115 `TestElementService` tests.

## Backwards compatibility

Purely additive. All new kwargs default to `None`. No signatures removed, no return types changed, no error semantics changed for any existing path.

## Files touched

- `TM1py/Services/ElementService.py` — two new private helpers, kwargs added to three public methods, two existing methods refactored to thin delegations.
- `Tests/ElementService_filtering_helpers_test.py` — new file, unit tests for the helpers.
- `Tests/ElementService_test.py` — new `TestElementFiltering` class.
- `Tests/fixtures/element_filtering_snapshots/` — regression oracle snapshots (9 files).

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -1728,3 +1728,160 @@ class ElementService(ObjectService):
         from TM1py import CellService
 
         return CellService(self._rest)
+
+
+# ---------------------------------------------------------------------------
+# Filtering helpers (private, module-level)
+# ---------------------------------------------------------------------------
+
+_TYPE_NAME_TO_CODE = {"numeric": 1, "string": 2, "consolidated": 3}
+_VALID_TYPE_CODES = {1, 2, 3}
+_INVALID_ELEMENT_TYPE_MSG = (
+    "Invalid element_type {value!r}: expected 'numeric', 'string', " "'consolidated', Element.Types enum, or int 1/2/3"
+)
+
+
+def _coerce_one_element_type(value) -> int:
+    """Coerce a single element_type input to its OData type code (1, 2, or 3).
+
+    Accepts Element.Types enum, str (case-insensitive), or int in {1, 2, 3}.
+    Raises ValueError on anything else.
+    """
+    # Reject bool first: bool is a subclass of int in Python, but True/False
+    # are not valid type codes.
+    if isinstance(value, bool):
+        raise ValueError(_INVALID_ELEMENT_TYPE_MSG.format(value=value))
+    if isinstance(value, Element.Types):
+        return int(value.value)
+    if isinstance(value, str):
+        code = _TYPE_NAME_TO_CODE.get(value.lower())
+        if code is None:
+            raise ValueError(_INVALID_ELEMENT_TYPE_MSG.format(value=value))
+        return code
+    if isinstance(value, int) and value in _VALID_TYPE_CODES:
+        return value
+    raise ValueError(_INVALID_ELEMENT_TYPE_MSG.format(value=value))
+
+
+def _coerce_element_types(value) -> List[int]:
+    """Normalize element_type input to a deduplicated list of OData type codes.
+
+    :param value: Element.Types enum, str ('numeric'/'string'/'consolidated', case-insensitive),
+        int in {1, 2, 3}, an iterable of any of those, or None.
+    :return: Deduplicated list of OData type codes (e.g. [1, 3]), preserving first-seen order.
+        Returns [] when value is None.
+    :raises ValueError: on unknown values or empty iterables.
+    """
+    if value is None:
+        return []
+
+    # Single-value path (these come BEFORE the iterable check because str is iterable)
+    if isinstance(value, (Element.Types, str, int)):
+        return [_coerce_one_element_type(value)]
+
+    # Iterable path
+    try:
+        items = list(value)
+    except TypeError:
+        raise ValueError(_INVALID_ELEMENT_TYPE_MSG.format(value=value))
+    if not items:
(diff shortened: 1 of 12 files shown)


MariusWirtz:
Hi @meyersrl,

Sorry for the late review. This looks good and useful! Thanks for the addition.

We have a small conflict now because other PRs have been merged in the meantime, and Copilot left a few comments. 
I think we can merge after we resolve these two issues

meyersrl:
Hi @MariusWirtz — thanks for the review! Both Copilot findings are addressed in 1f0e491, and the branch is rebased on the latest master so the conflict is resolved.

**1. `**kwargs` forwarding in `get_elements_dataframe` trio path**
The internal `get_element_names` call now forwards `**kwargs` so request-level options (`timeout`, `cancel_at_timeout`, `async_requests_mode`) reach the REST call, matching the rest of the method.

**2. URL escaping in `_odata_str_literal`**
The filter clause is concatenated raw into the URL query string (`url += "&$filter=" + clause`), so a `name_pattern` like `Sales & Marketing` would split the query at `&`. Fixed by delegating to `build_url_friendly_object_name` from `TM1py.Utils`, which is the established convention elsewhere in the codebase for user-supplied strings landing in URLs. It percent-encodes `%`, `#`, `?`, `&` and still handles the OData single-quote doubling.

**Tests added** in `Tests/ElementService_filtering_helpers_test.py`:
- `TestOdataStrLiteralUrlSafety` — 6 cases covering each reserved char individually, the OData quote-escape regression guard, and an end-to-end check via `_build_elements_filter` proving the produced clause is URL-safe.
- `TestGetElementsDataFrameForwardsKwargs` — mocks `get_element_names` on a service instance and confirms `timeout` / `cancel_at_timeout` / `async_requests_mode` propagate.

All 55 helpers tests pass locally.

MariusWirtz (approved):

PR #1407, opened 2026-05-28
fix: resolve f-string and format_url conflict for control dimensions

neilnyn:
get_parents_of_all_elements used f-string inside format_url(), causing
  single "}" error  when dimension_name contains '}' (e.g. TM1 control objects
  like }APQ Time Second). Replaced with named .format() placeholders."


MariusWirtz:
Great catch, thank you for fixing this @neilnyn 

Could you also add a regression test for the case where the dimension or hierarchy name contains a closing brace (`}`), so we protect against this in the future?

neilnyn:
> Great catch, thank you for fixing this @neilnyn
> 
> Could you also add a regression test for the case where the dimension or hierarchy name contains a closing brace (`}`), so we protect against this in the future?

yes,of course

MariusWirtz (approved):

PR #1413, opened 2026-06-04
Fix get_measure_dimension; add get/set for measure & time dimension (#1412)

### LINKED ISSUE #1412: get_measure_dimension
get the measures dimension of a cube

Didn't get the expected result? Describe:
get_measure_dimension method in CubeServices is returning the last element of the ordered list of the cube dimension. This is however not always the case - not all measures dimension is the last one.

**Version**
- TM1py [e.g. 1.3.1] 
- TM1 Server Version: [e.g. 11.4]

**Additional context**
If you encounter an error, please add the error message and the stack trace

nicolasbisurgi:
## Summary

Fixes #1412.

`CubeService.get_measure_dimension` returned the cube's **last dimension**, assuming the measures dimension is always last. That assumption does not hold — a cube's measures dimension can be assigned to any dimension — so the method could return the wrong dimension (confirmed live: `plan_BudgetPlan`'s measures dimension is `plan_business_unit`, not its last dimension).

This PR reads the cube's **`MeasuresDimension` navigation property** instead, which reflects the actual assignment and is available on **both v11 and v12**. It also adds the matching time-dimension getter and setters for both.

## Changes

- **`get_measure_dimension`** — now reads `/Cubes('x')/MeasuresDimension`. When no measures dimension is explicitly assigned (server returns `204`), falls back to the last dimension — TM1's own default — preserving prior behavior for that case.
- **`get_time_dimension`** *(new)* — reads `/Cubes('x')/TimeDimension`; returns `None` when unassigned.
- **`set_measure_dimension` / `set_time_dimension`** *(new)* — bind the navigation property via `PATCH` using `@odata.bind`. Decorated `@require_data_admin` and validate that the target dimension belongs to the cube (raising `ValueError` otherwise).
- **`_validate_dimension_in_cube`** *(new private helper)*.

## Testing

5 new tests in `Tests/CubeService_test.py` (set+get round-trips for measure & time, the unset-time `None` case, and `ValueError` on invalid dimensions). Existing `test_get_measure_dimension` retained.

Verified live against both servers in `Tests/config.ini`:
- **v11** (`tm1srv01`, 11.8): 6/6 pass
- **v12** (`tm1srv02`, PA Engine): 6/6 pass

Black and Ruff pass on the changed files.

## Note for reviewers

The v12 `$metadata` marks `MeasuresDimension` with `@Core.Revisions: [{Version: 12.0.0, Kind: Deprecated}]`. It is deprecated but currently functional on v12; flagging in case a different long-term endpoint is preferred.

### DIFF
--- a/TM1py/Services/CubeService.py
+++ b/TM1py/Services/CubeService.py
@@ -101,13 +101,15 @@ class CubeService(ObjectService):
 
         return int(self._rest.GET(url=format_url("/Cubes/$count"), **kwargs).text)
 
+    @require_version(version="11.8.018")
     def get_measure_dimension(self, cube_name: str, **kwargs) -> str:
         """Get the measures dimension of a cube.
 
-        Reads the cube's MeasuresDimension navigation property (available on both v11
-        and v12). When no measures dimension is explicitly assigned, TM1 treats the
-        last dimension of the cube as the measures dimension, so that is returned as
-        a fallback.
+        Reads the cube's MeasuresDimension navigation property. This property is
+        available from v11.8.018 onwards (and on v12); earlier versions do not expose
+        it. When no measures dimension is explicitly assigned, TM1 treats the last
+        dimension of the cube as the measures dimension, so that is returned as a
+        fallback.
 
         :param cube_name:
         :return: name of the measures dimension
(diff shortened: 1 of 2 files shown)


MariusWirtz:
I think we must add `required_version` decorators for these functions.

In version 11.8.01700.1 `MeasuresDimension` and `TimeDimension` properties are not listed as properties of the Cube in the $metadata XML.

The `test_set_and_get_time_dimension` fails in my environment with 11.8.01700.1. They do work on PAoC with 11.8.03500.6

PR #1414, opened 2026-06-05
Change default value for element_type_column parameter to ElementType…

wimgielis:
… instead of Type

tm1.elements.get_elements_dataframe() has a default value of "Type" for element_type_column. tm1.hierarchies.update_or_create_hierarchy_from_dataframe() has a default value of "ElementType" for element_type_column.

This seems to add unnecessary downstream work.

### DIFF
--- a/TM1py/Services/ElementService.py
+++ b/TM1py/Services/ElementService.py
@@ -344,7 +344,7 @@ class ElementService(ObjectService):
         use_blob: bool = False,
         allow_empty_alias: bool = True,
         attribute_suffix: bool = False,
-        element_type_column: str = "Type",
+        element_type_column: str = "ElementType",
         **kwargs,
     ) -> "pd.DataFrame":
         """


MariusWirtz:
@wimgielis I hear you. 
It would have been the right decision to call it "`ElementType`" when we initially wrote it.

We can't change the name of the column now, as it wouldn't be backwards compatible and break countless scripts on upgrade.

Any other ideas to avoid the pitfall and the renaming work (for the column name mismatch) when using both functions together?

Perhaps we can make the `update_or_create_hierarchy_from_dataframe` more lenient to accept `Type` as the element type column if `ElementType` isn't provided in the df?
Just thinking loud...

wimgielis:
Hello Marius, that suggestion seems good to me considering indeed that the
rename change in the code now is too late.



------

Op ma 8 jun 2026 om 10:15 schreef Marius Wirtz ***@***.***>

MariusWirtz:
Closed for https://github.com/cubewise-code/tm1py/issues/1418

PR #1415, opened 2026-06-08
feat: raise TM1pyNetworkException for HTML/Cloudflare responses

### LINKED ISSUE #1393: Raise `TM1pyNetworkException` for HTML / Cloudflare responses instead of `TM1pyRestException`
# Description

In some cases, requests to IBM Planning Analytics on Cloud are blocked upstream by Cloudflare (WAF / bot protection).

Instead of returning JSON, the endpoint responds with an HTML error page (“Sorry, you have been blocked”).

TM1py currently raises a TM1pyRestException, which is misleading because the request never reached TM1.

---

#Example

```
TM1pyRestException: <!DOCTYPE html>
<title>Attention Required! | Cloudflare</title>
...
Cloudflare Ray ID: 9ef02047ab37aac9
```

---

# Expected Behavior

Detect non-JSON / HTML responses (e.g. Content-Type: text/html) and raise:

`TM1pyNetworkException`

Optionally include:

Hint that the request was blocked upstream (e.g. Cloudflare)
Extracted Ray ID if present

---

# Why

This improves troubleshooting by clearly distinguishing:

TM1 errors (valid REST responses)
Network / security layer issues (HTML block pages)

jamestwilkinson:
**PR: Raise TM1pyNetworkException for HTML / Cloudflare responses**

**Summary**
This PR adds a new exception type, TM1pyNetworkException, to handle network/WAF/Cloudflare HTML error pages that are not valid TM1 REST responses, and misleading as TM1 has not received any message. These responses previously caused TM1pyRestException to be raised with confusing error messages.

**Problem**
When TM1 is fronted by Cloudflare, load balancers, proxies, or WAFs, upstream failures often return HTML pages, not JSON.
RestService.verify_response attempted to parse these as TM1 REST errors, resulting in:

- misleading exception messages
- JSON decode errors
- exceptions being caught incorrectly by except TM1pyRestException blocks

_This behaviour is tracked in Issue #1393._

**Solution**
1. Added new exception: TM1pyNetworkException
2. Extracts Cloudflare Ray ID when present
3. Provides clearer error messages for HTML responses
4. Is not a subclass of TM1pyRestException

Updated RestService.verify_response:

1. Detects HTML responses via Content-Type or <html> signature
2. Raises TM1pyNetworkException instead of TM1pyRestException

Added unit tests:
1. Exception properties and formatting
2. Ray ID extraction (colon, space, Cloudflare formats)
3. Inheritance behaviour
4. Integration test for verify_response raising the new exception

**Impact**
- No breaking changes for existing TM1 REST workflows
- Existing except TM1pyRestException blocks continue to behave correctly
- HTML/WAF/Cloudflare failures now produce clear, actionable errors
- Fully backwards‑compatible for environments not using Cloudflare

**Fixes**
Fixes #1393

### DIFF
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -34,7 +34,12 @@ except ImportError:
 
 import http.client as http_client
 
-from TM1py.Exceptions import TM1pyTimeout, TM1pyVersionDeprecationException, TM1pyNetworkException, TM1pyRestException
+from TM1py.Exceptions import (
+    TM1pyNetworkException,
+    TM1pyRestException,
+    TM1pyTimeout,
+    TM1pyVersionDeprecationException,
+)
 
 
 class AuthenticationMode(Enum):
@@ -1151,10 +1156,10 @@ class RestService:
         :Exceptions:
             TM1pyException, raises TM1pyException when Code is not 200, 204 etc.
         """
-              
         if not response.ok:
             content_type = response.headers.get("Content-Type", "")
-            if "text/html" in content_type.lower() or response.text.lstrip().startswith("<!DOCTYPE"):
+            body_start = response.text.lstrip().lower()
+            if "text/html" in content_type.lower() or body_start.startswith(("<!doctype", "<html")):
                 raise TM1pyNetworkException(
                     response.text,
                     status_code=response.status_code,
--- a/TM1py/Services/RestService.py
+++ b/TM1py/Services/RestService.py
@@ -20,7 +20,6 @@ from requests.adapters import HTTPAdapter
 from requests.auth import HTTPBasicAuth
 from urllib3._collections import HTTPHeaderDict
 
-from TM1py.Exceptions.Exceptions import TM1pyTimeout, TM1pyVersionDeprecationException, TM1pyNetworkException
 from TM1py.Utils import (
     CaseAndSpaceInsensitiveSet,
     HTTPAdapterWithSocketOptions,
@@ -35,7 +34,7 @@ except ImportError:
 
 import http.client as http_client
 
-from TM1py.Exceptions import TM1pyRestException
+from TM1py.Exceptions import TM1pyTimeout, TM1pyVersionDeprecationException, TM1pyNetworkException, TM1pyRestException
 
 
 class AuthenticationMode(Enum):
@@ -507,7 +506,7 @@ class RestService:
             except TM1pyTimeout:
                 # Re-raise timeout exceptions as-is
                 raise
-            except TM1pyRestException:
+            except (TM1pyRestException, TM1pyNetworkException):
                 # Re-raise TM1 exceptions as-is
                 raise
             except Exception as retry_error:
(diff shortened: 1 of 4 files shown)


TM1py/Services/RestService.py
```diff
@@ -1152,9 +1152,21 @@ def verify_response(response: Response):
         :Exceptions:
             TM1pyException, raises TM1pyException when Code is not 200, 204 etc.
         """
+              
         if not response.ok:
+            content_type = response.headers.get("Content-Type", "")
+            if "text/html" in content_type or response.text.lstrip().startswith("<!DOCTYPE"):
```
nicolasbisurgi: let's make this case insensitive just in 'case' (pond intended)

nicolasbisurgi:
Hi @jamestwilkinson - thanks for working on this, I agree that is much needed as we don't have dedicated network exceptions.
I've reviewed the code a bit and it looks good, but there are two main issues to address I think:
1. **TM1pyNetworkException isn't exported from `TM1py/Exceptions/__init__.py`**
 Since people import these with `from TM1py.Exceptions import …`, the exception you just built can't be caught by anyone using the library. (Small related thing: `RestService.py` imports it from the deep `…Exceptions.Exceptions` path instead of the package, which is different from how the other services do it.) Just adding it to `__init__.py` and tidying that import sorts it.

2. **The retry loop swallows the new exception**
This one's sneaky. In the reconnect-and-retry block, the code catches `TM1pyRestException` and re-raises it, but anything else falls into a generic `except Exception` that just shrugs and retries. Since the new exception isn't a `TM1pyRestException`, a Cloudflare/HTML error during a retry slips into that generic catch, the retries run out, and the user ends up seeing the *original* disconnect error instead of the real reason — which is exactly the confusing behavior this PR set out to kill. Easy fix: have line 510 catch both exception types.

Please let me know your thoughts.

jamestwilkinson:
Hi @nicolasbisurgi, thanks for the review. I've addressed the points raised and pushed the updated code. Thanks, James.

nicolasbisurgi:
Thanks @jamestwilkinson !
I did one last check when running the tests and one small issue popped up: when switching `RestService.py` to import from the `TM1py.Exceptions` package pulled in `TM1pyVersionDeprecationException`, which was never exported in `__init__.py`. That broke `import TM1py` entirely on the branch (`ImportError` on import). I exported the missing name to fix it.

The rest is all tidy-up coming from ruff and other linter:
- Broadened HTML detection — now case-insensitive and also matches a bare `<html>`, not just uppercase `<!DOCTYPE`
- `Optional[str]` type hints on `_extract_cloudflare_ray_id` and `ray_id`
- Added a docstring note that not subclassing `TM1pyRestException` is intentional
- Added regression tests: a genuine JSON error still raises `TM1pyRestException`, each HTML-detection branch tested in isolation, and an OK response raises nothing
- Whitespace / EOF newline / import-sort cleanup

All 20 tests pass and `ruff check .` is clean. Thanks for the quick turnaround on the earlier feedback — the `CF-RAY` header check was a nice touch. 

Let me know if you are ok with these changes and then we can merge it.

jamestwilkinson:
Thanks @nicolasbisurgi, please go ahead by all means. Kind regards, James.

MariusWirtz (approved):

PR #1421, opened 2026-06-13
fix: two CellService bugs — multi-hierarchy dict overwrite and missing kwargs

### LINKED ISSUE #1068: Bug in clear_with_dataframe() when using multiple hierarchies in dimension_mapping
**Describe the bug**
There is a bug in the Iteration over the hierarchies [here](https://github.com/cubewise-code/tm1py/blob/master/TM1py/Services/CellService.py#L712). 
We initialize the mdx_selections dict and while iterating over the hierarchies we override as we can have just one key for the dimension name in the dict mdx_selections 
```

dimension_mapping = {"mydimension" : ["hierarchie_1", "hierarchie_2"]


 mdx_selections[dimension] = MdxHierarchySet.tm1_subset_all(
                            dimension=dimension,
                            hierarchy=hierarchy).filter_by_level(0)
	#mdx_selections[dimension].to_mdx() = "{TM1FILTERBYLEVEL({TM1SUBSETALL([mydimension].[hierarchie_1])},0)}"
	#mdx_selections[dimension].to_mdx() = "{TM1FILTERBYLEVEL({TM1SUBSETALL([mydimension].[hierarchie_2])},0)}"
```
	
We get a final mdx something like this: mdx_builder.to_mdx() =

```
SELECT
NON EMPTY 
*{[year].[year].[2024]} 
*{[month].[month].[04]} 
*{TM1FILTERBYLEVEL({TM1SUBSETALL([mydimension].[hierarchie_2])},0)} 
DIMENSION PROPERTIES MEMBER_NAME ON 0
FROM [mycube]
```

**Expected behavior**
We would except this mdx result:

mdx_builder.to_mdx() =
```
SELECT
NON EMPTY 
*{[year].[year].[2024]} 
*{[month].[month].[04]} 
* {TM1FILTERBYLEVEL({TM1SUBSETALL([mydimension].[hierarchie_1])},0)}
*{TM1FILTERBYLEVEL({TM1SUBSETALL([mydimension].[hierarchie_2])},0)} 
DIMENSION PROPERTIES MEMBER_NAME ON 0
FROM [mycube]
```

Would be nice to have some more logging, like print the final mdx statement so one can see the what's going on.

There is another minimal Bug: [L692](https://github.com/cubewise-code/tm1py/blob/master/TM1py/Services/CellService.py#L692) 
the variable  dimension is not assigned yet and should write column instead:

` raise ValueError(f"Value for key '{column}' in dimension_mapping must be of type str")`

ToDo: We should check if the mapping-hierarchies match with the ones from the tm1 dimension similar with the `unmatched_dimension_names`

**Version**
TM1py 2.0.0
TM1 Server Version:  11.8

**Additional context**

### LINKED ISSUE #1113: Passing kwargs to CellService.execute_mdx_dataframe_pivot
**Describe the bug**
Cannot pass any key word arguments to the function "execute_mdx_dataframe_pivot" in CellService. 

**To Reproduce**
try to pass a keyword argument to "execute_mdx_dataframe_pivot". For example, tm1.cells.execute_mdx_dataframe_pivot(mdx, cell_properties=["FormattedValue"])

**Expected behavior**
No Error, keyword argument gets interpreted and returns the result with the given keyword.

**Version**
TM1py 2.0.2
TM1 Server Version: 11.8.02200.2

**Additional context**
Seems like an easy fix of adding ", **kwargs" to the arguments of the function, and to the "return self.extract_cellset_dataframe_pivot(
            cellset_id=cellset_id,
            dropna=dropna,
            fill_value=fill_value,
            sandbox_name=sandbox_name)" of the function

kaizeenn:
## Summary

Fixes two bugs in `CellService.py` that have been open for a while.

---

### Fix #1068 — `clear_with_dataframe` overwrites hierarchies

**Root cause:** When `dimension_mapping` maps a dimension to a *list* of hierarchies, the loop at line ~776 does:
```python
for hierarchy in hierarchies:
    mdx_selections[dimension] = MdxHierarchySet.tm1_subset_all(...)
```
Each iteration overwrites the same dict key, so only the last hierarchy survives and the generated MDX is wrong.

**Fix:** Accumulate a `MdxHierarchySet.union()` across all hierarchies so every one contributes to the MDX crossjoin:
```python
combined = None
for hierarchy in hierarchies:
    hset = MdxHierarchySet.tm1_subset_all(dimension=dimension, hierarchy=hierarchy).filter_by_level(0)
    combined = hset if combined is None else combined.union(hset)
mdx_selections[dimension] = combined
```

Also fixed a minor error-message bug on the same line (reported in the same issue): the `ValueError` for a non-str mapping value referenced loop variable `dimension` instead of the DataFrame column variable `column`.

---

### Fix #1113 — `execute_mdx_dataframe_pivot` swallows kwargs

**Root cause:** The method signature had no `**kwargs`, so calling it with e.g. `cell_properties=["FormattedValue"]` or `use_compact_json=True` raised `TypeError` even though `extract_cellset_dataframe_pivot` accepts those arguments.

**Fix:** Add `**kwargs` to the signature and forward it:
```python
def execute_mdx_dataframe_pivot(self, mdx, dropna=False, fill_value=None, sandbox_name=None, **kwargs):
    cellset_id = self.create_cellset(mdx=mdx, sandbox_name=sandbox_name)
    return self.extract_cellset_dataframe_pivot(
        cellset_id=cellset_id, dropna=dropna, fill_value=fill_value, sandbox_name=sandbox_name, **kwargs
    )
```

---

## Tests

Added `Tests/test_cellservice_unit.py` — 6 unit tests that run without a live TM1 server (mocking the REST layer and cube service):

- Single hierarchy str mapping still works
- Two hierarchies both appear in MDX with UNION
- Three hierarchies all appear
- `**kwargs` forwarded correctly to `extract_cellset_dataframe_pivot`
- Extra kwargs like `cell_properties` forwarded
- No kwargs still works

All 6 pass.

### DIFF
--- a/TM1py/Services/CellService.py
+++ b/TM1py/Services/CellService.py
@@ -772,10 +772,17 @@ class CellService(ObjectService):
                     ).filter_by_level(0)
 
                 elif isinstance(hierarchies, Iterable):
+                    # Build a union across all hierarchies so every hierarchy
+                    # contributes to the MDX crossjoin instead of the last one
+                    # overwriting all previous entries (issue #1068).
+                    combined: Optional["MdxHierarchySet"] = None
                     for hierarchy in hierarchies:
-                        mdx_selections[dimension] = MdxHierarchySet.tm1_subset_all(
+                        hset = MdxHierarchySet.tm1_subset_all(
                             dimension=dimension, hierarchy=hierarchy
                         ).filter_by_level(0)
+                        combined = hset if combined is None else combined.union(hset)
+                    if combined is not None:
+                        mdx_selections[dimension] = combined
 
                 else:
                     raise ValueError(f"Unexpected value type for key '{dimension}' in dimension_mapping")
@@ -3123,7 +3130,7 @@ class CellService(ObjectService):
 
     @require_pandas
     def execute_mdx_dataframe_pivot(
-        self, mdx: str, dropna: bool = False, fill_value: bool = None, sandbox_name: str = None
+        self, mdx: str, dropna: bool = False, fill_value: bool = None, sandbox_name: str = None, **kwargs
     ) -> "pd.DataFrame":
         """Execute MDX Query to get a pandas pivot data frame in the shape as specified in the Query
 
@@ -3131,11 +3138,12 @@ class CellService(ObjectService):
         :param dropna:
         :param fill_value:
         :param sandbox_name: str
+        :param kwargs: Additional keyword arguments forwarded to extract_cellset_dataframe_pivot
         :return:
         """
         cellset_id = self.create_cellset(mdx=mdx, sandbox_name=sandbox_name)
         return self.extract_cellset_dataframe_pivot(
-            cellset_id=cellset_id, dropna=dropna, fill_value=fill_value, sandbox_name=sandbox_name
+            cellset_id=cellset_id, dropna=dropna, fill_value=fill_value, sandbox_name=sandbox_name, **kwargs
         )
 
     def execute_mdx_cellcount(self, mdx: str, sandbox_name: str = None, **kwargs) -> int:
--- a/Tests/CellService_test.py
+++ b/Tests/CellService_test.py
@@ -3254,6 +3254,23 @@ class TestCellService(unittest.TestCase):
         self.assertEqual(pivot.shape, (7, 5 * 5))
 
     @skip_if_no_pandas
+    def test_execute_mdx_dataframe_pivot_with_kwargs(self):
+        mdx = (
+            MdxBuilder.from_cube(self.cube_name)
+            .add_hierarchy_set_to_row_axis(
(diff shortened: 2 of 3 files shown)


MariusWirtz:
Thank you @kaizeenn 
Code looks good and all tests succeed.

Please consider the feedback from Copilot. It identified 2 unused imports.
Please also add a regression test that has a real interaction with TM1.

kaizeenn:
Hi @MariusWirtz, I've updated the PR! I removed the 2 unused imports from the unit tests, and I've also added two live integration regression tests in `Tests/CellService_test.py` (one for the execute_mdx_dataframe_pivot kwargs, and one for the clear_with_dataframe multiple-hierarchies logic) that run against a live TM1 instance. Ready for review!

PR #1429, opened 2026-06-26
feat: model TM1 12.6.1 Arrow/Parquet/Flight TI datasources

skido-cw:
## Summary

Adds first-class modelling for the three new TurboIntegrator **input** datasource types introduced in **TM1 v12, build 12.6.1** — Apache Arrow IPC/Feather (`ARROW`), Apache Parquet (`PARQUET`), and Apache Arrow Flight (`FLIGHT`). This is pure object modelling of the `Process` `DataSource` block; no transport coupling and no new service methods. Previously these types fell through to an empty `"DataSource": {}`.

## Changes

- **`Process`**: four new `datasource_flight_*` scalars (constructor, `from_dict`, properties) and two new `_construct_body_as_dict` branches:
  - `ARROW`/`PARQUET` — file name or http(s) URL via `dataSourceNameForServer`/`dataSourceNameForClient`, plus optional `jsonRootPointer` / `jsonVariableMapping` for nested (struct/list/map) columns; no `ascii*` keys.
  - `FLIGHT` — `flightLocation` / `flightDescriptorType` / `flightDescriptor` / `flightAuth`.
  - Both branches match `Type` **case-insensitively** (see "Verified" below).
- **`ProcessService.get` / `get_all`**: select the four flight fields.
- `datasource_type` stays a free string (no `Type` whitelist) and no version gating is added — `ProcessDataSource` is an OData open type whose `Type` is validated by the engine at *execute* time.

## Verified against a live PA 12.6.1 server

End-to-end create → read-back confirmed two behaviors worth flagging:
- The Arrow Flight wire fields are `flightLocation` / `flightDescriptorType` / `flightDescriptor` / `flightAuth` (a `dataSourceFlight*` payload is silently dropped).
- The server **canonicalizes `Type` to title-case** on read-back (`Arrow` / `Parquet` / `Flight`), which is why the body branches match case-insensitively — otherwise `from_dict(get(...))` would round-trip to an empty `DataSource`.

## Testing

- Offline unit tests (no server) for the body shapes, `from_dict` ⇄ `body_as_dict` round-trips, and the title-case `Type` round-trip — **17 tests, all pass**.
- Optional **live** round-trip test gated to TM1 ≥ 12.6.1 (`@skip_if_version_lower_than`), self-cleaning — passes against a live 12.6.1 server.
- `black --check .` and `ruff check .` pass.

## Backward compatibility

No changes to the existing datasource branches (`ASCII` / `None` / `ODBC` / `TM1CubeView` / `TM1DimensionSubset` / `JSON`); the public API stays byte-compatible for every existing datasource type.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/TM1py/Objects/Process.py
+++ b/TM1py/Objects/Process.py
@@ -65,6 +65,11 @@ class Process(TM1Object):
         datasource_subset: str = "",
         datasource_json_root_pointer: str = "",
         datasource_json_variable_mapping: str = "",
+        # --- NEW: Apache Arrow Flight (TM1 12.6.1+) ---
+        datasource_flight_location: str = "",
+        datasource_flight_descriptor_type: str = "",
+        datasource_flight_descriptor: str = "",
+        datasource_flight_auth: str = "",
     ):
         """Default construcor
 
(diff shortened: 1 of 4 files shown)


skido-cw:
Superseded by #1431, which contains this datasource-modelling commit plus the new `write_dataframe(use_parquet=/use_arrow=)` columnar blob-write path. Closing in favor of the combined PR.

Issue #1365, opened 2026-03-06
Cookie Conflict Error


psegal:
**Describe the bug**
Hi, running into a cookies issue where I'm getting an error "requests.cookies.CookieConflictError: There are multiple cookies with name, 'TM1SessionId'". The scripts involved have been working for some time, but having been getting this issue since the TM1py version has been updated,

**To Reproduce**
Run the script.

**Expected behavior**
The script should run without errors.

[ANI_Adj.py](https://github.com/user-attachments/files/25794206/ANI_Adj.py)

**Version**
TM1py : 2.2.2
TM1 Server Version: 11.8.03100.4

**Additional context**
Also using Pulse, and the user running Pulse processes is the same admin user as is running the python scripts

<img width="957" height="483" alt="Image" src="https://github.com/user-attachments/assets/85d43be2-e6c8-4b9a-a3f4-3c9232f814c5" />

PR #1477, opened 2026-09-13
Fix CookieConflictError on reconnect when TM1SessionId exists under two cookie scopes

RobbyMeyers:
### What

`RestService` raises `requests.cookies.CookieConflictError: There are multiple cookies with name, 'TM1SessionId'` when the session jar holds that cookie under two different domain/path scopes. Two call sites are affected:

- `_start_session`'s `finally` block — `self._s.cookies.pop("TM1SessionId", None)`
- the `session_id` property — `self._s.cookies["TM1SessionId"]`

### Why it happens

`_start_session` deliberately re-sets the cookie to survive reverse proxies:

```python
session_id = self._s.cookies.pop("TM1SessionId", None)
if session_id is not None:
    self._s.cookies.set("TM1SessionId", session_id)
```

`cookies.set()` with no `domain`/`path` writes an **unscoped** cookie (`domain=''`, `path='/'`). TM1's own `Set-Cookie` is **scoped** to the host and `/api/v1`. Both entries coexist in the jar.

On the next reconnect, `pop` routes through `RequestsCookieJar._find_no_duplicates`, which **raises** when a name exists under more than one scope. The `None` default never applies — this is a lookup failure, not a miss. The reconnect aborts and the error surfaces on whatever request happened to trigger it.

`session_id` has the same flaw: it catches `KeyError`, but `CookieConflictError` derives from `RuntimeError`, so it propagates uncaught.

Minimal reproduction, no TM1 server needed:

```python
import requests
from requests.cookies import create_cookie

s = requests.Session()
s.cookies.set("TM1SessionId", "AAA")  # what _start_session writes
s.cookies.set_cookie(create_cookie(
    "TM1SessionId", "BBB", domain="tm1.example.com", path="/api/v1"))  # what TM1 returns

s.cookies.pop("TM1SessionId", None)   # CookieConflictError
s.cookies["TM1SessionId"]             # CookieConflictError
```

### When it bites

Any implicit reconnect — most commonly `_execute_sync_request` → `self.connect()` on the `_re_connect_on_session_timeout` path, after TM1 has aged out an idle session (`IdleConnectionTimeOutSeconds`). First connect is always fine; the failure only appears on reconnect, which makes it look intermittent.

We hit this in production on a long-lived service: 28 × HTTP 500 on a dimension-listing endpoint over a 22-hour window, every one of them a browse against a connection idle past the timeout.

### Relationship to #1365

[#1365](https://github.com/cubewise-code/tm1py/issues/1365) reported this same exception and was closed as user error (calling `logout()` twice). Double-logout is one way to reach it, but not the only one — the reconnect path above involves no logout at all. The underlying cookie-scoping issue was never addressed, so the error is still reachable on current `master`.

### The fix

Add two small helpers that tolerate duplicate scopes and prefer the **most specific** one — the cookie TM1 itself most recently issued:

- `_get_cookie(name)` — duplicate-safe read; raises `KeyError` when absent, preserving the `session_id` property's existing v12 `paSession` fallback
- `_pop_cookie(name)` — duplicate-safe `pop`; also clears the stale copies so the jar cannot keep accumulating conflicting entries

Both call sites switch to the helpers. Behaviour is unchanged when only one cookie is present.

### Tests

`Tests/RestService_cookies_test.py`, 9 cases, **no TM1 server required**:

- duplicate scopes on both `_get_cookie` and `_pop_cookie`
- the `session_id` property with duplicates, and its `paSession` (v12) fallback
- `pop` on a missing cookie still returns `None`
- unrelated cookies survive
- a repeated connect/reconnect cycle stays stable

Verified failing before the change and passing after. `black` and `ruff check` are clean on both touched files.


RobbyMeyers:
Pushed a follow-up commit after reviewing this more adversarially. Summary of what was verified, in case it saves a reviewer the work.

### The duplicate is load-bearing — please don't "fix" it by scoping the re-set

The tidiest-looking fix is to avoid creating the duplicate at all, by preserving the original `domain`/`path` when re-setting. **That breaks the feature this block exists for.** Measured against a local server that issues `Set-Cookie: TM1SessionId=...; Path=/api/v1`, then calling a proxy-rewritten path:

| strategy | `/api/v1/Cubes` | `/tm1/db/api/v1/Cubes` (proxy-rewritten) | value freshness |
|---|---|---|---|
| this PR | sent | **sent** | current |
| preserve original scope | sent | **not sent** | n/a — request is unauthenticated |
| prefer the unscoped copy | sent | sent | **stale — pinned to the first id forever** |

The unscoped re-set on the current line 1041 is deliberate and must stay, which is why the read side has to tolerate duplicates instead. I've added `test_reset_cookie_stays_unscoped_for_reverse_proxies` to pin it, so that refactor fails a test rather than silently unauthenticating proxied users.

### Other read-side approaches that don't work

- `cookies.get(...)` — also routes through `_find_no_duplicates`, still raises.
- `try/except` around `pop`, or `get_dict()` + `set` — no exception, but the duplicate survives and the client then sends **two** session ids in one header: `Cookie: TM1SessionId=SID_3; TM1SessionId=SID_1`. The server picks arbitrarily.

With this PR the header is a single `TM1SessionId=SID_3`.

### Why the scoped copy is the right one to prefer

It's the one the server just issued; the unscoped copy is the previous generation left by the last re-set. Across consecutive connects the jar holds `{(host,'/api/v1'): SRV3, ('','/'): SRV2}`, so preferring the scoped copy is preferring the fresh id. That's also RFC 6265 §5.4 precedence (longer path first). `test_rotating_session_ids_keep_the_newest` pins it.

### Follow-up commit

- sort key gains a `domain` tiebreaker, so the ordering is total and never depends on jar insertion order (unreachable when a service talks to one host, but free)
- the two regression tests above
- fixture built from a real `RestService` instance instead of a `MagicMock`, plus a guard test pinning the raw-jar behaviour these helpers exist to avoid

### On #1365

That issue reported this exact exception and was closed as user error (logging out twice). Double-logout is one route, but not the only one: this reproduces with **zero** logouts, purely via `_execute_sync_request` → `connect()` on session timeout, which is the default path (`re_connect_on_session_timeout=True`). Upstream `master` raises on the second connect:

```
--- connect attempt 1 ---
  jar: domain='127.0.0.1' path='/api/v1'
  after re-set: domain='' path='/'
--- connect attempt 2 ---
  jar: [scoped SERVER_SID_2, unscoped SERVER_SID_1]
  !!! requests.cookies.CookieConflictError: There are multiple cookies with name, 'TM1SessionId'
```

Offline test suite is unchanged at 310 passed with identical pre-existing failures (`KeyError: 'tm1srv01'`, needs a live TM1 server). The two collection errors in `HierarchyService_test.py` / `ServerService_test.py` are present on unpatched `master` as well.

MariusWirtz:
@RobbyMeyers thanks for this contribution! I just stumbled over the exact same issue in a project.

Copilot raised a finding on the PR, that the ordering can promote a cookie belonging to an unrelated host. I don't think it applies here. The session is created internally and only ever talks to the TM1/PA host, so the only thing that can put a TM1SessionId in the jar is TM1 itself. 

Merging as is.

